PR #1, opened 2018-05-11
OAuth integration

peterserfozo:


cnovak:
Do we need to put this in a beta branch in case issues need to be fixed in alpha?

mxr576:
@cnovak No, we should not wait with this until the first beta. We should not just add fixes to the next alphas. We should introduce *new features* (that should be available in the first stable) in the next alpa releases to be able to get feedbacks about them. We should not introduce new features in betas because we have to guarantee backward compatibility after the first beta has released. 

> The APIs are frozen enough so that contributed module and theme authors can start upgrading their projects.

The OAuth integration is a typical example of a feature that may introduce some changes in the module's API, but this is fine because we do not need to worry about BC in alphas.

https://www.drupal.org/docs/8/understanding-drupal-version-numbers/what-are-alpha-and-beta-releases-and-release-candidates

We should release new alphas regularly, probably after each sprint until we have not covered all user stories that should be available in the first stable release.

src/Form/AuthenticationForm.php
```diff
@@ -124,29 +124,147 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      $oauth_token_keys[$key_id] = $this->t('@key_name <a href=":url">Edit</a>', [
+        '@key_name' => $key_name,
+        ':url' => Url::fromRoute('entity.key.edit_form', ['key' => $key_id, 'destination' => 'admin/config/apigee-edge/settings'])->toString(),
+      ]);
+    }
+    $oauth_token_default_value = array_key_exists($config->get('active_key_token'), $oauth_token_keys) ? $config->get('active_key_token') : NULL;
+
+    // Display warning message whether there is no available key.
+    if (empty($basic_auth_keys) && (empty($oauth_keys) || empty($oauth_token_keys))) {
+      $this->messenger->addWarning(t('There is no available key for connecting to Apigee Edge server. <a href=":link">Create a new key.</a>', [
```
mxr576: Use `$this->t()` instead.

src/Form/AuthenticationForm.php
```diff
@@ -124,29 +124,147 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+        $description_link = $this->t('create a new OAuth (SAML) key');
+      }
+      else {
+        $title = $this->t('There is no available OAuth (SAML) token key for connecting to Apigee Edge.');
+        $description_link = $this->t('create a new OAuth (SAML) token key');
+      }
+      $form['authentication']['key_oauth'] = [
+        '#type' => 'item',
+        '#title' => $title,
+        '#description' => $this->t('Select a basic authentication key or <a href=":link">' . $description_link . '</a>.', [
```
mxr576: You should add the `$description_link` by using a placeholder too.

src/Form/AuthenticationForm.php
```diff
@@ -124,29 +124,147 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+        $title = $this->t('There are no available OAuth (SAML) and OAuth (SAML) token keys for connecting to Apigee Edge.');
+        $description_link = $this->t('create new OAuth (SAML) and OAuth (SAML) token keys');
+      }
+      elseif (empty($oauth_keys)) {
+        $title = $this->t('There is no available OAuth (SAML) key for connecting to Apigee Edge.');
+        $description_link = $this->t('create a new OAuth (SAML) key');
+      }
+      else {
+        $title = $this->t('There is no available OAuth (SAML) token key for connecting to Apigee Edge.');
+        $description_link = $this->t('create a new OAuth (SAML) token key');
```
mxr576: I'd call this variables as `$description_link_text`.

src/OauthTokenStorage.php
```diff
@@ -0,0 +1,172 @@
...
+   *   Number of seconds extracted from token's expiration date.
+   */
+  public function __construct(KeyInterface $key, int $leeway = 30) {
+    if (!($key->getKeyType() instanceof EdgeOauthTokenKeyTypeInterface)) {
+      throw new \InvalidArgumentException("Type of {$key->id()} key does not implement EdgeOauthTokenKeyTypeInterface.");
+    }
+
+    $this->key = $key;
+    $this->keyType = $key->getKeyType();
+    $this->leeway = $leeway;
```
mxr576: Because the value of the leeway could not come elsewhere it should come from the OAuth key storage I think. 
(It is not necessary to be fixed now just leave a TODO comment here, remove it from the constructor and set its default value to 30s.)

src/OauthTokenStorage.php
```diff
@@ -0,0 +1,172 @@
...
+   * @throws \Apigee\Edge\Exception\OauthAuthenticationException
+   *   Thrown when the OAuth data cannot be saved.
+   */
+  public function saveToken(array $data): void {
+    try {
+      $this->key->setKeyValue($this->keyType->serialize($data));
+      $this->key->save();
+    }
+    catch (EntityStorageException $exception) {
+      throw new OauthAuthenticationException('Could not save the OAuth response.');
```
mxr576: s/OAuth response/OAuth token data/g

src/OauthTokenStorage.php
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    if ($this->keyType->getExpiresIn($this->key) !== NULL && $this->keyType->getExpiresIn($this->key) !== 0) {
+      $this->expires = $this->keyType->getExpiresIn($this->key) + time();
```
mxr576: You should store the calculated expiration time (Unix timestamp) in the OAuth token storage too.
(If this function is being called next time, in a different page, what is going the be a return value of the `hasExpired()` method in your current code?)

mxr576: Please write a unit test for this class to ensure its implementation is always correct.

src/Plugin/KeyInput/BasicAuthKeyInput.php
```diff
@@ -121,12 +90,21 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
...
-      'password' => $input_values['password'],
-    ]);
+    $key_values = Json::decode($input_values['key_value']);
+
+    $key_values['organization'] = $input_values['organization'];
+    $key_values['username'] = $input_values['username'];
+    $key_values['password'] = $input_values['password'];
+
+    // Remove endpoint key from the JSON if the field is empty.
+    if ($input_values['endpoint'] !== '') {
```
mxr576: Why not `!empty()`?

src/Plugin/KeyInput/OauthKeyInput.php
```diff
@@ -0,0 +1,103 @@
...
+      '#required' => $key->getKeyProvider()->getPluginDefinition()['key_value']['required'],
+      '#default_value' => $values['client_id'] ?? $config->get('default_client_id'),
+      '#attributes' => ['autocomplete' => 'off'],
+    ];
+    $form['client_secret'] = [
+      '#type' => 'password',
+      '#title' => $this->t('Client secret'),
+      '#description' => t('A secret known only to the client and the authorization server. Leave empty to use the default "%client_secret" client secret.', ['%client_secret' => $config->get('default_client_secret')]),
+      '#required' => FALSE,
+      '#attributes' => ['autocomplete' => 'off'],
```
mxr576: Is there autocomplete on password fields by default? :O

src/Plugin/KeyInput/OauthKeyInput.php
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function processSubmittedKeyValue(FormStateInterface $form_state) {
+    $input_values = $form_state->getValues();
+    $key_values = Json::decode($input_values['key_value']);
+
+    // Remove endpoint key from the JSON if the field is empty.
```
mxr576: Actually "endpoint" is not in the list here. :)

src/Plugin/KeyProvider/EnvironmentVariablesKeyProvider.php
```diff
@@ -0,0 +1,146 @@
...
+  public function validateConfigurationForm(array &$form, FormStateInterface $form_state) {
+    $missing_env_variables = [];
+    foreach ($this->getEnvironmentVariables(TRUE) as $variable) {
+      if (!getenv($variable)) {
+        $missing_env_variables[] = $variable;
+      }
+    }
+
+    if (!empty($missing_env_variables)) {
+      $form_state->setError($form, t('The following environment variables are not set: @missing_env_variables.', [
```
mxr576: `$this->t()`

src/Plugin/KeyType/OauthKeyType.php
```diff
@@ -0,0 +1,104 @@
...
+namespace Drupal\apigee_edge\Plugin\KeyType;
+
+use Apigee\Edge\HttpClient\Plugin\Authentication\Oauth;
+use Drupal\apigee_edge\OauthTokenStorage;
+use Drupal\apigee_edge\Plugin\EdgeOAuthKeyTypeInterface;
+use Drupal\key\KeyInterface;
+use Http\Message\Authentication;
+
+/**
+ * Key type for Apigee Edge basic authentication credentials.
```
mxr576: s/basic/Oauth/g

src/Plugin/KeyType/OauthKeyType.php
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * @KeyType(
```
mxr576: Can't you subclass this `KeyType` annotation class as `EdgeKeyType`, where could set some default settings for our keys like 
* group => "apigee_edge"
* multivalue=>TRUE
* origanization, endpoint, username are required", etc

and use `@EdgeKeyType` in the  annotation block? That way you would not need to repeat yourself here.

src/Plugin/KeyType/BasicAuthKeyType.php
```diff
@@ -86,7 +102,7 @@ public function validateKeyValue(array $form, FormStateInterface $form_state, $k
   /**
    * {@inheritdoc}
    */
-  public function getAuthenticationMethod(KeyInterface $key) : Authentication {
+  public function getAuthenticationMethod(KeyInterface $key, KeyInterface $key_token = NULL) : Authentication {
```
mxr576: Key token interface is not needed here, can't you introduce a new optional parameter just in OauthKeyType class's `getAuthenticationMethod()` in the end of the original parameter list (`function getAuthenticationMethod(KeyInterface $key) : Authentication {}`).

src/SDKConnector.php
```diff
@@ -201,10 +203,36 @@ private function setCredentials(CredentialsInterface $credentials) {
...
 
+  /**
+   * Builds credentials, which depends on the KeyType of the key entity.
+   *
+   * @param \Drupal\key\KeyInterface $key
+   *   The key entity which stores the API credentials.
+   * @param \Drupal\key\KeyInterface|null $key_token
+   *   The OAuth token key entity.
+   *
+   * @return CredentialsInterface
```
mxr576: use FQCN.

mxr576 (changes requested):

PR #8, opened 2018-05-17
Oauth integration

peterserfozo:


src/Form/AuthenticationForm.php
```diff
@@ -118,35 +118,148 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      ]);
+    }
+    $oauth_token_default_value = array_key_exists($config->get('active_key_oauth_token'), $oauth_token_keys) ? $config->get('active_key_oauth_token') : NULL;
+
+    $form['authentication']['key_type'] = [
+      '#type' => 'select',
+      '#title' => $this->t('Apigee Edge API authentication type'),
+      '#options' => [
+        'apigee_edge_basic_auth' => $this->t('Basic authentication'),
+        'apigee_edge_oauth' => $this->t('OAuth (SAML)'),
```
mxr576, line 160: IIRC we have not got a clear answer yet whether we should remove "SAML" from the strings, we can do it in an upcoming sprint.

src/OauthTokenStorage.php
```diff
@@ -0,0 +1,169 @@
...
+  public function saveToken(array $data): void {
+    try {
+      $this->key->setKeyValue($this->keyType->serialize($data));
+      $this->key->save();
+    }
+    catch (EntityStorageException $exception) {
+      throw new OauthAuthenticationException('Could not save the OAuth token data.');
+    }
+
+    if ($this->keyType->getExpiresIn($this->key) !== NULL && $this->keyType->getExpiresIn($this->key) !== 0) {
```
mxr576: Are you saving the calculated expiration date (unix timestamp) to the token storage this way? Because it should be saved, you can not use `expires_in` (number of seconds while the access token is valid from the time when it has been generated) after this object got destructed because the current page request has ended.

mxr576 (changes requested):

src/Form/AuthenticationForm.php
peterserfozo: Agreed.

src/OauthTokenStorage.php
```diff
@@ -138,16 +134,15 @@ public function getScope(): string {
    */
   public function saveToken(array $data): void {
     try {
+      if ($this->keyType->getExpiresIn($this->key) !== NULL && $this->keyType->getExpiresIn($this->key) !== 0) {
```
mxr576: Why these conditions are needed? Why both of of them are needed?

mxr576 (approved):

peterserfozo:
Tests: https://travis-ci.org/peterserfozo/apigee-edge-drupal/builds/383044949

PR #9, opened 2018-05-18
Disable allow selecting multiple products

tamasd:


apigee_edge.module
```diff
@@ -844,6 +844,8 @@ function template_preprocess_app_credential(array &$variables) {
       }
     }
 
+    $value = (string) $value;
```
mxr576: So it was passed as on object before? Always or sometimes? If sometimes when it was passed as an object instead of a string? Shouldn't this has been fixed there instead of here?

src/Form/AppSettingsForm.php
```diff
@@ -90,6 +91,7 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       '#type' => 'checkbox',
       '#title' => $this->t('Allow selecting multiple products'),
       '#default_value' => $generalConfig->get('multiple_products'),
+      '#disabled' => !($this->config('apigee_edge.dangerzone')->get('skip_developer_app_settings_validation')) && (bool) DeveloperApp::loadMultiple(),
```
mxr576: Do not load all developer apps as object unnecessary. Just get the list of developer app *ids*.

mxr576 (changes requested):

src/Form/AppSettingsForm.php
mxr576: Please also add a comment to this line that "dangerzone variable" check was only needed to make tests pass.

apigee_edge.module
tamasd: Sometimes it is an object, sometimes it is a string if I have seen correctly. And on the other side, sometimes it is fine to pass it as an object, sometimes it is not.

That is why I chose this solution, because it is clean and simple.

mxr576: I see, but the questions is WHY? Why sometimes it is an object? If it is an object then should not be skipped by default rather? Casting to string is not always the best solution especially if the object does not support that.
Probably the date object properties (expiresAt, issuedAt) causing this, otherwise I would bet for the attributes property which is also an object.

mxr576: Couldn't you catch in which conditions exactly the $value is an object?

tamasd: The error is not catchable. What happens is that the `TranslatableMarkup` object will get passed to `htmlspecialchars()` (IIRC, or a similar funciton) and then I get a warning.

mxr576: Okay, I see now. Please add a comment about this above the line and then I can accept this PR.

peterserfozo (approved):

mxr576 (approved):

mxr576 (approved):

PR #25, opened 2018-06-06
Api product association

peterserfozo:


tamasd (approved):

src/Entity/Form/DeveloperAppCreateForm.php
```diff
@@ -73,6 +73,9 @@ public static function create(ContainerInterface $container) {
 
   /**
    * {@inheritdoc}
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
```
mxr576: I am little bit sceptical whether we should document *all* possible exceptions that Phpstorm offers for us. If we know for sure that an exception should not be thrown in normal circumstances (confidence lvl 99%) then we should not document it, because it could be misleading.
Ex.: `\Drupal\Component\Plugin\Exception\PluginNotFoundException` should never be thrown because "api_product" entity always exists in our module.

src/Entity/Form/DeveloperAppCreateForm.php
```diff
@@ -208,38 +209,38 @@ protected function actions(array $form, FormStateInterface $form_state) {
...
```
mxr576: Same as above.

src/Form/AppSettingsForm.php
```diff
@@ -69,90 +92,70 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
-    }
-    catch (EntityStorageException $e) {
-      // Apigee Edge credentials are missing/incorrect or something else went
-      // wrong. Do not redirect the user to the error page.
-      $product_list = [];
-      $this->messenger()->addError($this->t('Unable to retrieve API product list from Apigee Edge. Please ensure that <a href=":link">Apigee Edge connection settings</a> are correct.'), [
-        ':link' => Url::fromRoute('apigee_edge.settings')->toString(),
-      ]);
-    }
+    // It's necessary to add a wrapper, because if the ID is added to the
```
mxr576: Unnecessary comma.

src/Form/AppSettingsForm.php
```diff
@@ -69,90 +92,70 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+    // checkboxes form element, then that will not be properly rendered
+    // (the label is duplicated).
```
mxr576: s/is/gets/g

src/Form/AppSettingsForm.php
```diff
@@ -69,90 +92,70 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
-    catch (EntityStorageException $e) {
-      // Apigee Edge credentials are missing/incorrect or something else went
-      // wrong. Do not redirect the user to the error page.
-      $product_list = [];
-      $this->messenger()->addError($this->t('Unable to retrieve API product list from Apigee Edge. Please ensure that <a href=":link">Apigee Edge connection settings</a> are correct.'), [
-        ':link' => Url::fromRoute('apigee_edge.settings')->toString(),
-      ]);
-    }
+    // It's necessary to add a wrapper, because if the ID is added to the
+    // checkboxes form element, then that will not be properly rendered
```
mxr576: Unnecessary comma.

src/Form/AppSettingsForm.php
```diff
@@ -69,90 +92,70 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+    // (the label is duplicated).
+    $form['api_product']['default_api_product_multiple_container'] = [
+      '#type' => 'container',
+      '#id' => 'default-api-product-multiple',
+    ];
 
-    $form['api_product']['default_api_product_multiple'] = [
+    $form['api_product']['default_api_product_multiple_container']['default_api_product_multiple'] = [
       '#type' => 'checkboxes',
```
mxr576, line 158: You could have tried to use multiple checkbox form element if "checkboxes" that problematic :)

src/Form/AppSettingsForm.php
```diff
@@ -69,90 +92,70 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   *
+   * @return array
+   *   The renderable array.
+   *
+   * @see https://www.drupal.org/project/drupal/issues/2855139
+   */
+  public function apiProductListCallback(array &$form, FormStateInterface $form_state) : array {
+    return $form['api_product']['default_api_product_multiple_container'];
```
mxr576: I prefer if AJAX commands gets returned, like [ReplaceCommand](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Ajax%21ReplaceCommand.php/class/ReplaceCommand/8.5.x), instead of render elements directly.

https://api.drupal.org/api/drupal/core%21core.api.php/group/ajax/8.5.x

mxr576 (changes requested):

src/Form/AppSettingsForm.php
peterserfozo: But I should create custom validation for the multiple checkbox form element approach (must choose at least one product if user_select is unchecked).

docker-compose.yml
```diff
@@ -58,5 +58,11 @@ services:
     depends_on:
       - php
```
mxr576, line 61: This line is missing from the definition of the PHP container: https://github.com/mxr576/my-webdev-docker-setup/blob/master/docker-compose.yml#L52

tests/src/Functional/ApigeeEdgeTestTrait.php
```diff
@@ -0,0 +1,267 @@
...
+   * @param string $property_name
+   *   The name of your private/protected property.
+   *
+   * @return \ReflectionProperty
+   *   The property you asked for
+   *
+   * @throws \ReflectionException
+   *   If the class or object do not exist.
+   */
+  public static function getInvisibleProperty($object, $property_name) {
```
mxr576, line 291: Do you still need this?
If not now then in the next sprint we can probably replace it with \PHPUnit\Framework\Assert::readAttribute.

tests/src/FunctionalJavascript/AppSettingsFormTest.php
```diff
@@ -0,0 +1,99 @@
...
+
+namespace Drupal\Tests\apigee_edge\FunctionalJavascript;
+
+use Drupal\Core\Url;
+
+/**
+ * App settings form tests.
+ *
+ * @group apigee_edge
+ * @group apigee_edge_developer_app
```
mxr576: Maybe introducing a `apigee_edge_javascript` group would not be a bad idea for JS tests.

mxr576 (changes requested):

tests/src/FunctionalJavascript/AppSettingsFormTest.php
mxr576: This is the reason why this failed: https://travis-ci.org/peterserfozo/apigee-edge-drupal/jobs/391985870#L2311

tests/src/Functional/ApigeeEdgeTestTrait.php
peterserfozo: I agree. I will improve the existing tests in the next sprint.

docker-compose.yml
peterserfozo: Fixed.

mxr576 (approved):

PR #29, opened 2018-06-20
Sync developer attributes

peterserfozo:


src/Entity/Developer.php
```diff
@@ -105,11 +105,12 @@ public static function createFromDrupalUser(UserInterface $user): Developer {
...
-    foreach (\Drupal::config('apigee_edge.sync')->get('user_fields_to_sync') as $field_to_sync) {
-      $developer->setAttribute(
-        preg_replace('/^field_(.*)$/', '${1}', $field_to_sync),
-        Json::encode($user->get($field_to_sync)->getValue())
-      );
+    /** @var \Drupal\apigee_edge\FieldStorageFormatManager $format_manager */
+    $format_manager = \Drupal::service('plugin.manager.apigee_field_storage_format');
+    foreach (\Drupal::config('apigee_edge.sync')->get('user_fields_to_sync') as $field) {
+      $type = $user->getFieldDefinition($field)->getType();
+      $formatter = $format_manager->lookupPluginForFieldType($type);
```
tamasd: While it is unlikely that the formatter will be null, we shouldn't assume that it can't be null.

src/Job/DeveloperSync.php
```diff
@@ -130,21 +131,42 @@ protected function executeRequest() {
...
     parent::execute();
 
+    // Update Apigee Edge developers and Drupal users.
+    $identical_accounts = array_intersect($this->edgeAccounts, $this->drupalAccounts);
+    foreach ($identical_accounts as $search => $mail) {
+      /** @var \Apigee\Edge\Api\Management\Entity\DeveloperInterface $developer */
+      $developer = Developer::load($mail);
+      /** @var \Drupal\user\UserInterface $account */
+      $account = user_load_by_mail($mail);
+      $last_modified_delta = $developer->getLastModifiedAt()->getTimestamp() - $account->getChangedTime();
```
tamasd: What should happen when the `$last_modified_delta` is 0?

src/Job/DeveloperUpdate.php
```diff
@@ -19,41 +19,88 @@
...
+      $this->executeUpdate = TRUE;
+    }
+
+    $user_fields_to_sync = \Drupal::config('apigee_edge.sync')->get('user_fields_to_sync');
+    if (!empty($user_fields_to_sync)) {
+      /** @var \Drupal\apigee_edge\FieldStorageFormatManager $format_manager */
+      $format_manager = \Drupal::service('plugin.manager.apigee_field_storage_format');
+      foreach ($user_fields_to_sync as $field) {
+        $type = $account->getFieldDefinition($field)->getType();
+        $formatter = $format_manager->lookupPluginForFieldType($type);
```
tamasd: Same as above, you shouldn't expect that `$formatter` never be null.

src/Job/UserCreate.php
```diff
@@ -73,7 +73,12 @@ protected function executeRequest() {
         throw new EntityMalformedException((string) $violation->getMessage());
       }
     }
+
+    // If the developer-user synchronization is in progress, then saving
+    // developers while saving Drupal user should be avoided.
+    _apigee_edge_set_sync_in_progress(TRUE);
     $user->save();
+    _apigee_edge_set_sync_in_progress(FALSE);
```
tamasd: I'd rewrite this with a try-finally block:

```
try {
  _apigee_edge_set_sync_in_progress(TRUE);
  $user->save();
}
finally {
  _apigee_edge_set_sync_in_progress(FALSE);
}
```

src/Job/UserUpdate.php
```diff
@@ -0,0 +1,114 @@
...
+        }
+      }
+    }
+
+    if ($this->executeUpdate) {
+      // If the developer-user synchronization is in progress, then saving
+      // developers while saving Drupal user should be avoided.
+      _apigee_edge_set_sync_in_progress(TRUE);
+      $account->save();
+      _apigee_edge_set_sync_in_progress(FALSE);
```
tamasd: Same as above with the try-finally block.

src/Job/UserUpdate.php
```diff
@@ -0,0 +1,114 @@
...
+      $this->executeUpdate = TRUE;
+    }
+
+    $user_fields_to_sync = \Drupal::config('apigee_edge.sync')->get('user_fields_to_sync');
+    if (!empty($user_fields_to_sync)) {
+      /** @var \Drupal\apigee_edge\FieldStorageFormatManager $format_manager */
+      $format_manager = \Drupal::service('plugin.manager.apigee_field_storage_format');
+      foreach ($user_fields_to_sync as $field) {
+        $type = $account->getFieldDefinition($field)->getType();
+        $formatter = $format_manager->lookupPluginForFieldType($type);
```
tamasd: Same as above, a null check on the `$formatter` variable.

tamasd (changes requested):

src/Entity/Developer.php
peterserfozo: Fixed.

src/Job/DeveloperSync.php
peterserfozo: Fixed. Update the Drupal user based on the developer.

apigee_edge.module
```diff
@@ -962,6 +962,12 @@ function template_preprocess_app_credential(array &$variables) {
  * it has been in Edge.
  */
 function apigee_edge_user_presave(UserInterface $account) {
+  // If the developer-user synchronization is in progress, then saving
+  // developers while saving Drupal user should be avoided.
+  if (_apigee_edge_get_sync_in_progress()) {
```
cnovak: If the developer sync is in progress and we don't do run this logic, it would mean that the user would not be created in Edge when registering an account in the portal.  This would cause the user to see errors until the developer is synchronized.  Wouldn't we need to add a batch of create user requests to run after sync is done?  This may still cause odd errors if the user is logged in immediately after login and the sync takes a long time.

src/Entity/Developer.php
```diff
@@ -105,11 +105,19 @@ public static function createFromDrupalUser(UserInterface $user): Developer {
...
     $developer->setOwnerId($user->id());
 
-    foreach (\Drupal::config('apigee_edge.sync')->get('user_fields_to_sync') as $field_to_sync) {
-      $developer->setAttribute(
-        preg_replace('/^field_(.*)$/', '${1}', $field_to_sync),
-        Json::encode($user->get($field_to_sync)->getValue())
-      );
+    /** @var \Drupal\apigee_edge\FieldStorageFormatManager $format_manager */
+    $format_manager = \Drupal::service('plugin.manager.apigee_field_storage_format');
+    foreach (\Drupal::config('apigee_edge.sync')->get('user_fields_to_sync') as $field) {
```
mxr576: Use `$field_name` instead because it more self describing.

src/Entity/Developer.php
```diff
@@ -105,11 +105,19 @@ public static function createFromDrupalUser(UserInterface $user): Developer {
...
+      $type = $user->getFieldDefinition($field)->getType();
```
mxr576: Use `$field_type` instead because it more self describing.

src/Form/DeveloperAttributesSettingsForm.php
```diff
@@ -72,12 +75,26 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+        'attribute_name' => static::getAttributeName($field->getName()),
       ];
+      $formatter = $format_manager->lookupPluginForFieldType($field->getType());
+      if (isset($formatter)) {
+        $rc = new \ReflectionClass($format_manager->lookupPluginForFieldType($field->getType()));
+        $short_name = $rc->getShortName();
+        $options[$field->getName()]['field_storage_formatter'] = $short_name;
+      }
+      else {
+        $options[$field->getName()]['field_storage_formatter'] = '-';
```
mxr576: Meh, there is no "#empty-option" for table select, isn't it? What happens if you do not set this array key? Maybe it should be `t('- None -')` if we have to set this key's value.

src/Form/DeveloperSyncForm.php
```diff
@@ -39,11 +67,22 @@ public function getFormId() {
    * {@inheritdoc}
    */
   public function buildForm(array $form, FormStateInterface $form_state) {
+
```
mxr576: Unnecessary new line.

src/Job/DeveloperSync.php
```diff
@@ -130,21 +131,42 @@ protected function executeRequest() {
...
    * {@inheritdoc}
    */
-  public function execute() : bool {
+  public function execute(): bool {
     parent::execute();
 
+    // Update Apigee Edge developers and Drupal users if needed.
+    $identical_accounts = array_intersect($this->edgeAccounts, $this->drupalAccounts);
+    foreach ($identical_accounts as $search => $mail) {
+      /** @var \Apigee\Edge\Api\Management\Entity\DeveloperInterface $developer */
```
mxr576: It is a Drupal developer entity and not an SDK developer entity.

src/Job/DeveloperSync.php
```diff
@@ -130,21 +131,42 @@ protected function executeRequest() {
...
+      $developer = Developer::load($mail);
```
mxr576: I'd probably load all developers by email outside of this scope instead of loading them one by one in a loop.

src/Job/DeveloperSync.php
```diff
@@ -130,21 +131,42 @@ protected function executeRequest() {
...
+      /** @var \Drupal\user\UserInterface $account */
+      $account = user_load_by_mail($mail);
```
mxr576: The same of users objects, fetch their IDs by mail and then load all of them before this loop.

src/Job/DeveloperSync.php
```diff
@@ -130,21 +131,42 @@ protected function executeRequest() {
...
       }
     }
 
+    // Create missing Apigee Edge developers.
     foreach ($this->drupalAccounts as $search => $mail) {
       if (empty($this->edgeAccounts[$search])) {
-        $jobs = new JobList(TRUE);
-        /** @var \Drupal\user\Entity\User $account */
+        /** @var \Drupal\user\UserInterface $account */
         if (!($account = user_load_by_mail($mail))) {
```
mxr576: If all users loaded to an array above you do not need to load one by email here probably, you can find it in the array if it is indexed by email.

src/Job/DeveloperUpdate.php
```diff
@@ -19,41 +19,103 @@
...
+    if ($developer->getLastName() !== $account->get('last_name')->value) {
+      $developer->setLastName($account->get('last_name')->value);
+      $this->executeUpdate = TRUE;
+    }
+
+    $user_fields_to_sync = \Drupal::config('apigee_edge.sync')->get('user_fields_to_sync');
+    if (!empty($user_fields_to_sync)) {
+      /** @var \Drupal\apigee_edge\FieldStorageFormatManager $format_manager */
+      $format_manager = \Drupal::service('plugin.manager.apigee_field_storage_format');
+      foreach ($user_fields_to_sync as $field) {
```
mxr576: Use `$field_name` instead because it more self describing.

src/Job/DeveloperUpdate.php
```diff
@@ -19,41 +19,103 @@
...
+        $field_definition = $account->getFieldDefinition($field);
+        if (!isset($field_definition)) {
+          $this->recordMessage(t('Skipping @mail developer update, because the field @field does not exist.', [
```
mxr576: s/because the field @field does/because %field field does/g

src/Job/DeveloperUpdate.php
```diff
@@ -19,41 +19,103 @@
...
+            '@mail' => $this->mail,
+            '@field' => $field_definition->getName(),
+          ])->render());
+          continue;
+        }
+        $type = $field_definition->getType();
```
mxr576: Use `$field_type` instead because it more self describing.

src/Job/DeveloperUpdate.php
mxr576: Are you skipping a developer entity _completely_ just because one field does not exist? I think you are not. Just the displayed message is incorrect. You are skipping synching a field value of an entity because field does not exist.

src/Job/DeveloperUpdate.php
```diff
@@ -19,41 +19,103 @@
...
+        $formatter = $format_manager->lookupPluginForFieldType($type);
+        if (!isset($formatter)) {
+          $this->recordMessage(t('Skipping @mail developer update, because there is no available storage formatter for @field_type.', [
```
mxr576: The same as above, I think this message is incorrect and does not reflect what actually happens.

src/Job/UserUpdate.php
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    $user_fields_to_sync = \Drupal::config('apigee_edge.sync')->get('user_fields_to_sync');
+    if (!empty($user_fields_to_sync)) {
+      /** @var \Drupal\apigee_edge\FieldStorageFormatManager $format_manager */
+      $format_manager = \Drupal::service('plugin.manager.apigee_field_storage_format');
+      foreach ($user_fields_to_sync as $field) {
+        $field_definition = $account->getFieldDefinition($field);
+        if (!isset($field_definition)) {
+          $this->recordMessage(t('Skipping @mail developer update, because the field @field does not exist.', [
```
mxr576: I would add the same comments here as I did for DeveloperUpdate.

src/Job/UserUpdate.php
```diff
@@ -0,0 +1,133 @@
...
+            '@mail' => $this->mail,
+            '@field' => $field_definition->getName(),
+          ])->render());
+          continue;
+        }
+        $type = $field_definition->getType();
```
mxr576: field_type

src/Job/UserUpdate.php
```diff
@@ -0,0 +1,133 @@
...
+        $formatter = $format_manager->lookupPluginForFieldType($type);
+        if (!isset($formatter)) {
+          $this->recordMessage(t('Skipping @mail developer update, because there is no available storage formatter for @field_type.', [
```
mxr576: I would add the same comments here as I did for DeveloperUpdate.

src/Job/UserUpdate.php
```diff
@@ -0,0 +1,133 @@
...
+    if ($developer->getLastName() !== $account->get('last_name')->value) {
+      $account->set('last_name', $developer->getLastName());
+      $this->executeUpdate = TRUE;
+    }
+
+    $user_fields_to_sync = \Drupal::config('apigee_edge.sync')->get('user_fields_to_sync');
+    if (!empty($user_fields_to_sync)) {
+      /** @var \Drupal\apigee_edge\FieldStorageFormatManager $format_manager */
+      $format_manager = \Drupal::service('plugin.manager.apigee_field_storage_format');
+      foreach ($user_fields_to_sync as $field) {
```
mxr576: field_name

tests/src/Functional/DeveloperSyncTest.php
```diff
@@ -0,0 +1,216 @@
...
+        'email' => "{$this->prefix}.{$this->randomMachineName()}@example.com",
+        'userName' => $this->randomMachineName(),
+        'firstName' => $this->randomMachineName(),
+        'lastName' => $this->randomMachineName(),
+      ]);
+      $this->edgeDevelopers[$i]->save();
+    }
+
+    // Create users in Drupal.
+    _apigee_edge_set_sync_in_progress(TRUE);
```
mxr576: Add a comment that you are doing this because you do not want to create developer on Apigee Edge.

tests/src/Functional/DeveloperSyncTest.php
```diff
@@ -0,0 +1,216 @@
...
+
+  /**
+   * Verifies that the Drupal users and the Edge developers are synchronized.
+   */
+  protected function verify() {
+    $all_users = [];
+    /** @var \Drupal\user\UserInterface $account */
+    foreach (User::loadMultiple() as $account) {
+      $email = $account->getEmail();
+      if ($email && $email !== 'admin@example.com') {
```
mxr576: Document why we skip user with this email address.

tests/src/Functional/DeveloperSyncTest.php
```diff
@@ -0,0 +1,216 @@
...
+        $this->assertTrue($this->filter ? (bool) preg_match($this->filter, $email) : TRUE, "Email ({$email}) is filtered properly.");
```
mxr576: Don't we need a proper error message to be display if this assert fails? (Second parameter of method).

tests/src/Functional/DeveloperSyncTest.php
```diff
@@ -0,0 +1,216 @@
...
+        $all_users[$email] = $email;
+      }
+    }
+
+    unset($all_users[$this->rootUser->getEmail()]);
+
+    foreach ($this->edgeDevelopers as $edgeDeveloper) {
+      /** @var \Drupal\user\Entity\User $account */
+      $user = user_load_by_mail($edgeDeveloper->getEmail());
```
mxr576: Load all of them at once, do not load them one by one in a loop.

tests/src/Functional/DeveloperSyncTest.php
```diff
@@ -0,0 +1,216 @@
...
+      $this->assertEquals($edgeDeveloper->getUserName(), $user->getAccountName());
+      $this->assertEquals($edgeDeveloper->getFirstName(), $user->get('first_name')->value);
+      $this->assertEquals($edgeDeveloper->getLastName(), $user->get('last_name')->value);
+
+      unset($all_users[$edgeDeveloper->getEmail()]);
+    }
+
+    foreach ($this->drupalUsers as $drupalUser) {
+      /** @var \Drupal\apigee_edge\Entity\DeveloperInterface $developer */
+      $developer = Developer::load($drupalUser->getEmail());
```
mxr576: Load all of them at once, do not load them one by one in a loop.

tests/src/Functional/DeveloperSyncTest.php
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  /**
+   * Tests Drupal user synchronization.
+   *
+   * @throws \Behat\Mink\Exception\ResponseTextException
+   */
+  public function testUserSync() {
+    $this->drupalGet('/admin/config/apigee-edge/developer-settings/sync');
+    $this->clickLinkProperly(t('Now'));
```
mxr576: Either you or @cnovak will have to update your/his PR. It depends which one gets merged first. :) (The label of the button gets changed in Chris's PR #36)

tests/src/Functional/DeveloperSyncTest.php
```diff
@@ -0,0 +1,216 @@
...
+    $this->assertSession()->pageTextContains(t('Users are in sync with Edge.'));
+    $this->verify();
+  }
+
+  /**
+   * Tests scheduled Drupal user synchronization.
+   *
+   * @throws \Behat\Mink\Exception\ResponseTextException
```
mxr576: Do not document these exceptions, it suppresses PHPStorm's built-in warnings but documenting these are basically useless for developers.

tests/src/Functional/DeveloperSyncTest.php
```diff
@@ -0,0 +1,216 @@
...
+   * @throws \Drupal\Component\Plugin\Exception\PluginException
+   * @throws \Exception
+   */
+  public function testUserAsync() {
+    $this->drupalGet('/admin/config/apigee-edge/developer-settings/sync');
+    $this->clickLinkProperly(t('Background'));
```
mxr576: Same as above, one of the open PR needs to be updated. :) @cnovak

tests/src/Functional/DeveloperSyncTest.php
```diff
@@ -0,0 +1,216 @@
...
+    $this->assertSession()->pageTextContains(t('User synchronization is scheduled.'));
+    /** @var \Drupal\Core\Queue\QueueFactory $queue_service */
+    $queue_service = \Drupal::service('queue');
```
mxr576: Use `$this->container->get()`.

tests/src/Functional/DeveloperSyncTest.php
```diff
@@ -0,0 +1,216 @@
...
+  /**
+   * Tests the Drupal user synchronization started from the CLI.
+   */
+  public function testCliUserSync() {
+    $cli_service = $this->container->get('apigee_edge.cli');
+    $input = new ArgvInput();
+    $output = new BufferedOutput();
+
+    $cli_service->sync(new SymfonyStyle($input, $output), 't');
+
```
mxr576: Too many (unnecessary) empty lines :)

mxr576 (changes requested):

apigee_edge.module
```diff
@@ -1311,3 +1317,33 @@ function _apigee_edge_user_has_an_app_with_product(string $productName, AccountI
...
+  return $i;
+}
+
+/**
+ * Gets the user synchronization state.
+ *
+ * @return bool|null
+ *   TRUE if the user synchronization is in progress, else FALSE or NULL.
+ */
+function _apigee_edge_get_sync_in_progress() {
```
mxr576: Probably it should always return a boolean value at least I do not have an example for when we would need to distinguish NULL from FALSE. I also think it should be called as `_apigee_edge_is_sync_in_progress()` because it returns a boolean value.

mxr576 (changes requested):

apigee_edge.module
```diff
@@ -1321,3 +1327,34 @@ function _apigee_edge_user_has_an_app_with_product(string $productName, AccountI
...
+ * based on a developer should be avoided.
+ *
+ * @param bool|null $in_progress
+ *   Developer-user synchronization state.
+ *
+ * @return bool
+ *   TRUE if the developer-user synchronization is in progress, else FALSE.
+ */
+function _apigee_edge_set_sync_in_progress(?bool $in_progress = NULL) {
+  static $i;
```
mxr576: Little nitpicking, but `$i` is never meaningful enough as a variable name, not even in a 5 lines of function. Please rename it to `$state` for example.

src/Entity/Developer.php
```diff
@@ -105,11 +105,32 @@ public static function createFromDrupalUser(UserInterface $user): Developer {
...
-    foreach (\Drupal::config('apigee_edge.sync')->get('user_fields_to_sync') as $field_to_sync) {
-      $developer->setAttribute(
-        preg_replace('/^field_(.*)$/', '${1}', $field_to_sync),
-        Json::encode($user->get($field_to_sync)->getValue())
-      );
+    /** @var \Drupal\apigee_edge\FieldStorageFormatManager $format_manager */
+    $format_manager = \Drupal::service('plugin.manager.apigee_field_storage_format');
+    foreach (\Drupal::config('apigee_edge.sync')->get('user_fields_to_sync') as $field_name) {
+      $field_definition = $user->getFieldDefinition($field_name);
+      // If the field does not exist, then skip.
```
mxr576: s/skip/skip it/g

src/Entity/Developer.php
```diff
@@ -105,11 +105,32 @@ public static function createFromDrupalUser(UserInterface $user): Developer {
...
+          '%field_name' => $field_name,
+        ]);
+        continue;
+      }
+      $field_type = $field_definition->getType();
+      $formatter = $format_manager->lookupPluginForFieldType($field_type);
+      // If there is no available storage formatter for the field, then skip.
+      if (!isset($formatter)) {
+        \Drupal::logger('apigee_edge')->warning("Skipping %mail developer's %attribute_name attribute update, because there is no available storage formatter for %field_type field type.", [
+          '%mail' => $user->getEmail(),
```
mxr576: Because you have the user object here you can easily add a link to its user profile page which is rendered in Operations column of recent log messages. Use `link` key in the context.

src/Job/DeveloperSync.php
```diff
@@ -59,128 +58,128 @@ class DeveloperSync extends EdgeJob {
...
-    foreach ($mails as $mail) {
-      if (isset($mail)) {
-        $accounts[strtolower($mail)] = $mail;
+  protected function loadUsers(): array {
+    $users = [];
+    /** @var \Drupal\user\UserInterface $user */
+    foreach (User::loadMultiple() as $user) {
+      $email = $user->getEmail();
+      if (isset($email)) {
+        $users[strtolower($email)] = $user;
```
mxr576: Let's leave a reminder here that we did `strtolower($email)` because of this: "https://www.drupal.org/project/drupal/issues/2490294".

src/Job/DeveloperSync.php
```diff
@@ -59,128 +58,128 @@ class DeveloperSync extends EdgeJob {
...
+    $developers = [];
+    /** @var \Drupal\apigee_edge\Entity\DeveloperInterface $developer */
+    foreach (Developer::loadMultiple() as $developer) {
+      $email = $developer->getEmail();
+      if ($this->filter && !preg_match($this->filter, $email)) {
         continue;
       }
-      $accounts[strtolower($mail)] = $mail;
+      else {
+        $developers[strtolower($email)] = $developer;
```
mxr576, line 118: Let's leave a reminder here that we did strtolower($email) because of this: "https://www.drupal.org/project/drupal/issues/2490294".

src/Job/DeveloperSync.php
```diff
@@ -59,128 +58,128 @@ class DeveloperSync extends EdgeJob {
...
+      $user = $this->drupalUsers[$clean_email];
+
+      $last_modified_delta = $developer->getLastModifiedAt()->getTimestamp() - $user->getChangedTime();
+      // Update Drupal user because the Apigee Edge developer is fresher.
+      if ($last_modified_delta >= 0) {
+        $updateUserJob = new UserUpdate($user->getEmail());
+        $updateUserJob->setTag($this->getTag());
+        $this->scheduleJob($updateUserJob);
+      }
+      // Update Apigee Edge developer because the Drupal user is fresher.
```
mxr576: https://www.urbandictionary.com/define.php?term=fresher ? :) Is "more recent" for example.

src/Job/UserCreate.php
```diff
@@ -0,0 +1,157 @@
...
+      }
+    }
+
+    $user_fields_to_sync = \Drupal::config('apigee_edge.sync')->get('user_fields_to_sync');
+    if (!empty($user_fields_to_sync)) {
+      /** @var \Drupal\apigee_edge\FieldStorageFormatManager $format_manager */
+      $format_manager = \Drupal::service('plugin.manager.apigee_field_storage_format');
+      foreach ($user_fields_to_sync as $field_name) {
+        if (!array_key_exists(static::getAttributeName($field_name), $developer->getAttributes()->values())) {
+          $this->recordMessage(t('Skipping creating %email user field, because the developer does not have %attribute_name attribute on Apigee Edge.', [
```
mxr576: We should log these types of errors with logger too.

src/Job/UserCreate.php
```diff
@@ -0,0 +1,157 @@
...
+          ])->render());
+          continue;
+        }
+
+        $rollback = $user->get($field_name)->getValue();
+        $user->set($field_name, $formatter->decode($developer->getAttributeValue(static::getAttributeName($field_name))));
+        // Do not set the field value if a field constraint fails during
+        // validation.
+        $field_violations = $user->get($field_name)->validate();
+        if ($field_violations->count() > 0) {
```
mxr576, line 130: Awesome! :)

tests/src/Functional/DeveloperSyncTest.php
mxr576: ^

tests/src/Functional/DeveloperSyncTest.php
mxr576: ^

tests/src/Functional/DeveloperSyncTest.php
mxr576: ^

tests/src/Functional/DeveloperSyncTest.php
mxr576: ^

mxr576 (changes requested):

apigee_edge.module
tamasd: Also, by convention `$i` supposed to mean loop index.

tamasd (approved):

apigee_edge.module
peterserfozo: PHP does not preserve the application state between requests, so a static variable lives only through one request. (see `_apigee_edge_set_sync_in_progress()` and `_apigee_edge_is_sync_in_progress()` functions)

src/Job/DeveloperSync.php
mxr576: @peterserfozo  ^ another todo

tests/src/Functional/DeveloperSyncTest.php
mxr576: @peterserfozo  ping ^

mxr576 (approved):

PR #36, opened 2018-06-30
Copy changes for general settings

cnovak:
* Copy changes to general settings
* Change default name of “developer apps” to “apps”
* “Apigee Edge” should not be shortened to “Edge”
* Remove custom styling of developer sync page, change to match cron form
* Removed “please” from copy
* Page titles should only have first work capitalized

### DIFF
--- a/src/Form/DeveloperSyncForm.php
+++ b/src/Form/DeveloperSyncForm.php
@@ -46,11 +46,14 @@ class DeveloperSyncForm extends FormBase {
       '#title' => $this->t('Sync developers'),
       '#open' => TRUE,
     ];
-
+    
     $form['sync']['description'] = [
-      '#markup' => '<p>' . $this->t('Developer synchronization will run through all users in this portal, adding them as developers in the Apigee Edge org, and making sure all developers on the Apigee Edge org are added to this portal. The "Run Developer Sync" button will sync the developers, displaying a progress bar. The "Background Developer Sync" button will run the developer sync process in batches each time <a href=":cron_url">cron</a> runs.', [
-        ':cron_url' => Url::fromRoute('system.cron_settings')->toString()
-        ]) . '</p>',
+      '#type' => 'html_tag',
+      '#tag' => 'p',
+      '#value' => $this
+        ->t('Developer synchronization will run through all users in this portal, adding them as developers in the Apigee Edge org, and making sure all developers on the Apigee Edge org are added to this portal. The "Run Developer Sync" button will sync the developers, displaying a progress bar. The "Background Developer Sync" button will run the developer sync process in batches each time <a href=":cron_url">cron</a> runs.', [
+          ':cron_url' => Url::fromRoute('system.cron_settings')->toString()
+        ]),
     ];
 
     $form['sync']['sync_submit'] = [
--- a/src/Form/DeveloperSyncForm.php
+++ b/src/Form/DeveloperSyncForm.php
@@ -46,14 +46,13 @@ class DeveloperSyncForm extends FormBase {
       '#title' => $this->t('Sync developers'),
       '#open' => TRUE,
     ];
-    
+
     $form['sync']['description'] = [
       '#type' => 'html_tag',
       '#tag' => 'p',
-      '#value' => $this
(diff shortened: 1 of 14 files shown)


src/Form/DeveloperSyncForm.php
```diff
@@ -46,8 +46,15 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       '#title' => $this->t('Sync developers'),
       '#open' => TRUE,
     ];
+
+    $form['sync']['description'] = [
+      '#markup' => '<p>' . $this->t('Developer synchronization will run through all users in this portal, adding them as developers in the Apigee Edge org, and making sure all developers on the Apigee Edge org are added to this portal. The "Run Developer Sync" button will sync the developers, displaying a progress bar. The "Background Developer Sync" button will run the developer sync process in batches each time <a href=":cron_url">cron</a> runs.', [
```
mxr576: Please use the `#html_tag` render element instead if this really needs to wrapped in `<p>` tag. Is it?

https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21HtmlTag.php/class/HtmlTag/8.5.x

mxr576 (changes requested):

cnovak:
@mxr576 I followed the same pattern for description that the [Cron form](https://api.drupal.org/api/drupal/core%21modules%21system%21src%21Form%21CronForm.php/class/CronForm/8.5.x) uses: 

```
 /**
   * {@inheritdoc}
   */
    $form['description'] = [
      '#markup' => '<p>' . t('Cron takes care of running periodic tasks like checking for updates and indexing content for search.') . '</p>',
    ];
    $form['run'] = [
      '#type' => 'submit',
      '#value' => t('Run cron'),
      '#submit' => ['::runCron'],
    ];
```

The code and other forms are following this format, so if we don't put description in <p> the page may style differently than other pages.  Let me know if you still think we should change this code to use #html_tag render element. I think we should stay with the pattern other admin forms are using.

cnovak:
Talked to @mxr576 , will still wrap in <p> like other modules, but use #html_tag render element for more flexibility.

peterserfozo (approved):

src/Form/DeveloperSyncForm.php
```diff
@@ -46,8 +46,18 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       '#title' => $this->t('Sync developers'),
       '#open' => TRUE,
     ];
+    
+    $form['sync']['description'] = [
+      '#type' => 'html_tag',
+      '#tag' => 'p',
+      '#value' => $this
```
mxr576: Little nitpicking, but single method call should be in the same line, parameter list could be in multiple lines if it is necessary, otherwise, PR looks fine.

mxr576:
@cnovak Please monitor your Travis builds after you created a new PR every time. It was quite clear that your code changes introduced a new a coding style violation: https://travis-ci.org/cnovak/apigee-edge-drupal/builds/399633201

As a general practice I'd like to ask all developers to submit their Travis build in the URL of their PR so we could easily check whether their build is passing or not. The apigee/apigee-edge-drupal Travis CI is only checking code style, but it does not run tests with active Apigee Edge connection.

mxr576:
@cnovak Some tests are starting to get failed on your Travis after I have fixed the code style: https://travis-ci.org/cnovak/apigee-edge-drupal/jobs/399868817#L1173

mxr576 (changes requested):
Failing tests on Travis should be fixed.

cnovak:
I ran the build again after changing connection settings to fix the failed build, and still getting errors:

```
Caused by
Apigee\Edge\Exception\ClientErrorException: Forbidden
```
Not much info to go on.  What needs to be done to troubleshoot this error?

mxr576:
@cnovak would you like to update this PR or can we close this?

cnovak:
I need to finish this in this sprint.  Still having an issue where password is getting modified to something different using environment variables.  I debugged locally and could see it was being changed but did not have a chance to figure out why.

peterserfozo:
@cnovak It seems your credentials are wrong in Travis. Try to delete APIGEE_EDGE_USERNAME and APIGEE_EDGE_PASSWORD (https://travis-ci.org/cnovak/apigee-edge-drupal/settings) and set the values again for these environment variables surrounded by double quotes (`"`) then restart the latest build.

cnovak:
Going to resubmit this PR after merging

PR #40, opened 2018-07-09
Store client config in State

peterserfozo:


apigee_edge.install
```diff
@@ -87,6 +87,15 @@ function apigee_edge_install() {
     ];
     user_role_grant_permissions(RoleInterface::AUTHENTICATED_ID, $authenticated_user_permissions);
   }
+
+  // Set default client configuration.
+  \Drupal::state()->setMultiple([
```
mxr576: Why do not you use one `apigee_edge.client` state key (name) and serialize all these configurations under it? Currently you create multiple entries to store all client configs.

mxr576 (approved):

tamasd (approved):

PR #44, opened 2018-07-16
Pronovix hackathon fixes (part 2)

peterserfozo:


src/Controller/DeveloperAppListBuilderForDeveloper.php
```diff
@@ -76,7 +82,8 @@ public static function create(ContainerInterface $container) {
       $container->get('entity.manager')->getStorage($entity_type->id()),
       $container->get('entity.manager'),
       $container->get('renderer'),
-      $container->get('current_user')
+      $container->get('current_user'),
+      $container->get('request_stack')
```
mxr576, line 86: Why do you need request_stack?

src/Controller/DeveloperAppViewController.php
```diff
@@ -47,7 +48,7 @@ public function view(EntityInterface $developer_app, $view_mode = 'full') {
    */
   public function getPageTitle(RouteMatchInterface $routeMatch): string {
     return t('@name @developer_app', [
-      '@name' => $routeMatch->getParameter('developer_app')->getDisplayName(),
+      '@name' => Markup::create($routeMatch->getParameter('developer_app')->getDisplayName()),
```
mxr576: Please add a comment to your code or to the git commit message about why Markup::create() were the perfect solution for this double-escaping problem.

https://github.com/drupal/core/blob/10b7c918ab56f9b6b14ed52ed0afd2ab66f4b927/lib/Drupal/Component/Render/FormattableMarkup.php#L140

src/Entity/ListBuilder/DeveloperAppListBuilder.php
```diff
@@ -219,86 +260,119 @@ public function buildHeader() {
...
+   * products can be returned.
+   *
+   * @param \Drupal\apigee_edge\Entity\DeveloperAppInterface $developer_app
+   *   Developer app entity to be checked.
+   *
+   * @return array
+   *   An array containing the information about the credentials, products for
+   *   displaying warning messages.
+   */
+  protected function checkDeveloperApp(DeveloperAppInterface $developer_app): array {
```
mxr576: I'd rather call it `checkDeveloperAppCredentials()` instead.

src/Entity/ListBuilder/DeveloperAppListBuilder.php
```diff
@@ -219,86 +260,119 @@ public function buildHeader() {
...
-    $hasPendingCredProduct = FALSE;
-    $problematicApiProductName = NULL;
-    foreach ($entity->getCredentials() as $credential) {
+    $row['data'] += parent::buildRow($developer_app);
+    $rows[$infoRowId] = $row;
+    return $rows;
+  }
+
+  /**
+   * Checks the developer app and returns information for displaying warnings.
```
mxr576: Checks credentials of a developer app and returns warnings about them.

src/Entity/ListBuilder/DeveloperAppListBuilder.php
```diff
@@ -219,86 +260,119 @@ public function buildHeader() {
...
+   *
+   * Helper function for buildWarningRow(). Information about credentials,
```
mxr576: Second sentence does not add new information to this doc block so it can be removed.

src/Entity/ListBuilder/DeveloperAppListBuilder.php
```diff
@@ -219,86 +260,119 @@ public function buildHeader() {
...
+   * products can be returned.
+   *
+   * @param \Drupal\apigee_edge\Entity\DeveloperAppInterface $developer_app
+   *   Developer app entity to be checked.
+   *
+   * @return array
+   *   An array containing the information about the credentials, products for
```
mxr576: An array containing the information about the revoked credentials and revoked or pending products in a credential.

src/Entity/ListBuilder/DeveloperAppListBuilder.php
```diff
@@ -219,86 +260,119 @@ public function buildHeader() {
...
-     */
-    if ($entity->getStatus() === App::STATUS_APPROVED && ($hasRevokedCred || $hasPendingCredProduct || $hasRevokedCredProduct)) {
-      $build['status'] = $infoRow['app_status']['data'];
+    return $info;
+  }
+
+  /**
+   * Builds a warning row for a developer app in the entity listing.
+   *
+   * The warning row contains the warning messages if necessary.
```
mxr576: s/necessary/there is any/g.

src/Entity/ListBuilder/DeveloperAppListBuilder.php
```diff
@@ -219,86 +260,119 @@ public function buildHeader() {
...
+   *
+   * @param \Drupal\apigee_edge\Entity\DeveloperAppInterface $developer_app
+   *   The developer app entity for this row of the list.
+   * @param array $rows
+   *   A render array containing the info and warning rows to be displayed.
+   *
+   * @return array
+   *   A render array structure of fields for this developer app entity.
+   */
+  protected function buildWarningRow(DeveloperAppInterface $developer_app, array $rows): array {
```
mxr576: Why do you pass `$rows` if you only use it for one purpose, adding a new warning row to it if there were warning  and then return it?
You should either return only the warning row in this method (or an empty array for example) or pass $rows by reference and do not return anything in this method. I'd prefer the first version.

src/Entity/ListBuilder/DeveloperAppListBuilder.php
```diff
@@ -320,62 +394,66 @@ public function buildRow(EntityInterface $entity) {
...
           /** @var \Drupal\apigee_edge\Entity\ApiProductInterface $apiProduct */
-          $apiProduct = $this->getApiProductStorage()->load($problematicApiProductName);
+          $apiProduct = $this->getApiProductStorage()->load($info['problematicApiProductName']);
           $args['%name'] = $apiProduct->getDisplayName();
-          $warningRow['info']['data'] = $this->t("%name @apiproduct associated with this @developer_app is in @status status.", $args);
+          $row['data']['info']['data'] = $this->t("%name @apiproduct associated with this @developer_app is in @status status.", $args);
         }
         else {
-          $warningRow['info']['data'] = $this->t("At least one @apiproduct associated with one of the credentials of this @developer_app is in @status status.", $args);
+          $row['data']['info']['data'] = $this->t("At least one @apiproduct associated with one of the credentials of this @developer_app is in @status status.", $args);
```
mxr576: Single quote

src/Entity/ListBuilder/DeveloperAppListBuilder.php
```diff
@@ -320,62 +394,66 @@ public function buildRow(EntityInterface $entity) {
...
+          '@status' => $info['hasPendingCredProduct'] ? $this->t('pending') : $this->t('revoked'),
         ];
-        if (count($entity->getCredentials()) === 1) {
+        if (count($developer_app->getCredentials()) === 1) {
           /** @var \Drupal\apigee_edge\Entity\ApiProductInterface $apiProduct */
-          $apiProduct = $this->getApiProductStorage()->load($problematicApiProductName);
+          $apiProduct = $this->getApiProductStorage()->load($info['problematicApiProductName']);
           $args['%name'] = $apiProduct->getDisplayName();
-          $warningRow['info']['data'] = $this->t("%name @apiproduct associated with this @developer_app is in @status status.", $args);
+          $row['data']['info']['data'] = $this->t("%name @apiproduct associated with this @developer_app is in @status status.", $args);
```
mxr576: Single quoute.

src/Entity/ListBuilder/DeveloperAppListBuilder.php
```diff
@@ -219,86 +260,119 @@ public function buildHeader() {
...
+   *   A render array containing the info and warning rows to be displayed.
+   *
+   * @return array
+   *   A render array structure of fields for this developer app entity.
    */
-  public function buildRow(EntityInterface $entity) {
-    /** @var \Drupal\apigee_edge\Entity\DeveloperAppInterface $entity */
-    $request = \Drupal::request();
-    $appNameAsCssId = $this->getUniqueCssIdForApp($entity);
+  protected function buildInfoRow(DeveloperAppInterface $developer_app, array $rows): array {
```
mxr576: The same comment as for the buildWarningRow() regarding $rows.

mxr576 (changes requested):

src/Controller/DeveloperAppListBuilderForDeveloper.php
peterserfozo: Because of JS. As I see it uses the current URI and app ID to make difference between toggle buttons. I've replaced the global calling of the service with dependency injection.

peterserfozo:
Test: https://travis-ci.org/peterserfozo/apigee-edge-drupal/builds/405811113

src/Entity/ListBuilder/DeveloperAppListBuilder.php
```diff
@@ -320,62 +400,41 @@ public function buildRow(EntityInterface $entity) {
...
-      '@developer_app' => $this->getDeveloperAppEntityDefinition()->getLowercaseLabel(),
-    ]), 'entity.developer_app.add_form', [], ['attributes' => ['class' => 'btn btn-primary btn--add-app']])->toRenderable();
+  public function buildRow(EntityInterface $entity) {
+    /** @var \Drupal\apigee_edge\Entity\DeveloperAppInterface $entity */
+    $rows = [];
+    $appNameAsCssId = $this->getUniqueCssIdForApp($entity);
+
+    // Generate info row of the current developer app.
+    $infoRowId = "{$appNameAsCssId}-info";
+    $rows[$infoRowId] = $this->buildInfoRow($entity);
```
mxr576: Why don't you just pass `$rows` as reference to both `buildInfoRow()` and `buildWarningRow()` and let them to alter the structure of rows as they want? `$infoRowId` could be calculated in for an app in a dedicated method, same for the `$warningRowId`. You would not need to pass `$rows[$infoRowId]` to the `buildWarningRow()` just the `&$rows` because you could identify the related info row by calculating the info row id for the $entity with the dedicated method.
Consequently none of these method would need to return anything, they would just directly alter the $rows if they want.

mxr576 (changes requested):

mxr576 (approved):

peterserfozo:
Test: https://travis-ci.org/peterserfozo/apigee-edge-drupal/builds/406167736

PR #52, opened 2018-07-19
Detailed auth form messages, debug element

peterserfozo:


src/Form/AuthenticationForm.php
```diff
@@ -372,26 +392,187 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
+      $form_state->setError($form, $this->t('@suggestion Error message: %response.', [
+        // Use Markup instead of primitive string to avoid double
+        // sanitizing.
+        '@suggestion' => Markup::create($this->createSuggestion($exception, $key)),
+        '%response' => $exception->getMessage(),
+      ]));
+
+      // Display debug information.
+      $form['debug']['#access'] = TRUE;
+      $form['debug']['debug_text']['#access'] = TRUE;
```
mxr576: Debug text does not have `#access` set to FALSE according to your code so this line should be unnecessary.

src/Form/AuthenticationForm.php
```diff
@@ -372,26 +392,187 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
-        $form_state->setError($form, $this->t('Connection failed. Response from Apigee Edge: %response', [
-          '%response' => $exception->getMessage(),
-        ]));
+      // Failed request.
+      elseif ($exception->getCode() === 0) {
+        if ($exception->getPrevious() instanceof ApiRequestException && $exception->getPrevious()->getPrevious() instanceof NetworkException && $exception->getPrevious()->getPrevious()->getPrevious() instanceof ConnectException) {
+          /** @var \GuzzleHttp\Exception\ConnectException $curl_exception */
+          $curl_exception = $exception->getPrevious()->getPrevious()->getPrevious();
+          // The remote host was not resolved (authorization server).
+          if ($curl_exception->getHandlerContext()['errno'] === 6) {
```
mxr576: Use `CURLE_COULDNT_RESOLVE_HOST` instead.

src/Form/AuthenticationForm.php
```diff
@@ -372,26 +392,187 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
+          '%organization' => $key_type->getOrganization($key),
+        ]);
+      }
+      // Failed request.
+      elseif ($exception->getCode() === 0) {
+        if ($exception->getPrevious() instanceof NetworkException && $exception->getPrevious()->getPrevious() instanceof ConnectException) {
+          /** @var \GuzzleHttp\Exception\ConnectException $curl_exception */
+          $curl_exception = $exception->getPrevious()->getPrevious();
+          // Resolving timed out.
+          if ($curl_exception->getHandlerContext()['errno'] === 28) {
```
mxr576: Use `CURLE_OPERATION_TIMEDOUT` instead.

src/Form/AuthenticationForm.php
```diff
@@ -372,26 +392,187 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
+            $suggestion = $this->t('@fail_text Maybe the connection timeout threshold (%connect_timeout) or the request timeout (%timeout) is too low or something is wrong with the connection.', [
+              '@fail_text' => $fail_text,
+              '%connect_timeout' => $this->state->get('apigee_edge.client')['http_client_connect_timeout'],
+              '%timeout' => $this->state->get('apigee_edge.client')['http_client_timeout'],
+            ]);
+          }
+          // The remote host was not resolved (endpoint).
+          elseif ($curl_exception->getHandlerContext()['errno'] === 6) {
```
mxr576: Use `CURLE_COULDNT_RESOLVE_HOST` instead.

src/Form/AuthenticationForm.php
```diff
@@ -372,26 +392,187 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
+    ];
+
+    $keys = [
+      'key_type' => get_class($key_type),
+      'key_provider' => get_class($key->getKeyProvider()),
+    ];
+
+    if ($key_type instanceof OauthKeyType) {
+      /** @var \Drupal\apigee_edge\Plugin\KeyType\OauthKeyType $key_type */
+      $credentials['authorization_server'] = $key_type->getAuthorizationServer($key) ?? Oauth::DEFAULT_AUTHORIZATION_SERVER;
```
mxr576: Why `getAuthorizationServer()` does not return `Oauth::DEFAULT_AUTHORIZATION_SERVER` if it has not been changed to something else? The same question for `getClientId()` and `getClientSecret()`. Would it be complicated to change the implementation? (IIRC we talked about something like this before)

mxr576 (changes requested):

src/Form/AuthenticationForm.php
```diff
@@ -372,26 +392,187 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
+          $suggestion = $this->t('@fail_text The given username (%username) or password or client ID (%client_id) or client secret is probably incorrect.', [
+            '@fail_text' => $fail_text,
+            '%client_id' => $key_type->getClientId($key) ?? Oauth::DEFAULT_CLIENT_ID,
+            '%username' => $key_type->getUsername($key),
+          ]);
+        }
+        // Invalid credentials using default client_id/client_secret.
+        else {
+          $suggestion = $this->t('@fail_text The given username (%username) or password is probably incorrect.', [
+            '@fail_text' => $fail_text,
```
mxr576: What if you would inject fail_text with `'!fail_text' => $fail_text->render()`? Using `Markup::create()` would be still necessary above?

src/Form/AuthenticationForm.php
peterserfozo: It has because `debug_text` is a child element of `debug`.

src/Form/AuthenticationForm.php
```diff
@@ -372,26 +392,198 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
+        else {
+          $suggestion = $this->t('@fail_text The given username (%username) or password is incorrect.', [
+            '@fail_text' => $fail_text,
+            '%username' => $key_type->getUsername($key),
+          ]);
+        }
+      }
+      // Failed request.
+      elseif ($exception->getCode() === 0) {
+        if ($exception->getMessage() === 'Host can not be empty.') {
```
mxr576: Message should be defined in a constant.

src/Form/AuthenticationForm.php
```diff
@@ -372,26 +392,198 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
+      $suggestion = $this->t('@fail_text Check the debug information below for more details.', [
+        '@fail_text' => $fail_text,
+      ]);
+      // Invalid credentials.
+      if ($exception->getCode() === 401) {
+        // Invalid credentials using defined client_id/client_secret.
+        if ($key_type->getClientId($key) !== Oauth::DEFAULT_CLIENT_ID || $key_type->getClientSecret($key) !== Oauth::DEFAULT_CLIENT_SECRET) {
+          $suggestion = $this->t('@fail_text The given username (%username) or password or client ID (%client_id) or client secret is incorrect.', [
+            '@fail_text' => $fail_text,
+            '%client_id' => $key_type->getClientId($key) ?? Oauth::DEFAULT_CLIENT_ID,
```
mxr576: getClientId() never returns null.

src/Form/AuthenticationForm.php
mxr576: Btw, isn't there anything else specific in the exception that you can use in the condition? I do not really prefer writing a condition on an message that could change easily.

mxr576 (changes requested):

tests/src/FunctionalJavascript/AuthenticationTest.php
```diff
@@ -0,0 +1,409 @@
...
+
+    // Create a key entity using key's default type and provider, then visit
+    // authentication form and ensure that there are no available Apigee Edge
+    // keys.
+    Key::create([
+      'id' => 'default_key',
+    ])->save();
+    $this->drupalGet(Url::fromRoute('apigee_edge.settings'));
+    $this->assertSession()->pageTextContains('There is no available basic authentication key for connecting to Apigee Edge.');
+    $this->assertSession()->pageTextNotContains('There is no available OAuth key for connecting to Apigee Edge.');
```
mxr576: Why do not you define these specific messages as constants in the test class at least? Copy-pasting the same message is not efficient.

tests/src/FunctionalJavascript/AuthenticationTest.php
```diff
@@ -0,0 +1,409 @@
...
+    $this->assertEmpty($this->container->get('state')->get('apigee_edge.auth')['active_key_oauth_token']);
+    $this->assertSession()->pageTextContains('There is no available basic authentication key for connecting to Apigee Edge.');
+    $this->getSession()->getPage()->selectFieldOption('edit-key-type', 'apigee_edge_oauth');
+    $this->assertSession()->pageTextContains('There is no available OAuth key for connecting to Apigee Edge.');
+    $this->assertSession()->pageTextContains('There is no available OAuth token key for connecting to Apigee Edge.');
+
+    // Check Key form custom validations.
+    $this->assertKeyFormValidation();
+
+    // Only Apigee Edge keys are usable in SDK connector.
```
mxr576: s/are usable/can be used/g

tests/src/FunctionalJavascript/AuthenticationTest.php
```diff
@@ -0,0 +1,409 @@
...
+    if (empty($message)) {
+      $this->assertSession()->pageTextContains('The configuration options have been saved');
+      $this->assertSession()->pageTextNotContains('Debug information');
+      $this->assertEquals($this->container->get('state')->get('apigee_edge.auth')['active_key'], $key_id);
+      $this->assertEquals($this->container->get('state')->get('apigee_edge.auth')['active_key_oauth_token'], $key_token_id);
+      $this->container->get('apigee_edge.sdk_connector')->testConnection();
+    }
+    else {
+      $this->assertSession()->pageTextContains($message);
+      $this->assertSession()->pageTextContains('Debug information');
```
mxr576: This text can also come from a constant in this test class.

mxr576 (approved):

peterserfozo:
Test: https://travis-ci.org/peterserfozo/apigee-edge-drupal/builds/407127955

PR #57, opened 2018-07-30
Ignore invalid callback URL

peterserfozo:


src/Controller/DeveloperAppCallbackUrlCheckTrait.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+namespace Drupal\apigee_edge\Controller;
+
+use Drupal\apigee_edge\Entity\DeveloperAppInterface;
+use Drupal\Core\Entity\Entity\EntityViewDisplay;
+use Drupal\Core\Url;
+
+/**
+ * Provides a trait for checking developer app's callback URL..
```
mxr576: Too much dot. :)

src/Controller/DeveloperAppViewController.php
```diff
@@ -23,22 +23,61 @@
...
+    );
+  }
 
   /**
    * {@inheritdoc}
    */
   public function view(EntityInterface $developer_app, $view_mode = 'full') {
     /** @var \Drupal\apigee_edge\Entity\DeveloperAppInterface $developer_app */
     $this->checkDeveloperStatus($developer_app->getOwnerId());
+    $this->checkCallbackUrl($developer_app);
```
mxr576: This should get the view mode as a parameter and load the view mode instead of "default" always, isn't it?

src/Controller/DeveloperAppCallbackUrlCheckTrait.php
```diff
@@ -0,0 +1,58 @@
...
+    // If the Callback URL field is enabled then check its value.
+    if ($developer_app_view_display->getComponent('callbackUrl') !== NULL) {
+      // If the property value and the field value are different then the
+      // callback URL is not a valid URL.
+      if ($developer_app->getCallbackUrl() !== $developer_app->get('callbackUrl')->value) {
+        try {
+          Url::fromUri($developer_app->getCallbackUrl());
+        }
+        catch (\Exception $exception) {
+          $this->messenger->addWarning(t('The Callback URL value should be fixed. @message', [
```
mxr576: You should use `\Drupal::messenger()`  in the trait instead of $this->messenger then you can remove this property from those classes where this trait is being used and there is no other reason to inject the messenger service.

mxr576 (changes requested):

src/Controller/DeveloperAppCallbackUrlCheckTrait.php
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Checks whether the developer app's Callback URL value is valid.
+   *
+   * @param \Drupal\apigee_edge\Entity\DeveloperAppInterface $developer_app
+   *   The developer app entity.
+   *
+   * @see \Drupal\apigee_edge\Entity\DeveloperApp::set()
+   */
+  protected function checkCallbackUrl(DeveloperAppInterface $developer_app) {
+    $developer_app_view_display = EntityViewDisplay::load('developer_app.developer_app.default');
```
mxr576: Document that we only support the default view mode at this moment.

mxr576 (changes requested):
Just add one comment please :)

peterserfozo:
Test: https://travis-ci.org/peterserfozo/apigee-edge-drupal/builds/410406031

mxr576 (approved):

PR #60, opened 2018-08-01
Improves error handling.

tamasd:
* In case of an Edge error, Drupal will not redirect, just shows
  the error page. This gives the same experience as 404 or 403
  errors.
* When the developer cannot be loaded for a given user, an
  exception will be thrown instead of trying to show the page with
  empty data. The thrown exception will be caught by the error
  page handler.

### DIFF
--- a/apigee_edge.routing.yml
+++ b/apigee_edge.routing.yml
@@ -4,8 +4,7 @@ apigee_edge.error_page:
     _controller: '\Drupal\apigee_edge\Controller\ErrorPageController::render'
     _title_callback: '\Drupal\apigee_edge\Controller\ErrorPageController::getPageTitle'
   requirements:
-    _permission: 'access content'
-    _user_is_logged_in: 'TRUE'
+    _access: 'TRUE'
 
 apigee_edge.admin_config_edge:
   path: '/admin/config/apigee-edge'
@@ -86,6 +85,7 @@ apigee_edge.settings.product.caching:
     _title: 'API Product caching'
   requirements:
     _permission: 'administer apigee edge'
+
 apigee_edge.settings.product.access_control:
   path: '/admin/config/apigee-edge/product-settings/access-control'
   defaults:
--- a/apigee_edge.services.yml
+++ b/apigee_edge.services.yml
@@ -26,7 +26,11 @@ services:
 
   apigee_edge.exception_subscriber:
     class: Drupal\apigee_edge\EventSubscriber\EdgeExceptionSubscriber
-    arguments: ['@url_generator']
+    arguments:
+      - '@http_kernel'
+      - '@logger.channel.php'
+      - '@redirect.destination'
+      - '@router.no_access_checks'
     tags:
       - { name: event_subscriber }
 
--- a/src/Controller/DeveloperAppListBuilderForDeveloper.php
+++ b/src/Controller/DeveloperAppListBuilderForDeveloper.php
@@ -19,6 +19,7 @@
 
 namespace Drupal\apigee_edge\Controller;
 
+use Drupal\apigee_edge\Exception\DeveloperDoesNotExistException;
 use Drupal\apigee_edge\Entity\DeveloperAppInterface;
 use Drupal\apigee_edge\Entity\DeveloperStatusCheckTrait;
 use Drupal\apigee_edge\Entity\ListBuilder\DeveloperAppListBuilder;
@@ -95,7 +96,7 @@ class DeveloperAppListBuilderForDeveloper extends DeveloperAppListBuilder {
     // either there is connection error or the site is out of sync with
     // Apigee Edge.
     if ($developerId === NULL) {
-      return [];
+      throw new DeveloperDoesNotExistException($user->getEmail());
     }
 
     $query = $this->storage->getQuery()
--- a/src/Entity/DeveloperStatusCheckTrait.php
+++ b/src/Entity/DeveloperStatusCheckTrait.php
@@ -19,6 +19,7 @@
 
 namespace Drupal\apigee_edge\Entity;
 
+use Drupal\apigee_edge\Exception\DeveloperDoesNotExistException;
 use Drupal\Core\Url;
 use Drupal\user\Entity\User;
 
@@ -43,8 +44,11 @@ trait DeveloperStatusCheckTrait {
 
     $user = User::load($uid);
     /** @var \Drupal\apigee_edge\Entity\DeveloperInterface $developer */
-    $developer = Developer::load($user->getEmail());
-    if (!isset($developer) || $developer->getStatus() === Developer::STATUS_INACTIVE) {
+    if (!($developer = Developer::load($user->getEmail()))) {
+      throw new DeveloperDoesNotExistException($user->getEmail());
+    }
+
+    if ($developer->getStatus() === Developer::STATUS_INACTIVE) {
       // Displays different warning message for admin users.
       $message = $user->id() === \Drupal::currentUser()->id()
         ? t('Your developer account has inactive status so you will not be able to use your credentials until your account is enabled. Please contact the Developer Portal support for further assistance.')
--- a/src/EventSubscriber/EdgeExceptionSubscriber.php
+++ b/src/EventSubscriber/EdgeExceptionSubscriber.php
@@ -20,11 +20,9 @@
 namespace Drupal\apigee_edge\EventSubscriber;
 
 use Apigee\Edge\Exception\ApiException;
-use Drupal\Core\Routing\UrlGeneratorInterface;
-use Symfony\Component\EventDispatcher\EventSubscriberInterface;
-use Symfony\Component\HttpFoundation\RedirectResponse;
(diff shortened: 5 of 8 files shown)


src/Controller/DeveloperAppListBuilderForDeveloper.php
```diff
@@ -96,7 +97,7 @@ protected function getEntityIds(array $headers = [], UserInterface $user = NULL)
     // either there is connection error or the site is out of sync with
     // Apigee Edge.
     if ($developerId === NULL) {
-      return [];
+      throw new DeveloperDoesNotExistException("Edge user not found.");
```
mxr576: * Single quote
* Introduce a new developer email parameter to the constructor of this exception.
* Default message returned by the exception (if no $message is provided as the second parameter) should be something like: `Developer with %s email address not found.` Therefore you do not need to the repeat the message text every time when this exceptions needs to be thrown.
* If message override provided in the construction then $developerEmail should be available as a token. Ex.: throw new DeveloperDoesNotExistException('foobar@example.com', 'This prints out the email: %s') -> "This prints out the email: foobar@example.com"
* It should extend the ApiException, because it is actually an API issue and if it extends you do not need to check if the exception is instance of `DeveloperDoesNotExistException` in the route subscriber.

src/DeveloperDoesNotExistException.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Drupal\apigee_edge;
```
mxr576: Introduce a new `Exception` namepsace.

src/Entity/DeveloperStatusCheckTrait.php
```diff
@@ -43,8 +44,11 @@ private function checkDeveloperStatus(?int $uid) {
 
     $user = User::load($uid);
     /** @var \Drupal\apigee_edge\Entity\DeveloperInterface $developer */
-    $developer = Developer::load($user->getEmail());
-    if (!isset($developer) || $developer->getStatus() === Developer::STATUS_INACTIVE) {
+    if (!($developer = Developer::load($user->getEmail()))) {
+      throw new DeveloperDoesNotExistException('Developer not found on Edge');
```
mxr576: The same as above.

src/EventSubscriber/EdgeExceptionSubscriber.php
```diff
@@ -20,58 +20,40 @@
...
    *   The exception event.
    */
   public function onException(GetResponseForExceptionEvent $event) {
-    if ($event->getException() instanceof ApiException || $event->getException()->getPrevious() instanceof ApiException) {
-      $url = $this->urlGenerator->generateFromRoute('apigee_edge.error_page');
-      $event->setResponse(new RedirectResponse($url));
+    if (
+      $event->getException() instanceof ApiException ||
+      $event->getException()->getPrevious() instanceof ApiException ||
+      $event->getException() instanceof DeveloperDoesNotExistException ||
```
mxr576: See my comments above about this exception should extend the ApiException.

src/ParamConverter/DeveloperAppNameConverter.php
```diff
@@ -103,6 +104,7 @@ public function convert($value, $definition, $name, array $defaults) {
         // it has existed before because someone knows the URL of the view
         // app page of one of its app.
         $this->logger->critical('%class: Unable to find developer id for %user user.', ['%class' => get_called_class(), '%user' => $user->getDisplayName()]);
+        throw new DeveloperDoesNotExistException("Can't find user {$user->getDisplayName()} on Edge.");
```
mxr576: The same as above regarding exception message content.

tests/src/Functional/DeveloperDoesNotExistTest.php
```diff
@@ -0,0 +1,73 @@
...
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+namespace Drupal\Tests\apigee_edge\Functional;
+
+use Symfony\Component\HttpFoundation\Response;
+
+class DeveloperDoesNotExistTest extends ApigeeEdgeFunctionalTestBase {
```
mxr576: Don't we have a test already that validates redirection to the api-communication-error page? If there is one then this should be merged with that. There is no need to setup Drupal twice to validate the same behavior twice.

mxr576 (changes requested):

apigee_edge.routing.yml
```diff
@@ -5,7 +5,6 @@ apigee_edge.error_page:
     _title_callback: '\Drupal\apigee_edge\Controller\ErrorPageController::getPageTitle'
   requirements:
     _permission: 'access content'
-    _user_is_logged_in: 'TRUE'
```
mxr576: We initially added this to ensure that API communication error page is only visible to logged in users, it is not going to be indexed by search engines, etc. However, it could happen indeed that some content from Apigee Edge gets available to anonymous users as well and in case of a communication issue they should have access to this error page.

Question: can we ensure somehow that this page is not indexed by robots? Or is it already working like this?

apigee_edge.services.yml
```diff
@@ -26,7 +26,11 @@ services:
 
   apigee_edge.exception_subscriber:
     class: Drupal\apigee_edge\EventSubscriber\EdgeExceptionSubscriber
-    arguments: ['@url_generator']
+    arguments:
+      - '@http_kernel'
+      - '@logger.channel.php'
```
mxr576, line 31: Why not the `@logger.channel.apigee_edge`? Errors should be logged under our module's namespace in this even subscriber.

apigee_edge.routing.yml
tamasd: If this page needs a logged in user, then the subrequest will fail in the exception handler.

This page won't be linked directly, so I don't think that it will get indexed ever... but I'll take a look if the default Drupal error pages do something with search engine indexing.

peterserfozo: AFAIK if the page is not linked and/or if the path is not in sitemap.xml then it cannot be indexed. I don't know what happens if an exception is thrown when a crawler fetches one of the affected pages (if there is any for anonymous users). In my opinion, a comment is enough here. In this special case if the page is indexed then the maintainers/developers should add the path to the site's robots.txt.

src/DeveloperDoesNotExistException.php
peterserfozo: And there is an `Exception` folder for custom exception classes under `src`.

apigee_edge.routing.yml
mxr576: We just need an event subscriber that adds the robots => NOINDEX header to the api-communication-error page:
https://cgit.drupalcode.org/http_response_headers/tree/src/EventSubscriber/AddHTTPHeaders.php?id=de12feb#n71

tests/src/Functional/DeveloperDoesNotExistTest.php
```diff
@@ -0,0 +1,73 @@
...
+    $this->drupalUser->save();
+    _apigee_edge_set_sync_in_progress(FALSE);
+  }
+
+  /**
+   * Tests an unavailable developer.
+   */
+  public function testFail() {
+    $this->drupalLogin($this->drupalUser);
+    $paths = [
```
peterserfozo: `$routes = [`
`  'entity.developer_app.canonical_by_developer',`
`  'entity.developer_app.add_form_for_developer',`
`  ...`
`]`
`$route_parameters = [`
` 'user' => $this->drupalUser->id(),`
`]`
`...`
`$this->drupalGet(Url::fromRoute($route, $route_parameters));`

peterserfozo (changes requested):

apigee_edge.routing.yml
tamasd: Default Drupal error pages are not mentioned in the robots.txt or set special headers, since they never linked directly.

src/EventSubscriber/EdgeExceptionSubscriber.php
```diff
@@ -20,58 +20,34 @@
...
+  protected static function getPriority() {
+    return 1024;
   }
 
   /**
-   * Redirects user to the Edge error page.
-   *
-   * Redirects user to the Edge error page if the
-   * uncaught exception is an instance of ApiException.
+   * Displays the edge connection error page.
```
mxr576: s/edge/Edge/g

tests/modules/apigee_edge_entity_class/apigee_edge_entity_class.info.yml
```diff
@@ -0,0 +1,9 @@
+name: 'Apigee Edge: Entity class override test'
```
mxr576: Merge this with the Apigee Edge testing module and run all tests with the overridden classes and with that ensure that all functionalities are working properly even if someone overrides the entity classes.

mxr576 (changes requested):

peterserfozo (approved):

mxr576 (approved):

PR #63, opened 2018-08-10
Drupal 8.6 support and entity class overriding support

tamasd:


tests/src/Functional/ApigeeEdgeTestTrait.php
```diff
@@ -83,6 +84,7 @@ protected function invalidateKey() {
       'active_key_oauth_token' => '',
     ];
     $this->container->get('state')->set('apigee_edge.auth', $keys);
+    drupal_flush_all_caches();
```
mxr576: Clear states cache only with `$this->container->get('state')->resetCache()`

src/Entity/Controller/ApiProductController.php
```diff
@@ -31,40 +29,11 @@
...
-    }
-    $this->entityClass = $entityClass;
-  }
-
   /**
    * {@inheritdoc}
    */
-  protected function getEntityClass(): string {
-    return $this->entityClass;
+  protected function getInterface(): string {
```
mxr576: Let's call it `getEntityInterface()` instead.

src/Entity/Controller/DeveloperController.php
```diff
@@ -21,57 +21,35 @@
...
-   *   The API client.
-   * @param string $entityClass
-   *   The FQCN of the entity class that is used in Drupal.
-   * @param \Apigee\Edge\Serializer\EntitySerializerInterface|null $entitySerializer
-   *   The entity serializer.
-   * @param \Apigee\Edge\Api\Management\Controller\OrganizationControllerInterface|null $organizationController
-   *   The organization controller.
-   *
-   * @throws \ReflectionException
+   * {@inheritdoc}
```
mxr576: I would not use inheritdoc on constructor's description. Does PHPCS complains about the parameters if you do not provide a description to them?

src/Entity/Controller/DrupalEntityControllerAwareTrait.php
```diff
@@ -31,16 +34,39 @@
...
 
+  /**
+   * @var string
+   */
+  protected $entityClass;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(string $organization, ClientInterface $client, string $entityClass, ?EntitySerializerInterface $entitySerializer = NULL, ?OrganizationControllerInterface $organizationController = NULL) {
```
mxr576: Inheriting constructor from a trait? 😨`parent` does not point to anywhere in this context.

src/Entity/Controller/DrupalEntityControllerAwareTrait.php
```diff
@@ -31,16 +34,39 @@
...
+    parent::__construct($organization, $client, $entitySerializer, $organizationController);
+    $rc = new \ReflectionClass($entityClass);
```
mxr576: I'd rather create an utility function from this and call that in the constructor of the entity classes.

src/Entity/Controller/DrupalEntityControllerAwareTrait.php
```diff
@@ -31,16 +34,39 @@
  */
 trait DrupalEntityControllerAwareTrait {
 
+  /**
+   * @var string
+   */
+  protected $entityClass;
```
mxr576: Describe that this is an FQCN of an entity class.

src/Entity/Query/QueryFactory.php
```diff
@@ -69,33 +69,107 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager) {
...
+   * Otherwise the default \Drupal\apigee_edge\Entity\Query\Query class
+   * if being used.
+   *
+   * @param string $class
+   *   Entity class.
+   *
+   * @return null|string
+   *   Query class or null if not found.
+   */
+  protected function findQueryClass(string $class): ?string {
```
mxr576: If we would like to refactor this, can't we just introduce a new annotation in our entity classes that can defined the FQCN of the entity query class (with fallback to the default entity class)?

tests/src/Functional/ApigeeEdgeTestTrait.php
```diff
@@ -72,6 +72,7 @@ protected function restoreKey() {
...
```
mxr576: Clear states cache only with `$this->container->get('state')->resetCache()`

tests/src/Functional/DeveloperAppFieldTest.php
```diff
@@ -545,6 +546,7 @@ protected function assertFieldVisibleOnEntityForm(string $field_label, bool $vis
    */
   protected function assertFieldVisibleOnEntityDisplay(string $app_name, string $field_label, string $field_value, bool $visible = TRUE) {
     $this->drupalGet("/user/{$this->account->id()}/apps/{$app_name}");
+    $this->assertEquals(Response::HTTP_OK, $this->getSession()->getStatusCode());
```
mxr576: Why is this necessary?

mxr576 (changes requested):

tests/src/Functional/ApigeeEdgeTestTrait.php
tamasd: For some reason that doesn't do the trick in Drupal 8.6. I spent half a day trying to figure out why.

tests/src/Functional/DeveloperAppFieldTest.php
tamasd: When the test fails here for some reason, I get a more meaningful message (e.g. 404 instead of 200), and not an "element xpath=blabla not found on page".

mxr576: Ok.

tests/src/Functional/ApigeeEdgeTestTrait.php
mxr576: It would be really great of we could figure this out, because it is critical for us.

src/Entity/Controller/ApiProductController.php
mxr576: @tamasd  ^

mxr576:
TODOs:
* Version information should be modified in README.md as well
* 8.5.x-dev references in [.travis.yml](.travis.yml) should be replaced with 8.6.x-dev.

@tamasd ^

src/Entity/Query/QueryFactory.php
mxr576: @tamasd ^

src/Entity/Controller/DrupalEntityControllerAwareTrait.php
```diff
@@ -31,16 +34,47 @@
...
+   * The FQCN of the Drupal entity class.
+   *
+   * @var string
+   */
+  protected $entityClass;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(string $organization, ClientInterface $client, string $entity_class, ?EntitySerializerInterface $entity_serializer = NULL, ?OrganizationControllerInterface $organization_controller = NULL) {
```
mxr576: I think we should not inherit the constructor of a class from a trait.

mxr576 (changes requested):

mxr576:
Latest test results:
https://travis-ci.org/tamasd/apigee-edge-drupal/builds/426595129

mxr576 (approved):

peterserfozo (approved):

PR #66, opened 2018-09-24
Authentication fixes

peterserfozo:


tamasd (approved):

src/Form/AuthenticationForm.php
```diff
@@ -526,11 +530,21 @@ protected function createDebugText(\Exception $exception, KeyInterface $key, ?Ke
...
+      'username' => '',
     ];
 
+    try {
+      $credentials['endpoint'] = $key_type->getEndpoint($key);
+      $credentials['organization'] = $key_type->getOrganization($key);
+      $credentials['username'] = $key_type->getUsername($key);
+    }
+    catch (KeyValueMalformedException $exception) {
+      // Could not read the credentials because the key value storage is
```
mxr576, line 540: Should not you return this error message here instead of initializing endpoint/organization/username to empty strings above? Does it worth the run the remaining code in this method when this exception occurs?

mxr576 (commented):
See my comment inline.

src/Form/AuthenticationForm.php
peterserfozo: This exception is thrown if someone deleted the content of the key file by hand or removed an environment variable after saving (and validating) the key. The error message ("Failed to connect to Apigee Edge. Check the debug information below for more details. Error message: Apigee Edge API authentication key is malformed or not readable. Missing field: username.") is displayed for the user in line 391. This function is called after displaying the error message and only generates the text area that contains the debug information. Because the key values cannot be read we should catch this exception again when we try to access the key values. But it's a special case and I think the remaining code is necessary because that will print for example the key type/provider. Maybe the empty strings should be replaced by `cannot_be_read/malformed/damaged` placeholders?

mxr576 (approved):

peterserfozo:
https://travis-ci.org/peterserfozo/apigee-edge-drupal/builds/433357113

PR #70, opened 2018-10-03
Fixes compatibility with field widgets that rely on extractFormValues().

tamasd:


src/Entity/FieldableEdgeEntityBaseTrait.php
```diff
@@ -377,13 +377,25 @@ public function set($field_name, $value, $notify = TRUE) {
...
+          $value = $new_value;
+        }
+        else {
+          $new_value = NestedArray::getValue($value, ['value'], $exists);
+          if ($exists) {
+            $value = $new_value;
+          }
+          else {
+            $called_class = get_called_class();
+            $exported_value = var_export($value, TRUE);
```
mxr576: Could it be a security risk that the exception (which probably gets logged) contains a value of the field?

src/Entity/FieldableEdgeEntityInterface.php
```diff
@@ -26,7 +26,7 @@
 /**
  * Interface for fieldable Edge entities.
  */
-interface FieldableEdgeEntityInterface extends FieldableEntityInterface, EdgeEntityInterface {
+interface FieldableEdgeEntityInterface extends \IteratorAggregate, FieldableEntityInterface, EdgeEntityInterface {
```
mxr576, line 29: Where exactly leverage the newly extended interface in your PR? I see FieldItemInterface type hint in the array_map() below.

tamasd: If a fieldable entity does not implement `Traversable`, then the form display's `extractFormValues()` will always return an empty array (and a warning), since it tries to traverse the entity's fields with a foreach.

src/Entity/FieldableEdgeEntityBaseTrait.php
tamasd: It exposes the field's content, that's true. I am not sure if it can be a security problem, since you need admin permissions to access the logs.

If this part of the code runs then it is always a bug (the content of the field won't be saved), and it this makes troubleshooting so much easier.

mxr576: What if logs are redirected to an unsecured Redis instance? :) What I am trying to say is that are you sure that it adds that much value to debugging to risk this? If it would be encrypted with the Drupal instance's hash salt at least I would be more relieved, but yes we haven't added such thing to our module yet.
Anyway I can approve this PR like this.

src/Entity/FieldableEdgeEntityInterface.php
mxr576: Could you please add this information to the interface's documentation? That might be useful to know later.

mxr576:
Related test in Travis: https://travis-ci.org/tamasd/apigee-edge-drupal/builds/436617351

mxr576 (approved):

PR #79, opened 2018-10-22
Move mock client testing sub-module from Monetization to Edge

void--:


mxr576:
I think we can only move this to the apigee_edge module when _all_ MGMT API tests could be executed with this mock client. So the primary goal would be running all MGMT/MINT API tests with this mock client instead of calling a real Apigee Edge backend - without reinventing Apigee Edge in PHP. At this moment I do not see all required templates in this PR for running all MGMT API tests that apigee_edge module currently contains.

FYI, I also started to work on a mock client for the SDK that besides uses templates to format answers to API calls it also uses a persistent (like-ish) storage that will allow us to test complex API calls with random data - from the creation of an Apigee Edge entity (ex.: developer app) until its removal. This mock client could be leveraged both in the sdk and the Drupal 8 module - not just in the Drupal 8 module the content of this PR.

void--:
I agree that in an ideal world this wouldn't get moved over until all tests were converted, but I think the idea of moving it over now is so that we can avoid writing future tests twice -- new tests can leverage the apigee_mock_client module until we get around to converting the existing ones.

cnovak:
Yes, we plan to move the tests and update CONTRIBUTING.md to tell people to use this system going forward and idea on how to use.  @void-- can you update docs to reflect this in the PR?

tests/modules/apigee_mock_client/apigee_mock_client.info.yml
```diff
@@ -0,0 +1,11 @@
+name: Apigee Edge Test Module
```
cnovak: Should there be a README for this module, or are you going to add info to Apigee Edge module on how this is used /overview?  We can start w/something simple

tests/modules/apigee_mock_client/src/TwigExtension/ModulePathLoader.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+namespace Drupal\apigee_mock_client\TwigExtension;
+
+use Drupal\Core\Extension\ModuleHandlerInterface;
+
+/**
+ * Loads templates from the response-templates folder of a module.
+ */
+class ModulePathLoader extends \Twig_Loader_Filesystem {
```
cnovak: This class is loading response templates in the `/tests/response-templates` dir, not generic paths in a module. Suggest the name to be changed to something like `ResponseTemplateLoader` or something meaningful for what it is doing.  Examples in Drupal are `FileSystemLoader` and `ThemeRegistryLoader`, which convey what they are doing.

mxr576:
Okay, let's check this PR from a different point of view.

At this moment apigee_mock_client only provides *limited* support to mock MINT API calls. When it starts supporting MGMT calls as well then the MINT API related mock templates should not be available in the apigee_edge module if the MINT integration remains [a separate module](https://github.com/apigee/apigee-m10n-drupal) and not a sub-module of apigee_edge (like the [API product RBAC](https://github.com/apigee/apigee-edge-drupal/tree/8.x-1.x/modules/apigee_edge_apiproduct_rbac)).

So how I would organize these modules:
* apigee_edges/tests/apigee_mock_client - Base module that provides the mocking framework, also contains MGMT API related templates and tweaks.
* apigee_m10n/tests/apigee_mint_mock_client - Depends on the apigee_mock_client and contains the MINT API related templates and tweaks.

Even better: apigee_mock_client would be a separate contrib module/library in which both modules could depend on and it would only contain the pure mocking framework, the dependent modules would maintain its own mock response templates. That would make much easier the long term support and maintenance of these components.

Main reason behind these ideas: We do not want to release a new release from apigee_edge module just to adjust/add mock API client templates for the apigee_m10n module.

Also, I still think the API mocking should be implemented in the PHP API client level and not in a Drupal 8 module only which could be leveraged only in Drupal. I am still working on a POC in my free-time as a demonstration of this idea as we do not have an allocated sprint item for this.

tests/modules/apigee_mock_client/src/TwigExtension/ModulePathLoader.php
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+   * Constructs a new FilesystemLoader object .
+   *
+   * @param string $module_name
+   *   The name of the module to load templates for.
+   * @param \Drupal\Core\Extension\ModuleHandlerInterface $module_handler
+   *   The module handler service.
+   */
+  public function __construct($module_name, ModuleHandlerInterface $module_handler) {
```
cnovak: `module_name` is not being used, but being passed in by services.yml.  Is this constructor supposed to only get test templates paths for module passed in, or for all modules?

tests/modules/apigee_mock_client/src/TwigExtension/ModulePathLoader.php
```diff
@@ -0,0 +1,50 @@
...
+    $modules = $module_handler->getModuleList();
+
+    $paths = array_map(function($module) {
+      return $module->getPath() . '/tests/response-templates';
+    }, $modules);
+
+    $paths = array_filter($paths, function($path) {
```
cnovak: Can you add comment above line on why this array filter is needed?

void--:
@cnovak for sure, I'll update the docs

@mxr576 the template loader looks for a `response_templates` directory in all module's `tests` dir, so we shouldn't have an issue with releases. As for your other issues, I'm definitely open, but in the mean time should we merge this PR (once I fix the formatting issues which I'm working on right now) so that we have something we can use right now?

tests/modules/apigee_mock_client/src/MockHandlerStack.php
```diff
@@ -0,0 +1,164 @@
...
+    if (empty($this->responses)) {
+      // Get the module path for this module.
+      $module_path = \Drupal::moduleHandler()->getModule('apigee_mock_client')->getPath();
+      $this->responses = Yaml::parseFile($module_path . '/response_catalog.yml')['responses'];
+    }
+    // Loop through responses and add each one.
+    foreach ((array) $response_ids as $index => $item) {
+      // The catalog id should either be the item itself or the keys if an associative array has been passed.
+      $id = !is_array($item) ? $item : $index;
+      // Body text can have elements replaced in it for certain values. See: http://php.net/strtr
```
cnovak: Is this an out of date comment saying `See: http://php.net/strtr`?

tests/modules/apigee_mock_client/src/MockHttpClientFactory.php
```diff
@@ -0,0 +1,110 @@
...
+   */
+  public function fromOptions(array $config = []) {
+    $default_config = [
+      // Security consideration: we must not use the certificate authority
+      // file shipped with Guzzle because it can easily get outdated if a
+      // certificate authority is hacked. Instead, we rely on the certificate
+      // authority file provided by the operating system which is more likely
+      // going to be updated in a timely fashion. This overrides the default
+      // path to the pem file bundled with Guzzle.
+      'verify' => TRUE,
```
cnovak: The default for Guzzle 6 seems to be `'verify' => TRUE`: http://docs.guzzlephp.org/en/stable/request-options.html#verify Does Guzzle 6 no longer ship w/a CA file?

cnovak (changes requested):
Put in comments, also codesniffer is failing in CI tests

mxr576:
> @mxr576  but in the mean time should we merge this PR (once I fix the formatting issues which I'm working on right now) so that we have something we can use right now?

Why don't you keep this module as the sub-module of the apigee_m10n module? Why it is necessary to be part of the apigee_edge module? It does not provide any value to the apigee_edge module and it just makes its maintainable more complicated. If the mocking module gets matured enough and it can serve the purpose of the apigee_edge module it can be moved to that.

tests/modules/apigee_mock_client/src/MockHttpClientFactory.php
mxr576: This is also an unnecessary copy-paste of ClientFactory::fromOptions().
https://github.com/drupal/core/blob/17557c97a816bc78e70989a86064c562bc27027f/lib/Drupal/Core/Http/ClientFactory.php#L41.

Jaesin:
Should we rename `::queueFromResponseFile` to something that makes more since since changing to using twig templates?

tests/modules/apigee_mock_client/apigee_mock_client.info.yml
mxr576: Module's (display) name should be more unique.

tests/modules/apigee_mock_client/src/ApigeeMockClientServiceProvider.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+namespace Drupal\apigee_mock_client;
+
+use Drupal\Core\DependencyInjection\ServiceProviderBase;
+use Drupal\Core\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * Class ApigeeM10nTestServiceProvide
```
mxr576: Outdated name.

tests/modules/apigee_mock_client/src/ApigeeMockClientServiceProvider.php
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * This class is automatically picked up by the container builder.
+ * See: https://www.drupal.org/docs/8/api/services-and-dependency-injection/altering-existing-services-providing-dynamic-services
+ *
+ * @package Drupal\apigee_mock_client
```
mxr576: We do not use `@package` annotations in D8 code.

tests/modules/apigee_mock_client/src/ApigeeMockClientServiceProvider.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+class ApigeeMockClientServiceProvider extends ServiceProviderBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function alter(ContainerBuilder $container) {
+    // Override the ClientFactory with our mock client factory.
+    $container->getDefinition('apigee_edge.sdk_connector')
```
mxr576, line 44: Why don't you just created a service decorator and DI-d your client factory?
https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/tests/modules/apigee_edge_test/apigee_edge_test.services.yml#L20
https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/tests/modules/apigee_edge_test/src/SDKConnector.php

tests/modules/apigee_mock_client/src/MockHandlerStack.php
```diff
@@ -0,0 +1,164 @@
...
+namespace Drupal\apigee_mock_client;
+
+use Apigee\Edge\Exception\ClientErrorException;
+use Drupal\Core\Queue\QueueFactory;
+use GuzzleHttp\Handler\MockHandler;
+use GuzzleHttp\Psr7\Response;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Symfony\Component\Yaml\Yaml;
+
```
mxr576, line 30: Missing class description.

tests/modules/apigee_mock_client/src/MockHandlerStack.php
```diff
@@ -0,0 +1,164 @@
...
+class MockHandlerStack extends MockHandler {
+
+  /**
+   * Responses that have been loaded from the response file.
+   *
+   * @var array
+   */
+  protected $responses;
```
mxr576: Because it is an array according to its annotation it should be initialized. Has the type of the array elements already known? If yes the annotation should provide info about that, ex.: `string[]`

tests/modules/apigee_mock_client/src/MockHandlerStack.php
```diff
@@ -0,0 +1,164 @@
...
+      $headers = isset($this->responses[$id]['headers']) ? $this->responses[$id]['headers'] : [
+        'content-type' => 'application/json;charset=utf-8'
+      ];
+      // Set the default status code.
+      $status_code = !empty($this->responses[$id]['status_code']) ? $this->responses[$id]['status_code'] : 200;
+      $status_code = !empty($context['status_code']) ? $context['status_code'] : $status_code;
+
+      // Render the response content.
+      $template = str_replace('_', '-', $id) . '.json.twig';
+      $content = $this->twig->getLoader()->exists($template) ? $this->twig->render($template, $context) : '';
```
mxr576: How other modules can alter the response ids, status code and/or template file before the response object is being constructed?

tests/modules/apigee_mock_client/src/MockHandlerStack.php
```diff
@@ -0,0 +1,164 @@
...
+      }
+
+      return parent::__invoke($request, $options);
+    } catch (\Exception $ex) {
+      // Fake a client error so the error gets reported during testing. The
+      // `apigee_edge_user_presave` is catching errors in a way that this error
+      // never gets reported to PHPUnit. This won't work for all use cases.
+      //
+      // @todo: Find another way to ensure an exception gets registered with PHPUnit.
+      throw new ClientErrorException(new Response(500, [
```
mxr576, line 159: Try to call `\PHPUnit_Framework_Assert::fail()` here.

tests/modules/apigee_mock_client/src/MockHandlerStack.php
```diff
@@ -0,0 +1,164 @@
...
+
+        // Append items to a database queue so they can be accessed in process isolated environments (functional tests).
+        $this->database_queue->createItem([
+          'status' => $value->getStatusCode(),
+          'headers' => $value->getHeaders(),
+          'body' => (string) $value->getBody()
+        ]);
+
+      } else {
+        throw new \InvalidArgumentException('Expected a response. '
```
mxr576: Please fix code indentation style.

tests/modules/apigee_mock_client/src/MockHandlerStack.php
```diff
@@ -0,0 +1,164 @@
...
+        'Content-Type' => 'application/json',
+      ], '{"code": "4242", "message": "'.$ex->getMessage().'"}'), $request, $ex->getMessage(), $ex->getCode(), $ex);
+    }
+  }
+
+  public function append() {
+    foreach (func_get_args() as $value) {
+      if ($value instanceof ResponseInterface) {
+
+        // Append items to a database queue so they can be accessed in process isolated environments (functional tests).
```
mxr576: > Lines containing comments (including docblocks) must wrap as close to 80 characters as possible without going over, with a few exceptions (noted in the Tag Reference below).
https://www.drupal.org/docs/develop/standards/api-documentation-and-comment-standards

tests/modules/apigee_mock_client/src/MockHttpClientFactory.php
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * Whether or not integration is currently enabled.
+   *
+   * @var bool
+   */
+  protected $integration_enabled;
+
+  /**
+   * Constructs a new ClientFactory instance.
```
mxr576, line 57: Outdated class name.

tests/modules/apigee_mock_client/src/MockHttpClientFactory.php
```diff
@@ -0,0 +1,110 @@
...
+use Drupal\Component\Utility\NestedArray;
+use Drupal\Core\Http\ClientFactory;
+use Drupal\Core\Site\Settings;
+use Drupal\Core\State\StateInterface;
+use GuzzleHttp\Client;
+use GuzzleHttp\Handler\MockHandler;
+use GuzzleHttp\HandlerStack;
+
+
+class MockHttpClientFactory extends ClientFactory {
```
mxr576, line 33: Missing class description, run PHPCS to find similar code style errors.

tests/modules/apigee_mock_client/src/MockHttpClientFactory.php
```diff
@@ -0,0 +1,110 @@
...
+   *   The handler stack.
+   * @param \GuzzleHttp\Handler\MockHandler $mock_stack
+   * @param \Drupal\Core\State\StateInterface $state
+   */
+  public function __construct(HandlerStack $stack, MockHandler $mock_stack, StateInterface $state) {
+    $this->stack = $stack;
+    $this->mock_stack = $mock_stack;
+
+    // Check for the integration enabled environment variable.
+    if ($enabled = getenv('APIGEE_INTEGRATION_ENABLE')) {
```
mxr576: Use the name of the module in environment variables as prefix . `APIGEE_INTEGRATION_ENABLE ` is not unique enough.

tests/modules/apigee_mock_client/src/TwigExtension/ModulePathLoader.php
```diff
@@ -0,0 +1,50 @@
...
+
+use Drupal\Core\Extension\ModuleHandlerInterface;
+
+/**
+ * Loads templates from the response-templates folder of a module.
+ */
+class ModulePathLoader extends \Twig_Loader_Filesystem {
+
+  /**
+   * Constructs a new FilesystemLoader object .
```
mxr576: Outdated.

tests/modules/apigee_mock_client/src/MockHandlerStack.php
```diff
@@ -0,0 +1,164 @@
...
+   *     - `['foo' => [':bar' => 'baz']]` // w/ replacements for response body.
+   *
+   * @return $this
+   *
+   * @throws \Twig_Error_Loader
+   * @throws \Twig_Error_Runtime
+   * @throws \Twig_Error_Syntax
+   */
+  public function queueFromResponseFile($response_ids) {
+    $org_name = \Drupal::service('apigee_edge.sdk_connector')->getOrganization();
```
mxr576: Why did you DI-d the sdk connector?

tests/modules/apigee_mock_client/src/MockHandlerStack.php
```diff
@@ -0,0 +1,164 @@
...
+
+    if (empty($this->responses)) {
+      // Get the module path for this module.
+      $module_path = \Drupal::moduleHandler()->getModule('apigee_mock_client')->getPath();
```
mxr576: Module handler can be DI-d too.

mxr576 (changes requested):

tests/modules/apigee_mock_client/src/ApigeeMockClientServiceProvider.php
void--: That would be another way to do this, but is it necessarily better?

tests/modules/apigee_mock_client/src/MockHandlerStack.php
void--: We've had some issues with the sdk_connector getting constructed before credentials are available... This might have had something to do with that, but I think it's not the end of the world to leave it for now either way.

tests/modules/apigee_mock_client/src/MockHandlerStack.php
void--: You're right, but I'm going to leave this for now in the name of expediency. I'll add a @todo

tests/modules/apigee_mock_client/src/MockHandlerStack.php
cnovak: Great suggestion @mxr576.  Let’s put this in as a task to validate the approach.

tests/modules/apigee_mock_client/apigee_mock_client.info.yml
Jaesin: A readme has been added in: commit f134dd7.

tests/modules/apigee_mock_client/src/TwigExtension/ModulePathLoader.php
Jaesin: This has been renamed in commit f134dd7.

tests/modules/apigee_mock_client/src/MockHttpClientFactory.php
```diff
@@ -72,14 +72,14 @@ public function __construct(HandlerStack $stack, MockHandler $mock_stack, StateI
     $this->mock_stack = $mock_stack;
 
     // Check for the integration enabled environment variable.
-    if ($enabled = getenv('APIGEE_INTEGRATION_ENABLE')) {
+    if ($enabled = getenv('APIGEE_MOCK_CLIENT_INTEGRATION_ENABLE')) {
```
Jaesin: Can we change this to `APIGEE_EDGE_INTEGRATION_ENABLED` to be more closely aligned with the other environment variables that are documented in `CONTRIBUTING.md`.

tests/modules/apigee_mock_client/src/TwigExtension/ModulePathLoader.php
Jaesin: The `module_name` parameter has been removed in commit f134dd7.

tests/modules/apigee_mock_client/src/TwigExtension/ModulePathLoader.php
Jaesin: Comment added in commit f134dd7.

tests/modules/apigee_mock_client/src/MockHttpClientFactory.php
Jaesin: We can just leverage the base class with this: 

```
  /**
   * {@inheritdoc}
   */
  public function fromOptions(array $config = []) {
    $config = [
      'handler' => $this->integration_enabled ? $this->stack : $this->mock_stack,
    ];

    return parent::fromOptions($config);
  }
```

tests/modules/apigee_mock_client/src/MockHandlerStack.php
Jaesin: Resolved in commit f134dd7.

tests/modules/apigee_mock_client/src/MockHandlerStack.php
Jaesin: Description added in 3028f75

tests/modules/apigee_mock_client/src/MockHandlerStack.php
```diff
@@ -0,0 +1,192 @@
...
+    $this->database_queue = $queue_factory->get('apigee_edge_mock_responses', TRUE);
+
+    parent::__construct();
+  }
+
+  /**
+   * Queue a response that is in the catalog.
+   *
+   * Dynamic values can be passed and
+   * will be replaced in the response. For example `['foo' => [':bar' =>
```
Jaesin: This format is outdated: `[':bar'`.

tests/modules/apigee_mock_client/src/MockHandlerStack.php
Jaesin: There shouldn't be a need to alter response ids or the template file. The status code can be passed in i.e. `queueMockResponse(['foo' => ['status_code' => 201]])`.

mxr576:
@cnovak Please share your 2 cents about this:

> Why don't you keep this module as the sub-module of the apigee_m10n module? Why it is necessary to be part of the apigee_edge module? It does not provide any value to the apigee_edge module and it just makes its maintainable more complicated. If the mocking module gets matured enough and it can serve the purpose of the apigee_edge module it can be moved to that.

Previous comment: https://github.com/apigee/apigee-edge-drupal/pull/79#issuecomment-433983890

tests/modules/apigee_mock_client/src/MockHandlerStack.php
Jaesin: I guess you mean `throw new \PHPUnit_Framework_AssertionFailedError();`. It looks like `\PHPUnit_Framework_Assert::fail()` was removed.

tests/modules/apigee_mock_client/src/ApigeeMockClientServiceProvider.php
mxr576: Yes, because what you have DI-d you can easily replace when you write tests. Where you injected a service with `\Drupal::service()` you can not mock it that easily in tests.

tests/modules/apigee_mock_client/src/MockHandlerStack.php
mxr576: No, what I tried to achieve here that instead of throwing an exception, call PHPUnit's `fail()` method to fail a test and PHPUnit will do what it is necessary to do that.

tests/modules/apigee_mock_client/src/MockHandlerStack.php
mxr576: Use `sprintf()`, `strtr()` or `"{var}"`, etc for this.

tests/modules/apigee_mock_client/src/TwigExtension/ResponseTemplateLoader.php
```diff
@@ -0,0 +1,52 @@
...
+
+use Drupal\Core\Extension\ModuleHandlerInterface;
+
+/**
+ * Loads templates from the response-templates folder of a module.
+ */
+class ResponseTemplateLoader extends \Twig_Loader_Filesystem {
+
+  /**
+   * Constructs a new FilesystemLoader object .
```
Jaesin: This still needs to be updated.

tests/modules/apigee_mock_client/src/MockHttpClientFactory.php
cnovak: Maybe `APIGEE_MOCK_CLIENT_IS_MOCK_SERVER_ENABLED`?

tests/modules/apigee_mock_client/src/MockHttpClientFactory.php
void--: Done

tests/modules/apigee_mock_client/src/ApigeeMockClientServiceProvider.php
void--: Gotcha, I'll add a todo, but I don't think we need to get this in for this pull request.

tests/modules/apigee_mock_client/src/MockHttpClientFactory.php
void--: I went with `APIGEE_MOCK_CLIENT_MOCK_SERVER_ENABLED`, hope that's cool!

tests/modules/apigee_mock_client/src/TwigExtension/ModulePathLoader.php
void--: Done

tests/modules/apigee_mock_client/src/MockHandlerStack.php
void--: removed

tests/modules/apigee_mock_client/src/TwigExtension/ResponseTemplateLoader.php
void--: which part?

tests/modules/apigee_mock_client/src/MockHttpClientFactory.php
void--: we went with APIGEE_MOCK_CLIENT_MOCK_SERVER_ENABLED, but I'm not married to it. @cnovak @mxr576 would you be open to naming it `APIGEE_EDGE_INTEGRATION_ENABLED` or `APIGEE_EDGE_MOCK_SERVER_ENABLED`

tests/modules/apigee_mock_client/src/MockHttpClientFactory.php
void--: done

tests/modules/apigee_mock_client/src/MockHandlerStack.php
void--: Done.

tests/modules/apigee_mock_client/src/MockHandlerStack.php
void--: @mxr576 can you put a task in the backlog to address this? I think we need to get this pull request in.

cnovak:
We will leave this unmerged for now, because we would like to see this validated by making sure it run alongside the integration tests w/out having to change all tests in one full sweep.

cnovak:
Closing due to our newer approach makes this obsolete: #308

PR #81, opened 2018-11-05
Simplified authentication UI and refactored Apigee Edge credential storages

Jaesin:
TODO: 

- Adjust the functional tests to use the `OauthTokenCacheStorage`.
- Deprecate the `OauthTokenStorage` token storage.
- Adjust tests to use the `apigee_auth` key type and the `apigee_auth_input` key input plugins.
- Deprecate the `apigee_edge_basic_auth`, `apigee_edge_oauth` and `apigee_edge_oauth_token` key  types.
- Deprecate the `apigee_edge_basic_auth_input` and the `apigee_edge_oauth_input` key input plugins.


src/OauthTokenCacheStorage.php
```diff
@@ -0,0 +1,158 @@
...
+
+use Apigee\Edge\HttpClient\Plugin\Authentication\OauthTokenStorageInterface;
+use Drupal\Core\Cache\CacheBackendInterface;
+use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Storing and returning OAuth access token data.
+ */
+class OauthTokenCacheStorage implements OauthTokenStorageInterface, ContainerInjectionInterface {
```
mxr576: @cnovak So do you disagree with me about storing the Oauth authentication data in cache is insecure?

src/OauthTokenCacheStorage.php
```diff
@@ -0,0 +1,158 @@
...
+  public function __construct(CacheBackendInterface $cache) {
+    $this->cache = $cache;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
+    return new static(
+      $container->get('cache.apigee_edge')
```
mxr576: Cache id should be more unique than this.

Jaesin:
Hi @mxr576 Thanks for looking over this. `cache.apigee_edge` is the ID of the cache bin. It's not the CID. It's meant to be general purpose cache for the Apigee edge module. I don't think we want to dedicate a whole cache bin for just 2 entries. We will suggest that production should be using memcache or redis cache storage for this bin.

mxr576:
Yeah, storing OAuth access token in an unsecured Redis instance (as a cache backend) is one of my worst nightmares, this is why I am so against storing any kind of credentials (including OAuth access/refresh tokens) in the cache layer instead of in a Key module storage. Compared with the Key module that provides actually secure storages for storing Apigee Edge related credentials - although insecure storages also available - I do not consider neither the default (database backend) nor any of the available cache backends (via 3rd party modules) as secure credential storages.

src/Form/AuthenticationForm.php
```diff
@@ -356,37 +224,86 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      $processed_values = $input_plugin->processSubmittedKeyValue($plugin_form_state);
+
+      if ($form_state->isSubmitted()) {
+        // Save the processed value for the submit handler.
+        $form_state->set('processed_submitted', $processed_values['processed_submitted']);
+      }
+
+      $input_plugin->validateConfigurationForm($form, $plugin_form_state);
+      $form_state->setValue('key_input_settings', $plugin_form_state->getValues());
+      $this->moveFormStateErrors($plugin_form_state, $form_state);
```
mxr576, line 257: Are you sure that this is the only and the right way to inject the from provided by the key module and work with that? It seems hacky.

mxr576 (changes requested):

src/Form/AuthenticationForm.php
Jaesin: This is how the Key module adds the plugin forms to the key config form.
@see: `\Drupal\key\Form\KeyFormBase::form()`

src/EventSubscriber/EdgeExceptionSubscriber.php
```diff
@@ -46,7 +47,13 @@ protected static function getPriority() {
    *   The exception event.
    */
   public function onException(GetResponseForExceptionEvent $event) {
-    if ($event->getException() instanceof ApiException || $event->getException()->getPrevious() instanceof ApiException) {
+    $exception = $event->getException();
+
+    // Creating a sub request on `AuthenticationKeyValueMalformedException` can
+    // cause an infinite loop.
+    if (!($exception instanceof AuthenticationKeyValueMalformedException)
```
mxr576: 1. Isn't it should be in one line?
2. How we did not see this problem in the previous implementation? What has changed since that can possible cause this issue?

src/Form/AuthenticationForm.php
```diff
@@ -43,6 +47,11 @@
  */
 class AuthenticationForm extends ConfigFormBase {
 
+  /**
+   * The config named used by this form.
+   */
+  const CONFIG_NAME = 'apigee_edge.auth';
```
mxr576, line 54: Is it necessary to be public? Protected or rather private would not be enough?

src/Form/AuthenticationForm.php
```diff
@@ -106,21 +122,63 @@ public function getFormId() {
...
+    if (!($active_key_id = $config->get('active_key')) || !($active_key = Key::load($active_key_id))) {
+      $active_key = $this->generateNewAuthKey();
+    }
+
+    // Save the key for later use.
+    $this->activeKey = $active_key;
+
+    $form['connection_settings'] = [
+      '#type' => 'details',
+      '#title' => $this->t('Apigee edge connection settings'),
```
mxr576: s/edge/Edge/g

src/Form/AuthenticationForm.php
```diff
@@ -106,21 +122,63 @@ public function getFormId() {
...
+      '#description' => $this->t('Apigee connection settings.'),
```
mxr576: Missing "Edge".

src/Form/AuthenticationForm.php
```diff
@@ -106,21 +122,63 @@ public function getFormId() {
...
     $form = parent::buildForm($form, $form_state);
+
+    $config = $this->config(static::CONFIG_NAME);
+
     $form['#prefix'] = '<div id="apigee-edge-auth-form">';
     $form['#suffix'] = '</div>';
     $form['#attached']['library'][] = 'apigee_edge/apigee_edge.admin';
 
+    // Gets the active key.
+    if (!($active_key_id = $config->get('active_key')) || !($active_key = Key::load($active_key_id))) {
```
mxr576: I'd rather move the logic between line 141 and 146 to a new method and call that here. Reasons:
1) `$this->activeKey` should not be uninitialized ever. So you should either call `$this->generateNewAuthKey();` in the constructor of the class (which I do not recommend) or add `\Drupal\key\KeyInterface|null` to the annotation of the property.
2) If some would like to retrieve the active key based on the logic above somewhere else it would be more easier to him/her to call a method instead of copy-paste the code above.

src/Form/AuthenticationForm.php
mxr576: Okay, so please document in your code all those code places that you have "copy-pasted" from the Key module, because if the API of the Key module changes it will be much easier for us to fix our code.

src/Form/AuthenticationForm.php
```diff
@@ -597,4 +500,155 @@ public function ajaxCallback(array $form, FormStateInterface $form_state): array
...
+   * @return \Drupal\key\KeyInterface
+   *   A new auth key.
+   *
+   * @throws \Drupal\Core\Entity\EntityStorageException
+   * @throws \Drupal\key\Exception\KeyValueNotSetException
+   */
+  protected function generateNewAuthKey() {
+    // Create a new key name.
+    $new_key_id = 'apigee_edge_connection_default';
+    $file_path = "private://{$new_key_id}_apigee_edge.json";
```
mxr576: `apigee_edge_connection_default_apigee_edge.json` isn't it a weird name for this file? Is the apigee_edge suffix really necessary?

src/Form/AuthenticationForm.php
```diff
@@ -597,4 +500,155 @@ public function ajaxCallback(array $form, FormStateInterface $form_state): array
...
+  /**
+   * Creates a new auth key stored in a file.
+   *
+   * @return \Drupal\key\KeyInterface
+   *   A new auth key.
+   *
+   * @throws \Drupal\Core\Entity\EntityStorageException
+   * @throws \Drupal\key\Exception\KeyValueNotSetException
+   */
+  protected function generateNewAuthKey() {
```
mxr576: @cnovak I still disagree with this approach where we do not give the chance for users to save their Apigee Edge credentials an actually secure storage instead of an *unencrypted* file in the private file system. I thought we have made a step forward about this topic in our past conversations. We would just need to display the available key storages on the form, even if we hide them by default and display them in the "advanced" section.

src/Form/AuthenticationForm.php
```diff
@@ -597,4 +500,155 @@ public function ajaxCallback(array $form, FormStateInterface $form_state): array
...
+
+  /**
+   * Moves form errors from one form state to another.
+   *
+   * @param \Drupal\Core\Form\FormStateInterface $from
+   *   The form state object to move from.
+   * @param \Drupal\Core\Form\FormStateInterface $to
+   *   The form state object to move to.
+   */
+  protected function moveFormStateErrors(FormStateInterface $from, FormStateInterface $to) {
```
mxr576, line 601: It is also something that you have copied from the Key module?

src/Form/AuthenticationForm.php
```diff
@@ -597,4 +500,155 @@ public function ajaxCallback(array $form, FormStateInterface $form_state): array
...
+      'obscured' => $obscured,
+      'current' => (!empty($obscured)) ? $obscured : $processed_original,
+    ];
+  }
+
+  /**
+   * Get the Form entity.
+   *
+   * Some of the legacy `key_input` plugins rely on this being an entity form so
+   * we need to return the active key as the entity..
```
mxr576: To much dots :)

src/OauthCredentials.php
```diff
@@ -29,44 +28,34 @@
...
-   *   token key does not implement EdgeOauthTokenKeyTypeInterface.
+   *   does not implement EdgeOauthKeyTypeInterface.
    */
-  public function __construct(KeyInterface $key, KeyInterface $key_token) {
-    parent::__construct($key);
+  public function __construct(KeyInterface $key) {
 
-    if (!($key->getKeyType() instanceof EdgeOauthKeyTypeInterface)) {
-      throw new \InvalidArgumentException("Type of {$key->id()} OAuth key does not implement EdgeOauthKeyTypeInterface.");
+    if ($key->getKeyType() instanceof EdgeKeyTypeInterface
```
mxr576, line 44: Inline these conditions.

src/OauthTokenFileStorage.php
```diff
@@ -0,0 +1,179 @@
...
+
+/**
+ * Storing and returning OAuth access token data.
+ */
+class OauthTokenFileStorage implements OauthTokenStorageInterface {
+
+  /**
+   * The storage location for oauth data.
+   */
+  const OAUTH_TOKEN_PATH = 'private://.apigee_edge/oauth.dat';
```
mxr576: Should not consistent in the path and the name of the files that we usage as a storage? We have `private://apigee_edge...` something above and we have this here (which kinda looks like it would have come from the Drupal 7 version).

mxr576: Also, should not we allow to change to location and the file name with a config in this storage at least? (Similarly how it worked in Drupal 7). If we use this _unencrypted_ storage by default, we should allow users to move this file to wherever they want on their server that they feel even more secure than the private file system.
Like the nobackup folder on ACQ: https://docs.acquia.com/acquia-cloud/manage/files/system-files/private/

@cnovak ^

src/Plugin/KeyType/ApigeeAuthKeyType.php
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    foreach ($this->getPluginDefinition()['multivalue']['fields'] as $id => $field) {
+      if (isset($field['required']) && $field['required'] === FALSE) {
+        continue;
+      }
+
+      $error_element = $form['settings']['input_section']['key_input_settings'][$id] ?? $form;
+
+      /** @var \Drupal\Core\StringTranslation\TranslatableMarkup $field */
```
mxr576: Is this annotation actually valid and useful? I see only usages of array elements from the $field array below.

src/Plugin/KeyType/ApigeeAuthKeyType.php
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getAuthenticationMethod(KeyInterface $key): Authentication {
+    $values = $key->getKeyValues();
+    if ($values['auth_type'] === EdgeKeyTypeInterface::EDGE_AUTH_TYPE_OAUTH) {
+      // Use Oauth authentication.
+      return new OauthAuthentication($this->getUsername($key), $this->getPassword($key), \Drupal::service('apigee_edge.token_storage'), NULL, $this->getClientId($key), $this->getClientSecret($key), NULL, $this->getAuthorizationServer($key));
```
mxr576: Inject it with the overriding `create()` method instead of calling `\Drupal::service()` directly.

mxr576:
@Jaesin is it still in WIP?

@cnovak if this should be part of beta1 then it should be wrapped up today. I'll add my review to this morning.

mxr576 (changes requested):

tests/modules/apigee_edge_test/src/Plugin/KeyProvider/TestEnvironmentVariablesKeyProvider.php
```diff
@@ -0,0 +1,56 @@
...
+class TestEnvironmentVariablesKeyProvider extends EnvironmentVariablesKeyProvider {
+
+  const KEY_VALUE_STATE_ID = 'apigee_edge_test_key_value';
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setKeyValue(KeyInterface $key, $key_value) {
+    // Store the value in state for functional callbacks.
+    \Drupal::state()->set(static::KEY_VALUE_STATE_ID, $key_value);
```
mxr576: Why don't you inject State service by overriding the `create()` method? Calling services statically in an OOP code is highly discouraged.

tests/src/Kernel/Form/AuthenticationFormTest.php
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Test generating a new auth key.
+   *
+   * @throws \Drupal\Core\Form\EnforcedResponseException
+   * @throws \Drupal\Core\Form\FormAjaxException
+   */
+  public function testGenerateNewAuthKey() {
+    $form_state = new FormState();
+    $form = \Drupal::formBuilder()->buildForm(AuthenticationForm::class, $form_state);
```
mxr576, line 88: Use DI instead of `\Drupal::formBuilder()`.

tests/src/Kernel/Form/AuthenticationFormTest.php
```diff
@@ -0,0 +1,113 @@
...
+    // The form should have created a new key and saved some empty values to it.
+    $active_key = Key::load($this->config(AuthenticationForm::CONFIG_NAME)->get('active_key'));
+    static::assertSame($active_key->id(), $form_state->getFormObject()->getEntity()->id());
+
+    $decoded = Json::decode($active_key->getKeyValue());
+
+    static::assertSame(EdgeKeyTypeInterface::EDGE_AUTH_TYPE_BASIC, $form["connection_settings"]["auth_type"]["#value"]);
+    static::assertEmpty($form["connection_settings"]["organization"]["#value"]);
+    static::assertEmpty($form["connection_settings"]["username"]["#value"]);
+    static::assertEmpty($form["connection_settings"]["password"]["#value"]);
```
mxr576: single quotes instead of double quotes...

mxr576 (changes requested):

apigee_edge.services.yml
```diff
@@ -123,3 +130,6 @@ services:
   apigee_edge.converter.user_developer:
     class: Drupal\apigee_edge\UserDeveloperConverter
     arguments: ['@config.factory', '@entity_type.manager', '@plugin.manager.apigee_field_storage_format', '@apigee_edge.converter.field_attribute']
+
+  apigee_edge.token_storage:
```
mxr576: I was a little bit sceptical whether this needs to be a service but at least this way we can swap it and replace it with a more secure storage solution if needed on customer projects.
Should not this be called `apigee_edge.oauth_token_storage` instead? Or `apigee_edge.authentication.oauth_token_storage`? "token_storage" is way too general I think.

apigee_edge.services.yml
```diff
@@ -10,6 +10,13 @@ services:
     factory: cache_factory:get
     arguments: [apigee_edge_entity]
 
+  cache.apigee_edge:
```
mxr576: Is it still necessary?

src/Form/AuthenticationForm.php
Jaesin: Change to `private://.apigee_edge/{$new_key_id}.json`.

src/OauthTokenFileStorage.php
Jaesin: Make configurable via `settings.php`

src/Plugin/KeyType/ApigeeAuthKeyType.php
Jaesin: The service is not used unless `auth_type` == `oauth` (memory optimization).

src/Form/AuthenticationForm.php
Jaesin: We are going to address this in another ticket. We intend to present links to the documentation for advance configuration and give suggestions for best practices.

tests/src/Kernel/Form/AuthenticationFormTest.php
Jaesin: This is a test class without `ContainerInjectionInterface`. I could use `$this->container->get('form_builder')` but this gets type me suggestions.

apigee_edge.services.yml
Jaesin: Not anymore. I removed it.

src/Form/AuthenticationForm.php
```diff
@@ -275,6 +275,12 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
 
     // Test the connection.
     if (!empty($test_key->getKeyValue())) {
+      /** @var \Drupal\apigee_edge\Plugin\KeyType\ApigeeAuthKeyType $test_key_type */
+      $test_auth_type = $test_key->getKeyType()->getAuthenticationType($test_key);
+      if ($test_auth_type === EdgeKeyTypeInterface::EDGE_AUTH_TYPE_OAUTH) {
+        // Clear existing token data.
+        $this->oauthTokenStorage->saveToken([]);
```
mxr576: According to the interface definition `removeToken()` should do this, but it is fine to keep it like this for now.
https://github.com/apigee/apigee-client-php/blob/2.x/src/HttpClient/Plugin/Authentication/OauthTokenStorageInterface.php#L96

src/Plugin/KeyProvider/PrivateFileKeyProvider.php
```diff
@@ -116,7 +116,12 @@ public function getKeyValue(KeyInterface $key) {
    *   Thrown when the key cannot be saved.
    */
   public function setKeyValue(KeyInterface $key, $key_value) {
-    if (!\file_unmanaged_save_data($key_value, $this->getFileUri($key), FILE_EXISTS_REPLACE)) {
+    $file_uri = $this->getFileUri($key);
+    $file_path = dirname($file_uri);
+    // Make sure the folder exists.
+    file_prepare_directory($file_path, FILE_CREATE_DIRECTORY | FILE_MODIFY_PERMISSIONS);
+    // Save the token data.
+    if (!\file_unmanaged_save_data($key_value, $file_uri, FILE_EXISTS_REPLACE)) {
```
mxr576: \ is not necessary.

src/OauthTokenFileStorage.php
```diff
@@ -139,7 +137,19 @@ public function removeToken(): void {
...
+    file_unmanaged_delete(static::oauthTokenPath());
+  }
+
+  /**
+   * Gets the storage location for OAuth token data.
+   */
+  public static function oauthTokenPath() {
+    static $token_path;
+
+    $token_path = $token_path
```
mxr576: [priority] Why not a config object in apigee_edge.auth.yml? That could be also overridden from settings.php and it can be synched if needed.

mxr576 (changes requested):

apigee_edge.module
```diff
@@ -1459,3 +1458,11 @@ function apigee_edge_get_my_developer_apps_title(?UserInterface $user = NULL): T
 
   return $title;
 }
+
+/**
+ * Implements hook_cache_flush().
+ */
+function apigee_edge_cache_flush() {
+  // Clear token data.
+  \Drupal::service('apigee_edge.authentication.oauth_token_storage')->saveToken([]);
```
mxr576: [priority] Why it should be flushed on cache clear?

src/Form/AuthenticationForm.php
```diff
@@ -597,4 +522,177 @@ public function ajaxCallback(array $form, FormStateInterface $form_state): array
...
+   */
+  protected function moveFormStateErrors(FormStateInterface $from, FormStateInterface $to) {
+    foreach ($from->getErrors() as $name => $error) {
+      $to->setErrorByName($name, $error);
+    }
+  }
+
+  /**
+   * Moves storage variables from one form state to another.
+   *
```
mxr576, line 609: This is also copied from the Key module.

src/Form/AuthenticationForm.php
```diff
@@ -597,4 +522,177 @@ public function ajaxCallback(array $form, FormStateInterface $form_state): array
...
+   */
+  protected function moveFormStateStorage(FormStateInterface $from, FormStateInterface $to) {
+    foreach ($from->getStorage() as $index => $value) {
+      $to->set($index, $value);
+    }
+  }
+
+  /**
+   * Creates a FormStateInterface object for a plugin.
+   *
```
mxr576, line 625: This is also copied from the Key module.

src/OauthCredentials.php
```diff
@@ -29,44 +28,34 @@
...
+      && ($auth_type = $key->getKeyType()->getAuthenticationType($key))
+      && $auth_type === EdgeKeyTypeInterface::EDGE_AUTH_TYPE_OAUTH
+    ) {
+      parent::__construct($key);
     }
-    if (!($key_token->getKeyType() instanceof EdgeOauthTokenKeyTypeInterface)) {
-      throw new \InvalidArgumentException("Type of {$key_token->id()} OAuth token key does not implement EdgeOauthTokenKeyTypeInterface.");
+    else {
+      throw new \InvalidArgumentException("The `{$key->id()}` key is not configured for OAuth.");
```
mxr576: Use the new module specific InvalidArgumentException instead of the global one.

src/OauthTokenFileStorage.php
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function removeToken(): void {
+    // Remove the data from the static cache.
+    static::$token_data = NULL;
+    // Remove the token data from storage.
+    file_unmanaged_delete(static::oauthTokenPath());
```
mxr576: Maybe this should not delete the file rather clear its content only. Just like `saveToken([])` does.

src/Plugin/KeyInput/BasicAuthKeyInput.php
```diff
@@ -35,6 +35,9 @@
  *   label = @Translation("Apigee Edge basic authentication credentials input fields."),
  *   description = @Translation("Provides input text fields for Apigee Edge basic authentication credentials.")
  * )
+ *
+ * @deprecated deprecated since version 8.x-1.0-alpha9.
```
mxr576: [priority] If something is deprecated let's remove it. We can do this before beta1.

mxr576 (changes requested):

src/Plugin/KeyInput/OauthKeyInput.php
```diff
@@ -34,6 +34,9 @@
  *   label = @Translation("Apigee Edge OAuth credentials input fields."),
  *   description = @Translation("Provides input text fields for Apigee Edge OAuth credentials.")
  * )
+ *
+ * @deprecated deprecated since version 8.x-1.0-alpha9.
```
mxr576: [priority] If something is deprecated let's remove it. We can do this before beta1.

src/Plugin/KeyProvider/PrivateFileKeyProvider.php
```diff
@@ -116,7 +116,12 @@ public function getKeyValue(KeyInterface $key) {
...
       throw new KeyValueNotSetException();
```
mxr576: Throwing an exception with an empty message is not helpful at all.

src/Plugin/KeyProvider/PrivateFileKeyProvider.php
```diff
@@ -116,7 +116,12 @@ public function getKeyValue(KeyInterface $key) {
    *   Thrown when the key cannot be saved.
    */
   public function setKeyValue(KeyInterface $key, $key_value) {
-    if (!\file_unmanaged_save_data($key_value, $this->getFileUri($key), FILE_EXISTS_REPLACE)) {
+    $file_uri = $this->getFileUri($key);
+    $file_path = dirname($file_uri);
+    // Make sure the folder exists.
+    file_prepare_directory($file_path, FILE_CREATE_DIRECTORY | FILE_MODIFY_PERMISSIONS);
```
mxr576, line 122: I'd check and log whether the directory could be created or not, not just whether the file could be saved.

src/Plugin/KeyType/BasicAuthKeyType.php
```diff
@@ -58,6 +58,9 @@
  *     }
  *   }
  * )
+ *
+ * @deprecated deprecated since version 8.x-1.0-alpha9.
```
mxr576: [priority] Do not leave deprecated code in the module!

src/Plugin/KeyType/OauthKeyType.php
```diff
@@ -71,6 +70,9 @@
  *     }
  *   }
  * )
+ *
+ * @deprecated deprecated since version 8.x-1.0-alpha9.
```
mxr576: Again.

mxr576 (changes requested):

src/OauthTokenFileStorage.php
Jaesin: Good idea

src/OauthTokenCacheStorage.php
cnovak: Changed to filesystem

apigee_edge.module
Jaesin: It's a cache item that is auto generated.

apigee_edge.module
```diff
@@ -1464,5 +1464,5 @@ function apigee_edge_get_my_developer_apps_title(?UserInterface $user = NULL): T
  */
 function apigee_edge_cache_flush() {
   // Clear token data.
-  \Drupal::service('apigee_edge.authentication.oauth_token_storage')->saveToken([]);
+  \Drupal::service('apigee_edge.authentication.oauth_token_storage')->removeToken([]);
```
mxr576: This method has empty parameter list.

apigee_edge.module
mxr576: I would not consider access token as a "cache item". Why we want to enforce re-authentication to Apigee Edge with username/pwd/client_id/client_secret on cache clear? Because this is what we achieve with this.

@cnovak What do you think?

src/Plugin/KeyProvider/PrivateFileKeyProvider.php
```diff
@@ -122,7 +122,7 @@ public function setKeyValue(KeyInterface $key, $key_value) {
     file_prepare_directory($file_path, FILE_CREATE_DIRECTORY | FILE_MODIFY_PERMISSIONS);
     // Save the token data.
     if (!file_unmanaged_save_data($key_value, $file_uri, FILE_EXISTS_REPLACE)) {
-      throw new KeyValueNotSetException();
+      throw new KeyValueNotSetException('Error saving connection settings to file.');
```
mxr576: I'd also add the file path to the error message because that could be helpful in debugging, but this is probably just nitpicking.

Jaesin: I think it's fine to show the file location since this should only ever happen for admins that have access to change connection settings.

src/OauthTokenFileStorage.php
Jaesin: It's true that using settings keeps this out of config. I thought that's what we wanted.

mxr576: No, because we would like to allow to sync this custom token location - for example when all servers on ACQ - but we do not want to expose this configuration option on the UI.

Jaesin: @cnovak What do you think, should this be exportable configuration or just something that would be in `settings.php`?

cnovak: I was thinking a config object the way we discussed earlier, I don't see any issue w/making it a config value.  I was thinking it would not be in the  apigee_edge.auth.yml however since this isn't something we need to store w/secure settings, but maybe that is OK to have it there

Jaesin: I changed it to `$config['apigee_edge.auth']['oauth_token_storage_location']`. There are only  two items in this config item now. The active key and the token storage location.

apigee_edge.module
Jaesin: It's can be used for sanitization or making sure the SDK client re-authenticates. There isn't much of a downside as far as I see it. Forcing Apigee Edge to re-authenticate is part of the point.

cnovak: I don't see much of a downside of using the refresh as a way to clear out the tokens, it may be the easiest approach to support the module if we find out any token errors are happening.

mxr576: I do agree that we should provide a way to flush out access token/refresh token the system for debug purposes, but probably we should have a dedicated button for that in the client configuration UI instead. Cache clear could happen multiple times when update hooks are running in Drupal (not just our module's update hooks) and my concern is that we can run into some issues if we send too many re-authentication requests to Apigee Edge when Drupal updates are running.
It is easy to add this later if we decide to do it this way, but it also works for me if it leave it here until the next release when we re-consider whether this is the right way to solve this.

src/OauthTokenFileStorage.php
```diff
@@ -144,8 +142,12 @@ public function removeToken(): void {
   public static function oauthTokenPath() {
     static $token_path;
 
-    $token_path = $token_path
-      ?? rtrim(Settings::get(static::OAUTH_TOKEN_FILE_PATH_SETTINGS_KEY, 'private://.apigee_edge'), " \t\n\r\0\x0B\\/") . '/oauth.dat';
+    if (!isset($token_path)) {
+      // Get the file location from config.
+      $location = \Drupal::config(AuthenticationForm::CONFIG_NAME)->get('oauth_token_storage_location');
```
mxr576, line 154: Inject the config factory instead.

src/OauthTokenFileStorage.php
```diff
@@ -144,8 +142,12 @@ public function removeToken(): void {
   public static function oauthTokenPath() {
     static $token_path;
```
mxr576, line 150: Do not use static here, set the location in a property in the constructor of this class.

mxr576 (dismissed):

src/OauthTokenFileStorage.php
mxr576: Even if you do not DI the config factory, the location should be stored in a property.

src/OauthTokenFileStorage.php
```diff
@@ -144,8 +142,12 @@ public function removeToken(): void {
...
 
-    $token_path = $token_path
-      ?? rtrim(Settings::get(static::OAUTH_TOKEN_FILE_PATH_SETTINGS_KEY, 'private://.apigee_edge'), " \t\n\r\0\x0B\\/") . '/oauth.dat';
+    if (!isset($token_path)) {
+      // Get the file location from config.
+      $location = \Drupal::config(AuthenticationForm::CONFIG_NAME)->get('oauth_token_storage_location');
+
+      $token_path = (!empty($location) ? rtrim($location, " \t\n\r\0\x0B\\/") : 'private://.apigee_edge') . '/oauth.dat';
```
mxr576, line 156: Are you sure rtrim() is necessary here? I do not think the path needs to be sanitized this much. If someone sets an invalid location value that will trigger an error when the system tries to save the token.

src/OauthTokenFileStorage.php
Jaesin: It's a static method.

src/OauthTokenFileStorage.php
```diff
@@ -0,0 +1,193 @@
...
+   * @var int
+   */
+  protected $leeway = 30;
+
+  /**
+   * An internally cached token data store.
+   *
+   * @var array
+   */
+  private static $token_data;
```
mxr576, line 49: This is a service, therefore it is like a singleton (unless it is a shared service) in a page request. So you do not need a static variable as a cache, non static variables also behaves as cache and stores the information until the object has been destructed (page request is active).

src/OauthTokenFileStorage.php
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+namespace Drupal\apigee_edge;
+
+use Apigee\Edge\HttpClient\Plugin\Authentication\OauthTokenStorageInterface;
+use Drupal\apigee_edge\Form\AuthenticationForm;
```
mxr576, line 23: This is why I questions whether that config name constant should be a public inside this class. Why a service depends on a form class? :) I haven't seen like this in Drupal core, also Drupal core does not define constant for config objects, not even in config forms. So this looks weird to me, but this can be fixed later.

src/OauthTokenFileStorage.php
mxr576: But this is a service so it does not have to be static.

src/OauthTokenFileStorage.php
cnovak: LGTM

src/OauthTokenFileStorage.php
cnovak: LGTM

cnovak:
@mxr576 I validated the CLAs look good, and your change looks good too

cnovak (approved):

PR #86, opened 2018-11-16
Refactor entity architecture - introduce controller services

mxr576:
Covers: DRUP-423 DRUP-424 DRUP-414 DRUP-434

This is a huge PR again but I created small, detailed commits. The order the of the commits also helps in the review process.

# Add to release notes
* Controllers in Drupal\apigee_edge\Entity\Controller\ now
	* decorates (encapsulates) the PHP SDK controllers and not extend them,
	* services or they have a factory service to initiate them (check apigee_edge.services.yml)
	* return PHP SDK entities always, entity storage implementation converts them to Drupal entities.
* Entities in  Drupal\apigee_edge\Entity now decorates SDK entities not extend them.
* FieldableEdgeEntityBaseTrait got deprecated in favor of FieldableEdgeEntityBase. This class provides a base class for all Apigee Edge entities that support field UI.
* AttributesAwareFieldableEdgeEntityBase is a new base class for those Apigee Edge (Management API) entities that support attributes and their field values can be save to attributes in Apigee Edge.
* Attribute as field storage support also moved from FieldableEdgeEntityStorageBase to a new AttributesAwareFieldableEdgeEntityStorageBase class.
* User to developer conversion does not leave empty attributes on developers if a converted field value is an empty string.
* Minor improvements in logging meanwhile developer synchronization.
* New module specific Runtime- and InvalidargumentExceptions.

# TODOs
* ~Fix failing tests if any.~
* ~Update test classes to use new controller services.~
* ~Finalize "Add to release notes" section.~
* ~Developer fields leave empty attributes on developer entities their values gets removed. (Probably an easy fix.)~
* ~Fix TODO in `\Drupal\apigee_edge\Entity\Storage\EdgeEntityStorageBase::invokeStorageLoadHook()`~


tamasd (approved):

mxr576:
@Jaesin @cnovak I really would like to get an input from you about the content of this PR, because if all tests are going to pass I'd like to merge it to dev-

Jaesin:
Is this still [WIP]?

mxr576:
Yes @Jaesin it was in WIP state because there was one failing test still that I had to fix. Yesterday, I have just want to ask you to start reviewing this and sharing your questions/comments (if there is any) before I would finish with fixing the failing tests.

Now there is only two issues that I am still investigating:
a) New MYSQL server has gone away error when drupal_flush_all_caches() gets called in a test.
b) For the first sight, even if we should have better caching after these changes, according to Travis our test suites became slower again. I am investigating whether this issue is in the Drupal 8 module or is it caused by a performance issue on Apigee Edge.

So if there are no concerns until tomorrow I'll merge this PR as is.

https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/457416624

mxr576:
Based on the result of the test build in my previous comment I think this PR is ready to be merged. I removed all data from my test org and that fixed the performance issue as well.

Jaesin:
Thanks @mxr576, I'll look this over today. `|8^)`.

src/Entity/EdgeEntityBase.php
```diff
@@ -0,0 +1,158 @@
...
+   *
+   * @return string
+   *   FQCN of the entity class.
+   */
+  abstract protected static function decoratedClass() : string;
+
+  /**
+   * {@inheritdoc}
+   *
+   * We have to override this to make it compatible with the SDK's
```
Jaesin, line 82: Does the drupal entity really need to implement `\Apigee\Edge\Entity\EntityInterface`? This could be changed in the SDK.

src/Entity/FieldableEdgeEntityBase.php
```diff
@@ -0,0 +1,544 @@
...
+    // because they can return the correct information.
+    // Because the current object implements the interface of the decorated
+    // object there should be any getter on the decorated object that does not
+    // have a decorator in the current class (that potentially also calls to the
+    // decorated getter method under the hood.)
+    if (method_exists($this, $getter)) {
+      return call_user_func([$this, $getter]);
+    }
+
+    return NULL;
```
Jaesin, line 275: Is there a good reason the Drupal entity needs to implements the SDK entities interface. That's allot of getters and setters to write. If they don't really need to look exactly like SDK entities, we can fall back to `elseif (method_exists($this->decorated, $getter)) {` after `if (method_exists($this, $getter)) {`.

mxr576:
@Jaesin I am not sure whether you had time to review this and just forgot to approve it or not but I am going to merge this PR now. Let me know if you find anything later. :)

Jaesin (commented):
`ContentEntityBase` has `public function &__get($name) {`. It is very common to access properties via `$entity->propertyName`. It would be nice to have.

src/Entity/FieldableEdgeEntityBase.php
Jaesin: `\Drupal\apigee_edge\Entity\FieldableEdgeEntityBase::getProperties()` searches for `get` and `is` but this only looks for `get`.

src/Entity/EdgeEntityBase.php
mxr576: Interfaces should not change after the stable release, they define the public API of our software. So this is fine.

mxr576:
@Jaesin 
> ContentEntityBase has public function &__get($name) {. It is very common to access properties via $entity->propertyName. It would be nice to have.

My problem is with magic getters that there is no IDE that would support them. No auto-complete, no static analyses that could figure out that you are trying to access to a non-existing property like this. (Of course with `get()` you also does not know whether the property exist or not.)
My point is, you only spare 5 characters but your code does not get better form that.
Moreover, we inherited all property getters from the SDK entity classes, so it is easier to call them instead of `get()` or magic get. Fields in Edge entity classes use properties from the decorated SDK entity object as a storage so it does not matter whether you call `$developer->getName()` or `$developer->get('name')->value`, but autocomplete perfectly works on these inherited property getters.

src/Entity/FieldableEdgeEntityBase.php
mxr576: First comment: 
This how decorator pattern is working :) Decorator object should implement the interface (public behavior) of the decorated object.
At this moment we are not leveraging this behavior but it could be really useful later, for example when we have to transform the returned data from the decorated object before we would return it. Also, we may need to use Drupal entity object in a context where the SDK expects an SDK interface, like in SDK methods calls.

Second comment:
I do not get your point, "but this only looks for get.". Can you rephrase it?

Jaesin: Maybe it should be something like: 

```
  /**
   * {@inheritdoc}
   */
  protected function getFieldValue(string $field_name) {
    $getter = 'get' . ucfirst($field_name);
    $boolean_accessor = 'is' . ucfirst($field_name);

    // We call the getters on the current object instead of the decorated one
    // because they can return the correct information.
    // Because the current object implements the interface of the decorated
    // object there should be any getter on the decorated object that does not
    // have a decorator in the current class (that potentially also calls to the
    // decorated getter method under the hood.)
    if (method_exists($this, $getter)) {
      return call_user_func([$this, $getter]);
    }
    elseif (method_exists($this, $boolean_accessor)) {
      return call_user_func([$this, $boolean_accessor]);
    }

    return NULL;
  }
```

mxr576: Ahh, now I get it which method you are talking about. Good catch! That could be problematic in the MINT entities. MGMT entities do not really have issers.

mxr576: https://github.com/apigee/apigee-edge-drupal/pull/88

PR #91, opened 2018-11-27
Licence information fix

mxr576:
```sh
$ composer validate
./composer.json is valid, but with a few warnings
See https://getcomposer.org/doc/04-schema.md for details on the schema
License "GPL-2.0" is a deprecated SPDX license identifier, use "GPL-2.0-or-later" instead
The package "behat/mink" is pointing to a commit-ref, this is bad practice and can cause unforeseen issues.
```

Drupal core also uses the same license: https://github.com/drupal/core/blob/8.6.x/composer.json#L5

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "drupal/apigee_edge",
-    "license": "GPL-2.0",
+    "license": "GPL-2.0-or-later",
     "type": "drupal-module",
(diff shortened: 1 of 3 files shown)


cnovak (approved):
I do not see any behat/mink change, but rest looks good

mxr576:
We need that specific version from the behat/mink library so we can not resolve that issue.

PR #94, opened 2018-11-28
Oauth token storage fixes

mxr576:
@Jaesin Please see the remaining TODO items below.

TODOs:
* Check TODOs in AuthenticationForm.php.
* UX bug: "Send request" clears the Password field therefore I have to set the password again before I could save the form - because I have to set my password again I can make a typo in that, so even if "Send request" returned connection successful I can get an form/connection error when I try to save the form. This has not worked like this before.
* Validate how the new Authentication form reacts when:
  * The active key is not the default key that the Authenticaton form created. Someone overrode it by modifying the related CMI key value.
  * The key storage of the default key has been modified from private file system to something like. For example to the environment that does not provide any input fields. Should "Send request" button work in this case as well to allow validation of the authentication credentials coming from the env vars?
* Remove the OAuth token storage file if the authentication method changes from Oauth to HTTP. We should not leave remnants in the database/filesystem.
* Double check whether apigee_edge.schema.yml contains any remnant data related to authentication keys.

https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/463803995


tamasd (approved):

apigee_edge.install
```diff
@@ -90,6 +91,19 @@ function apigee_edge_install() {
...
+/**
+ * Implements hook_uninstall().
+ */
+function apigee_edge_uninstall() {
+  // Remove OAuth token data file.
+  // We do not check whether OAuth was the authentication method in use
+  // because this function does not trigger an error if it was not.
+  $storage = \Drupal::service('apigee_edge.authentication.oauth_token_storage');
+  if ($storage instanceof OauthTokenFileStorage) {
+    $storage->removeTokenFile();
```
Jaesin, line 103: Good idea.

config/install/apigee_edge.auth.yml
```diff
@@ -1 +1,2 @@
 active_key: ''
+oauth_token_storage_location: ''
```
Jaesin, line 2: A default doesn't have to be set for this config item. If it is, it should probably be the default value.

mxr576:
We had a call with @Jaesin and we agreed this PR should be merged and follow up PRs should be created to fix the remaining TODO items.

@Jaesin please add your review or accept this PR.

cc @cnovak

cnovak:
Sounds good @mxr576

apigee_edge.module
```diff
@@ -1463,6 +1464,18 @@ function apigee_edge_get_my_developer_apps_title(?UserInterface $user = NULL): T
  * Implements hook_cache_flush().
  */
 function apigee_edge_cache_flush() {
-  // Clear token data.
-  \Drupal::service('apigee_edge.authentication.oauth_token_storage')->removeToken();
+  // Clear OAuth token storage and enforce re-authentication to Apigee Edge.
+  $active_key_id = \Drupal::config('apigee_edge.auth')->get('active_key');
+  if ($active_key_id) {
```
Jaesin: Checking the active key is an oauth is not necessary and would leave token data to persist if the active key has recently changed. As in `apigee_edge_uninstall` it does not trigger an error to clear the data. What is achieved for checking the active key?

src/Form/AuthenticationForm.php
```diff
@@ -279,11 +278,15 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
...
-        $this->oauthTokenStorage->removeToken();
-      }
       try {
+        if ($test_auth_type === EdgeKeyTypeInterface::EDGE_AUTH_TYPE_OAUTH) {
+          // Check the requirements first.
+          $this->oauthTokenStorage->checkRequirements();
+          // Clear existing token data.
+          $this->oauthTokenStorage->removeToken();
+        }
+        // TODO If email field contains an invalid email request should
```
Jaesin: What is this comment about? We aren't sending an email here.

tests/src/Kernel/OauthTokenFileStorageTest.php
```diff
@@ -0,0 +1,235 @@
...
+    $storage = $this->tokenStorage();
+    // Save the token.
+    $storage->saveToken($this->testTokenData);
+    $this->assertTrue(file_exists($this->tokenFileUri()), 'Oauth token data file should not exist after the module got uninstalled.');
+    /** @var \Drupal\Core\Extension\ModuleInstallerInterface $installer */
+    $installer = $this->container->get('module_installer');
+    $installer->uninstall(['apigee_edge']);
+    $this->assertFalse(file_exists($this->tokenFileUri()), 'Oauth token data file should not exist after the module got uninstalled.');
+  }
+
```
Jaesin, line 232: Is there a good reason the `testCacheClear` test was removed? There should still be coverage for that.

src/OauthTokenFileStorage.php
```diff
@@ -19,16 +19,21 @@
...
+use Drupal\Core\Logger\LoggerChannelInterface;
+use Drupal\Core\Site\Settings;
 
 /**
- * Storing and returning OAuth access token data.
- *
- * @internal The implementation may change before the first stable release.
+ * Stores OAuth token data in a file.
  */
 final class OauthTokenFileStorage implements OauthTokenStorageInterface {
```
Jaesin, line 30: Why is this service marked as final?

src/OauthTokenFileStorage.php
```diff
@@ -116,85 +160,93 @@ public function markExpired(): void {
...
    */
-  public static function oauthTokenPath() {
-    static $token_path;
-
-    if (!isset($token_path)) {
-      // Get the file location from config.
-      $location = \Drupal::config(AuthenticationForm::CONFIG_NAME)->get('oauth_token_storage_location');
-
-      $token_path = (!empty($location) ? rtrim($location, " \t\n\r\0\x0B\\/") : 'private://.apigee_edge') . '/oauth.dat';
+  public function removeTokenFile(): void {
```
Jaesin, line 209: Should this be added to the interface?

apigee_edge.module
mxr576: For start, this method calls `removeToken()` not `removeTokenFile()`. There is a huge difference between them.

To answer your question, there are multiple reasons why we have to do this sanity check:
* The module should not always trigger an Oauth token cache clear.
  * Not when the module has just installed and 100% it has not been configured yet. Calling `removeToken()` creates the token file it has not existed yet. This file should always exists if Oauth authentication is in use.
 * The private file system in Drupal is not configured by default. So if cache clear tries to clear the token file if the private file system is not configured (when the module is installed for example or has not been configured yet) because "private://" scheme in Drupal only exist if private filesystem path is configured. That is also going to trigger error.
  * Not when the HTTP authentication method is in use - it should not create an empty token storage file, etc.
  * ...

So basically we are doing a shortcut by not doing a check in the `hook_uninstall()` to see whether private filesystem was configured or not and we just assume it was.

config/install/apigee_edge.auth.yml
mxr576: It should be set, you should not just configure an object in the scheme as string and you leave it as null by not defining that. Yes this could be the default location instead of an empty string, both options can be justified with good reasons why it should be like that. I would just leave it like this for now.

src/Form/AuthenticationForm.php
mxr576: Meh, not email request. The "Send request" button should not send an request to Apigee Edge if the form validation fails, like when the content of the email field is not an email

src/OauthTokenFileStorage.php
mxr576: Our service _implementations_ are not for extension. If someone would like to define its own implementation for a service s/he should create a new class and implement the service's interface definition. That describes the requirements of a service and not a concrete class implementation. This is the reason why all services should have an interface definition with their basic features (public methods).
By marking a class as final we are not making a class non-reusable, we just driving the developers to the right directions, forcing them to prefer encapsulation over inheritance when they would like build something in the top of that class.
Thanks for this approach keeping BC between releases will be much easier for us. We can change the internal behavior of our classes (property names, methods, etc.) as we want/need until we keep out class implementations compatible with the interfaces that they implement.

src/OauthTokenFileStorage.php
mxr576: It should not, because there is no `OauthTokenFileStorageInterface`, only a `OauthTokenStorageInterface` which defines the basic requirements of an Oauth token store service implementations but it does not "care" about the actual implementation of them. (How and where they store the token data - removeTokenFile() is related to _how_ the data is being stored.)
This concrete implementation stores the Oauth token data in a file, a custom implementation can store it in a config object and the implementation could have a public `removeConfigObject()`.

tests/src/Kernel/OauthTokenFileStorageTest.php
mxr576: See the related  TODO on the top of the file: https://github.com/apigee/apigee-edge-drupal/pull/94/files#diff-0a036b86ad28ed8eb12935d8098980e0R33

mxr576:
Conflict resolved.

src/OauthTokenFileStorage.php
mxr576: When you define a service you should keep in my that it either gets **decorated** or **overridden** by external developers if it is needed. If you define an interface for a service you can force developers to keep the contract the the service's original implementation describes and all usages of that services can depend on that. Also, do not forget you should use interfaces in type hints wherever you require a service.

mxr576:
Changed the logic in `removeToken()` based on what we have discussion with @Jaesin in our daily standup. If this test passes I'll merge this PR. This PR also addresses DRUP-482.

https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/466388533

tamasd (approved):

mxr576:
Changed the destination branch to the devteams. Waiting for info from @cnovak to confirm the new workflow.

mxr576:
Tests passed. merging the PR.

PR #95, opened 2018-11-28
Add APIGEE_EDGE_AUTH_TYPE var to travis builds

cnovak:
Add APIGEE_EDGE_AUTH_TYPE and update prepare-test-env.sh to validate the var is set.  Also changed validation to validate each environment variable separately.


mxr576 (approved):

cnovak:
Here is the build, one test is failing, we do have an issue in for fixing that test: https://travis-ci.org/cnovak/apigee-edge-drupal/builds/460844333

PR #99, opened 2018-12-11
DRUP-478 Company API integration - pre-requirements, part 1

mxr576:
~This PR contains all changes that we had to do on the Apigee Edge module when we worked on a company API integration. I created this PR to keep open PR's size as small as possible. So the Company API integration is coming in a separated PR after this PR got merged.~

~This branch is based on #97 because controller level entity caching is also a pre-requirement of the Company API integration. If #97 gets merged the size of this PR automatically gets reduced.~

~First commit of this PR is: d1c3be033d5. I'd recommend to start reviewing the PR commit by commit until #97 has not been merged.~

~WIP - More changes are coming to this PR around developer app entity related storages/forms. New parent classes gets introduced.~

Latest Travis build: https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/469515166


src/Entity/Form/DeveloperAppCreateFormForDeveloper.php
```diff
@@ -84,52 +73,71 @@ public function __construct(DeveloperAppCredentialControllerFactoryInterface $ap
...
+  }
+
+  /**
+   * The user from the route or the current user if it is not available.
+   *
+   * @return \Drupal\user\UserInterface
+   *   User object.
+   */
+  protected function getUser(): UserInterface {
+    return $this->user ?? $this->entityTypeManager->getStorage('user')->load(\Drupal::currentUser()->id());
```
tamasd: Can you inject the `current_user` service?

src/Entity/Form/DeveloperAppFormTrait.php
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    return (bool) $app;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function appEntityDefinition(): EntityTypeInterface {
+    return \Drupal::entityTypeManager()->getDefinition('developer_app');
```
tamasd, line 81: Can you inject the entity type manager here? E.g. with an abstract method.

src/Entity/Form/DeveloperAppFormTrait.php
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function appOwnerEntityDefinition(): EntityTypeInterface {
+    return \Drupal::entityTypeManager()->getDefinition('developer');
```
tamasd, line 88: Can you inject the entity type manager here? E.g. with an abstract method.

src/Entity/Form/DeveloperAppFormTrait.php
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function appCredentialLifeTime(): int {
+    return \Drupal::config('apigee_edge.developer_app_settings')->get('credential_lifetime');
```
tamasd, line 95: Can you inject the configuration here? E.g. with an abstract method.

src/Entity/Form/DeveloperAppCreateFormForDeveloper.php
mxr576: I can, but because this is a fallback solution (that should never be triggered) I did not want to add an extra parameter to the constructor just for this.

src/Entity/Form/DeveloperAppFormTrait.php
mxr576: I can and I started to implement it a way like that but then I realised creating n+1 abstract method definitions, keeping them in sync and then resolving them in classes that uses this trait seems less maintainable. "Luckily" Drupal 8 provides these beautiful static methods to access services. So it seems to be a fair trade for better maintainability. In a non-Drupal 8 code I'd probably choose a different solution.

As a fallback solution, if we really would like to mock the entity type manager somewere an $entityTypeManager property can be introduced in this trait (but then its visibility must be the same as in all classes that uses this trait) and the could above can be altered like `($this->entityTypeManager ?? \Drupal::entityTypeManager())->getDefinition('developer_app');`

tamasd (approved):

peterserfozo (approved):

mxr576:
@cnovak I think it would be better to introduce these changes in the 8.x-1.x branch (in the next upcoming release) even if we may not finish with the Company API integration in the devteams branch still. Some of these changes are "BC breaking" changes that we should deliver to customers rather sooner than later. Do you agree?

cnovak:
Yes @mxr576 we probably should get out changes like that sooner than later.  The title of this PR sounds like it is part of devteams, is that not true?  Can we get info of what JIRA tickets/features are part of this PR?  If it is more than one feature/ticket, would be good to have it broken into parts.  Also, is this a work in progress at this point or is it complete?  I see that it has been approved.

mxr576:
@cnovak The complete epic is related to this PR: DRUP-478 (which does not yet contain team app related user stories but I have already finished with some of them... two birds with one stone.)

All commits in this PR contains a small set of change with detailed descriptions about the reasons (except commits that fixes failing tests) therefore this PR can be reviewed commit by commit easily.

mxr576:
Conflict resolved with rebase.

peterserfozo (approved):

mxr576:
All tests passed, PR is ready to be merged.

PR #102, opened 2018-12-18
DRUP-478 Company API integration - pre-requirements, part 2

mxr576:
Second batch of changes that we needs to apply on the Apigee Edge module before we could add the sub module that provides the company API integration.

This is a follow up on #99, ~this needs to be re-based after #99 gets merged.~

Highlights:
* (Developer) apps related admin menus restructured. The new structure provides a better seperation between common app settings and developer app specific settings.
* [BC breaking change] Callback URL validation configuration move  to common app
settings, because we could not easily build re-usable a solution that could allow to set different settings for team- and developer apps by using field widget configuration form.
* getCompanies() on Drupal developer entities should always return the actual list of companies that a developer is currently member.
* Fixes and improvements around developer and developer app related entity cache invalidation.
* Introducing a new event based solution for alter Apigee Edge entities' "Manage fields" UI.

Necessary hook_update_N()-s are coming to the BC breaking changes before the next beta release IF we can provide upgrade path to all BC breaking changes that we have introduced since beta1.

Latest Travis build: https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/476721185

new_app_menu_structure_demo


src/Entity/Developer.php
```diff
@@ -223,7 +250,58 @@ public function hasApp(string $appName): bool {
    * {@inheritdoc}
    */
   public function getCompanies(): array {
-    return $this->decorated->getCompanies();
+    // If companies is null it means the original API response that this
```
mxr576, line 254: @peterserfozo Could you please write a test for this method to ensure it always works as it is described?

mxr576:
Rebased after #99 had been merged.

peterserfozo (approved):

peterserfozo (approved):

PR #103, opened 2018-12-19
[DRUP-499] Authentication form improvements

Jaesin:
Answers https://github.com/apigee/apigee-edge-drupal/pull/94#issue-234252394

Remaining issues:
1. (review comments inline)
2. Enabled the module on a site where private FS has not been configured yet. Opened `/admin/config/apigee-edge/settings` and I got a WSOD.
image
`Drupal\key\Exception\KeyValueNotSetException: Error saving connection settings to file (). Make sure the directory is writable. in Drupal\apigee_edge\Plugin\KeyProvider\PrivateFileKeyProvider->setKeyValue() (line 126 of /var/www/html/web/modules/contrib/apigee_edge/src/Plugin/KeyProvider/PrivateFileKeyProvider.php).`
The same happens if the private FS configured BUT incorrectly and the file could not be created.
Why do we create a file immediately when the configuration page gets opened? Should not it just happen after a form submission? 
3. `Notice: Only variables should be passed by reference in Drupal\apigee_edge\Form\AuthenticationForm->ajaxCallback() (line 568 of /var/www/html/web/modules/contrib/apigee_edge/src/Form/AuthenticationForm.php`
4. (Minor) If I uninstall the module the content of the .apigee_edge folder gets deleted but the folder does not.
5. (Minor) Why the endpoint is escaped? (maybe this is an old one)
image
6. (Idea) Good news, environment key storage seems to be working again. The description looks good, maybe I'd add a link to the active Apigee Edge key's edit form where I a user could see the active providers description.
image


src/Form/AuthenticationForm.php
```diff
@@ -145,13 +159,16 @@ protected function getEditableConfigNames() {
...
-    $form['#prefix'] = '<div id="apigee-edge-auth-form">';
-    $form['#suffix'] = '</div>';
     $form['#attached']['library'][] = 'apigee_edge/apigee_edge.admin';
 
     // Save the key for later use.
     $this->activeKey = $active_key = $this->getActiveKey();
 
+    // Placeholder for messages.
+    $form['messages'] = [
+      '#markup' => '<div id="apigee-edge-auth-form-messages"></div>',
```
mxr576: I'd rather use the `html_tag` render element to prepare this placeholder for messages.

src/Form/AuthenticationForm.php
```diff
@@ -186,6 +203,10 @@ public function buildForm(array $form, FormStateInterface $form_state) {
         ':input[name="key_input_settings[username]"]' => ['filled' => TRUE],
       ],
     ];
+    // Placeholder for debug.
+    $form['debug_placeholder'] = [
+      '#markup' => '<div id="apigee-edge-auth-form-debug-info"></div>',
```
mxr576: The same comment as above.

src/Form/AuthenticationForm.php
```diff
@@ -199,6 +220,8 @@ public function buildForm(array $form, FormStateInterface $form_state) {
     ];
 
     $form['test_connection'] = [
+      '#prefix' => '<div id="apigee-edge-connection-info">',
```
mxr576: I usually prefer using`container` render element for creating the AJAX wrapper instead of prefix+suffix.
https://www.metaltoad.com/blog/avoiding-drupal-7-ajax-pitfalls
https://github.com/drupal/core/blob/17557c97a816bc78e70989a86064c562bc27027f/modules/editor/editor.module#L137

src/Form/AuthenticationForm.php
```diff
@@ -456,77 +490,85 @@ protected function createSuggestion(\Exception $exception, KeyInterface $key): M
...
+    $response = new AjaxResponse();
+    // Get any status messages so they can be rendered in the placeholder.
+    $status_messages = $this->renderer->render(StatusMessages::renderMessages());
+    // Clear any existing messages from the initial page load.
+    $response->addCommand(new ReplaceCommand('div.messages', ''));
+    $response->addCommand(new ReplaceCommand('#apigee-edge-auth-form-messages', "<div id=\"apigee-edge-auth-form-messages\">{$status_messages}</div>"));
+
+    // Only render the debug element if it's been enabled by validation.
+    if ($form['debug']['#access']) {
+      // Add a prefix to the debug section to replace the wrapper.
```
mxr576: But this "wrapper" could have been set on the form element even if it is not visible by default, isn't it?
https://github.com/apigee/apigee-edge-drupal/pull/103/files#diff-c1ab508d8adb4da185ab3867ced29eb6R210

src/OauthTokenFileStorage.php
```diff
@@ -200,7 +200,14 @@ public function checkRequirements(): void {
    * {@inheritdoc}
    */
   public function removeToken(): void {
-    $this->saveToken([]);
+    // Prevents a token file from being created if it doesn't already exist.
+    if (file_exists($this->tokenFilePath)) {
```
mxr576: It is not enough to call `file_exist()`, you must run the same sanity checks as you can see in `removeTokenFile()` otherwise calling this function with a insufficiently configured private filesystem can generate errors. This the reason why I tried to call `saveToken([])` here simple and eliminate calling this method in `apigee_edge_cache_flush()` when it should not be call.

src/OauthTokenFileStorage.php
```diff
@@ -200,7 +200,14 @@ public function checkRequirements(): void {
...
+      $this->saveToken([]);
+    }
+    else {
+      // Removes cached token data.
+      unset($this->tokenData);
```
mxr576: tokenData must be an array, always, check the annotation.

mxr576 (changes requested):

mxr576:
@peterserfozo Could you please download this patch and test it in your local development environment to make sure we did not miss to solve any edge cases? Also, you created the debug UI for the connection configuration form so you have more insight about what have been displayed there and whether that information can be still captured. Thanks.

src/OauthTokenFileStorage.php
mxr576: Isn't this "fix" got required **only** because you had added this call to the authentication form?
https://github.com/apigee/apigee-edge-drupal/pull/103/files#diff-c1ab508d8adb4da185ab3867ced29eb6R364

src/Form/AuthenticationForm.php
```diff
@@ -456,77 +490,85 @@ protected function createSuggestion(\Exception $exception, KeyInterface $key): M
...
+    ], [
+      '$1***refresh-token***$3',
+      '$1***mfa-token***$3',
+      '$1***password***$3',
+      '$1***credentials***',
+    ], (string) $exception);
+
+    return json_encode($credentials, JSON_PRETTY_PRINT) . PHP_EOL .
+      json_encode($keys, JSON_PRETTY_PRINT) . PHP_EOL .
+      json_encode($this->config('apigee_edge.client')->get(), JSON_PRETTY_PRINT) . PHP_EOL .
```
peterserfozo: It prints the config hash too:
`{
   “http_client_connect_timeout”: 30,
   “http_client_timeout”: 30,
   “http_client_proxy”: “”,
   “_core”: {
       “default_config_hash”: “k8IHdAQdw4tE74jPYyd8LGjY0mbjOBGrj5bBzhGISf8”
   }
}`

src/Form/AuthenticationForm.php
```diff
@@ -326,12 +346,26 @@ public function validateForm(array &$form, FormStateInterface $form_state) {
   public function submitForm(array &$form, FormStateInterface $form_state) {
     // Get the processed value from the form state.
     $processed_submitted = $form_state->get('processed_submitted');
+    $active_key = $this->activeKey;
 
-    if (!empty($processed_submitted) && $this->keyIsWritable($this->activeKey)) {
+    if (!empty($processed_submitted) && $this->keyIsWritable($active_key)) {
```
peterserfozo, line 395: It seems the environment variables key provider does not work properly because the class `EnvironmentVariablesKeyProvider` implements the interface `KeyProviderSettableValueInterface`. Because it is checked whether the provider class implements that interface, the key is not loaded correctly so the connection cannot be tested using the "Send request" button.
It works correctly on my local development environment if I remove that interface name from the interface list and the functions `setKeyValue()` and `deleteKeyValue()` in the class `EnvironmentVariablesKeyProvider`.

src/Form/AuthenticationForm.php
Jaesin: I don't mind changing it but `\Drupal\Core\Render\Element\HtmlTag::preRenderHtmlTag()` converts a tag render element to `['#prefix' => '<div>', '#suffix' => '</div>',...]` probably making it less performant.

mxr576: But it is much easier to alter in a 3rd party module, because it becomes html_tag is a render element where the value and how the value is rendered is separated from each other. (Probably this form is never going to be altered, but we can not be sure about that. I always prefer to use html_tag instead of markup every time when markup's value would contain html tags.

src/Form/AuthenticationForm.php
mxr576: This is not a major one, but it would be nice if the `_core` key and its value would not be printed.

src/Form/AuthenticationForm.php
mxr576: @Jaesin Could you fix this one too? Environment variable key storage provided by the module should work properly.

src/OauthTokenFileStorage.php
mxr576: @Jaesin Please investigate this one too.

Jaesin: I dont' think there is anything to fix here. `::saveToken` handles validation.

src/Form/AuthenticationForm.php
Jaesin: This is true. I was trying to solve for a case before I started returning an ajax response instead of a form element.

src/Form/AuthenticationForm.php
Jaesin: This brings up the question why is `EnvironmentVariablesKeyProvider` settable?

Jaesin: This has been addressed by making sure `keyIsWritable()` checks the input type isn't `none`.

src/Form/AuthenticationForm.php
Jaesin: I filtered out anything beginning with an `_`.

src/OauthTokenFileStorage.php
mxr576: Pulled this patch and run an updb commend, I could run a "drush cr" command too on a site that private filesystem has not been configured. I got the error that I referred to in my first comment. `saveToken()` may handles this issue but `file_exists()` triggers an error if you call it in an non-registered file stream wrapper.

```
$ drush updb -y 
 --------------- -------------------------- --------------- --------------------------------------------------------------------------------- 
  Module          Update ID                  Type            Description                                                                      
 --------------- -------------------------- --------------- --------------------------------------------------------------------------------- 
  admin_toolbar   8001                       hook_update_n   Rebuild routes to mitigate issue 2938884.   @see                                 
                                                             https:www.drupal.orgprojectadmin_toolbarissues2938884                            
  node            node_revision_views_data   post-update     Clear caches due to updated views data.                                          
  system          states_clear_cache         post-update     Force cache clear to ensure aggregated JavaScript files are regenerated.   @see  
                                                             https:www.drupal.orgprojectdrupalissues2995570                                   
 --------------- -------------------------- --------------- --------------------------------------------------------------------------------- 

 // Do you wish to run the specified pending updates?: yes.                                                             

 [notice] Update started: admin_toolbar_update_8001
 [error]  The "company" entity type does not exist. 
 [error]  Update failed: admin_toolbar_update_8001 
 [warning] file_exists(): Unable to find the wrapper "private" - did you forget to enable it when you configured PHP? OauthTokenFileStorage.php:204
 [warning] file_exists(): Unable to find the wrapper "private" - did you forget to enable it when you configured PHP? OauthTokenFileStorage.php:204
```

tests/src/FunctionalJavascript/Form/AuthenticationFormTest.php
```diff
@@ -137,69 +145,150 @@ public function testAuthenticationForm() {
...
     // Make sure the form is still enabled.
     $this->assertFalse($this->cssSelect('input[data-drupal-selector="edit-test-connection-submit"]')[0]->hasAttribute('disabled'));
     $this->assertFalse($this->cssSelect('input[data-drupal-selector="edit-submit"]')[0]->hasAttribute('disabled'));
 
     // Test the connection with basic auth.
     $this->assertSendRequestMessage('.messages--status', 'Connection successful.');
-    $this->assertEmpty($this->cssSelect('details[data-drupal-selector="edit-debug"]'));
+    $web_assert->elementNotExists('css', 'details[data-drupal-selector="edit-debug"]');
+    // Make sure the token file is removed when switching to basic auth.
+    $token_file_path = \Drupal::service('file_system')->realpath(OauthTokenFileStorage::DEFAULT_DIRECTORY . '/oauth.dat');
```
mxr576: Use `$this->container `instead of` \Drupal::service()`.

mxr576 (changes requested):

src/Form/AuthenticationForm.php
```diff
@@ -456,77 +501,82 @@ protected function createSuggestion(\Exception $exception, KeyInterface $key): M
...
+    // Get any status messages so they can be rendered in the placeholder.
+    $status_messages = $this->renderer->render(StatusMessages::renderMessages());
+    // Clear any existing messages from the initial page load.
+    $response->addCommand(new ReplaceCommand('div.messages', ''));
+    $response->addCommand(new ReplaceCommand('#apigee-edge-auth-form-messages', "<div id=\"apigee-edge-auth-form-messages\">{$status_messages}</div>"));
+
+    // Replace the debug element (an empty wrapper if validation passes).
+    $response->addCommand(new ReplaceCommand(
+      '#apigee-edge-auth-form-debug-info',
+      $form['debug']['#access'] ? $this->renderer->render($form['debug']) : '<div id="apigee-edge-auth-form-debug-info" data-drupal-selector="edit-debug-placeholder"></div>'
```
mxr576, line 610: Could not you use the renderer service to render the original, empty form element if access is FALSE instead of returning this "custom" HTML markup?

src/Form/AuthenticationForm.php
```diff
@@ -456,77 +501,82 @@ protected function createSuggestion(\Exception $exception, KeyInterface $key): M
...
    */
-  public function ajaxCallback(array $form, FormStateInterface $form_state): array {
-    return $form;
+  public function ajaxCallback(array &$form, FormStateInterface $form_state): AjaxResponse {
+    $response = new AjaxResponse();
+    // Get any status messages so they can be rendered in the placeholder.
+    $status_messages = $this->renderer->render(StatusMessages::renderMessages());
+    // Clear any existing messages from the initial page load.
+    $response->addCommand(new ReplaceCommand('div.messages', ''));
+    $response->addCommand(new ReplaceCommand('#apigee-edge-auth-form-messages', "<div id=\"apigee-edge-auth-form-messages\">{$status_messages}</div>"));
```
mxr576: Could not you use the renderer service to render the form element with messages as its value instead of this inline template?

src/Form/AuthenticationForm.php
Jaesin: With `#access` set to FALSE, the renderer returns an empty string.

src/Form/AuthenticationForm.php
```diff
@@ -565,10 +565,10 @@ protected function createDebugText(\Exception $exception, KeyInterface $key): st
   public function ajaxCallback(array &$form, FormStateInterface $form_state): AjaxResponse {
     $response = new AjaxResponse();
     // Get any status messages so they can be rendered in the placeholder.
-    $status_messages = $this->renderer->render(StatusMessages::renderMessages());
+    $form['messages']['#value'] = $this->renderer->render(StatusMessages::renderMessages());
```
mxr576: Still... Only variables should be passed by reference. 
https://github.com/drupal/core/blob/17557c97a816bc78e70989a86064c562bc27027f/lib/Drupal/Core/Render/RendererInterface.php#L335

src/Form/AuthenticationForm.php
```diff
@@ -180,6 +192,32 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+        '#type' => 'html_tag',
+        '#tag' => 'label',
+        '#value' => $this->t('Unable to initialize configuration.'),
+      ];
+      $docs_link = Link::fromTextAndUrl('file_private_path', Url::fromUri('https://www.drupal.org/docs/8/core/modules/file/overview'))->toRenderable();
+      $docs_link['#attributes']['target'] = '_blank';
+      $form['connection_settings']['unconfigurable']['description'] = [
+        '#type' => 'html_tag',
+        '#tag' => 'div',
+        '#value' => $this->t("The Apigee Edge connection settings are not configured and we weren't able to automatically provision connection settings. Make sure the @docs_link is configured and try again.", [
```
mxr576: Just use the `$this->t('<a href=":url" target="_blank">title</a>', [':url' => 'http://example.com'])` pattern to add a link.

src/Form/AuthenticationForm.php
mxr576: Hm, indeed, that is unfortunate. There must be a better way to render the element instead of fake its HTML output, but let's leave it as it is for now.

src/OauthTokenFileStorage.php
Jaesin: @cnovak Should I work on this one or leave it for a follow up task?

cnovak: Let's add this to the GitHub issue queue and tag it as a known issue

mxr576 (approved):

mxr576:
@Jaesin Before you would merge this please double check this test because it has failed on my Travis:
https://travis-ci.org/mxr576/apigee-devportal-drupal/jobs/488391963#L3813

```
1) Drupal\Tests\apigee_edge\FunctionalJavascript\Form\AuthenticationFormJsTest::testAuthenticationForm
Behat\Mink\Exception\ExpectationException: The field "Apigee Edge endpoint" value is "https://api.enterprise.apigee.com/v1", but "" expected.
```

cnovak:
CLAs validated

Issue #110, opened 2019-01-16
[ERROR] Malformed inline YAML string:


rakeshjames:
[ERROR] Malformed inline YAML string: [ '@apigee_edge.sdk_connector',          
         '@apigee_edge.controller.organization',                                
         '@apigee_edge_teams.controller.cache.team',                            
         '@apigee_edge_teams.controller.cache.team_ids',                        
         '@apigee_edge.entity.controller.cache.general_app_cache_by_owner_factor
         y', '@apigee_edge.entity.controller.cache.app_name_by_owner_factory',  
         '@apigee_edge_teams.cache.company_membership_object',                  
         '@apigee_edge.controller.cache.developer_companies', at line 25 (near "
          '@apigee_edge.controller.cache.developer_companies',")

rakeshjames:
Please review https://github.com/apigee/apigee-edge-drupal/pull/116

mxr576:
#117 Fixes this, see details in #116. Thanks again @rakeshjames for reporting this!

PR #116, opened 2019-01-20
[ERROR] Malformed inline YAML string: - fixed

rakeshjames:
When I try the dev teams branch I got this error https://github.com/apigee/apigee-edge-drupal/issues/110, Please review the PR I have fixed that, yml error


mxr576:
@rakeshjames Thanks! It is weird that non of our automated tests has found this issue, It seems the same issue occurred with the XML sitemap module: https://www.drupal.org/project/xmlsitemap/issues/3007075
> I still don't understand why the tests aren't catching this, but this is easy for me to fix, so I'll get it fixed.


It is an important fix so I am adding it now as an individual commit. Please sign the CLA, so next time we could merge your PR as-is.

PR #129, opened 2019-02-02
DRUP-573: API Docs module base functionality

cnovak:
* Admin users can create, update, read, delete API Doc entities
* README explains how to setup Swagger UI
* View added to display API documentation to end users
* Added functional and Kernel test


mxr576:
@cnovak What is the priority of this PR? Do we want this in the next beta?
I would like to release a new version after #103 and #126 gets merged, which can happen in the next couple of hours.

mxr576 (changes requested):
~Parent module already has a license.txt, it is not necessary here.~ RESOLVED

modules/apigee_edge_apidocs/config/optional/views.view.apigee_edge_api_docs.yml
```diff
@@ -0,0 +1,291 @@
+langcode: en
+status: true
+dependencies:
+  config:
+  - system.menu.main
+  module:
+  - apigee_edge_apidocs
+  - text
+_core:
+  default_config_hash: CoPeJVeHBDWBWMm0Idt5gN0BMZD6dytv0SM1iT2_msQ
```
mxr576: Config hashes must not be exported.

modules/apigee_edge_apidocs/apidoc.page.inc
```diff
@@ -0,0 +1,30 @@
+<?php
```
mxr576: Filename should be [module_name].page.inc.

modules/apigee_edge_apidocs/apigee_edge_apidocs.links.task.yml
```diff
@@ -0,0 +1,22 @@
+# API Doc routing definition
+apidoc.settings_tab:
+  route_name: apidoc.settings
+  title: 'Settings'
+  base_route: apidoc.settings
```
mxr576, line 5: Do not want to display module's configuration under Configuration -> Apigee Edge? Or just not for now?
(I saw this module does not yet depend on the Apigee Edge module.)

modules/apigee_edge_apidocs/apigee_edge_apidocs.permissions.yml
```diff
@@ -0,0 +1,19 @@
+add apidoc entities:
+  title: 'Create new API Docs.'
```
mxr576: Unnecessary dot in the end of the title.

modules/apigee_edge_apidocs/apigee_edge_apidocs.permissions.yml
```diff
@@ -0,0 +1,19 @@
+add apidoc entities:
```
mxr576, line 1: Apigee Edge module depends on the Entity module - I know this module does not yet depend on the Apigee Edge module - but the Entity module provides handful utility classes that can autogenerate all these entity permissions.
`\Drupal\entity\EntityPermissionProviderBase`

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
```diff
@@ -0,0 +1,65 @@
...
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Access\AccessResult;
+
+/**
+ * Access controller for the API Doc entity.
+ *
+ * @see \Drupal\apigee_edge_apidocs\Entity\ApiDoc.
+ */
+class ApiDocAccessControlHandler extends EntityAccessControlHandler {
```
mxr576, line 33: Entity module also has an access handler for this `\Drupal\entity\EntityAccessControlHandler` which has been built to work with permissions provided by `\Drupal\entity\EntityPermissionProviderBase`. If you define your own access handler it would be great if you access checker would be covered by a test case.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
```diff
@@ -0,0 +1,65 @@
...
+
+    // Unknown operation, no opinion.
+    return AccessResult::neutral();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function checkCreateAccess(AccountInterface $account, array $context, $entity_bundle = NULL) {
+    return AccessResult::allowedIfHasPermission($account, 'add apidoc entities');
```
mxr576: Should not users with "administer apidoc entities" permission has access to this operation?

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
```diff
@@ -0,0 +1,65 @@
...
+    /** @var \Drupal\apigee_edge_apidocs\Entity\ApiDocInterface $entity */
+    switch ($operation) {
+      case 'view':
+        if (!$entity->isPublished()) {
+          return AccessResult::allowedIfHasPermission($account, 'view unpublished apidoc entities');
+        }
+        return AccessResult::allowedIfHasPermission($account, 'view published apidoc entities');
+
+      case 'update':
+        return AccessResult::allowedIfHasPermission($account, 'edit apidoc entities');
```
mxr576: Missing `cachePerUser()` calls.

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
```diff
@@ -0,0 +1,77 @@
...
+    $collection = parent::getRoutes($entity_type);
+
+    $entity_type_id = $entity_type->id();
+
+    if ($settings_form_route = $this->getSettingsFormRoute($entity_type)) {
+      $collection->add("$entity_type_id.settings", $settings_form_route);
+    }
+
+    $apidoc_collection_route = $collection->get('entity.apidoc.collection');
+    $apidoc_collection_route->setDefault('_title', 'API Docs');
```
mxr576: Don't we need the "Aliasing" feature of Apigee Edge entities on this entity type?

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
```diff
@@ -0,0 +1,77 @@
...
+  public function getRoutes(EntityTypeInterface $entity_type) {
+    $collection = parent::getRoutes($entity_type);
+
+    $entity_type_id = $entity_type->id();
+
+    if ($settings_form_route = $this->getSettingsFormRoute($entity_type)) {
+      $collection->add("$entity_type_id.settings", $settings_form_route);
+    }
+
+    $apidoc_collection_route = $collection->get('entity.apidoc.collection');
```
mxr576: The route may not be defined, wrap it in a condition.

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
```diff
@@ -0,0 +1,77 @@
...
+   * @return \Symfony\Component\Routing\Route|null
+   *   The generated route, if available.
+   */
+  protected function getSettingsFormRoute(EntityTypeInterface $entity_type) {
+    if (!$entity_type->getBundleEntityType()) {
+      $route = new Route("/admin/structure/{$entity_type->id()}/settings");
+      $route
+        ->setDefaults([
+          '_form' => 'Drupal\apigee_edge_apidocs\Form\ApiDocSettingsForm',
+          '_title' => "{$entity_type->getLabel()} settings",
```
mxr576: This is not going to be translateable, use a title callback.
https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/src/Entity/DeveloperAppRouteProvider.php#L218

modules/apigee_edge_apidocs/src/ApiDocListBuilder.php
```diff
@@ -0,0 +1,111 @@
...
+   * {@inheritdoc}
+   *
+   * We override ::render() so that we can add our own content above the table.
+   * parent::render() is where EntityListBuilder creates the table using our
+   * buildHeader() and buildRow() implementations.
+   */
+  public function render() {
+    $build['description'] = [
+      '#markup' => $this->t('Manage your API documentation. You can manage the fields on the <a href="@adminlink">API Docs settings page</a>.', [
+        '@adminlink' => $this->urlGenerator->generateFromRoute('apidoc.settings'),
```
mxr576: Simply use `\Drupal\Core\Url::fromRoute()`.

mxr576: Also you can see the preferred URL syntax here: https://github.com/drupal/core/blob/17557c97a816bc78e70989a86064c562bc27027f/modules/locale/locale.batch.inc#L102

modules/apigee_edge_apidocs/src/ApiDocTranslationHandler.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+namespace Drupal\apigee_edge_apidocs;
+
+use Drupal\content_translation\ContentTranslationHandler;
+
+/**
+ * Defines the translation handler for apidoc.
+ */
+class ApiDocTranslationHandler extends ContentTranslationHandler {
```
mxr576: Unnecessary child class, refer to the parent class where it is needed.

(Generated by Drupal Console? ;P )

modules/apigee_edge_apidocs/src/Entity/ApiDocViewsData.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+namespace Drupal\apigee_edge_apidocs\Entity;
+
+use Drupal\views\EntityViewsData;
+
+/**
+ * Provides Views data for API Doc entities.
+ */
+class ApiDocViewsData extends EntityViewsData {
```
mxr576: Unnecessary child class, refer to the parent class where it is needed.

modules/apigee_edge_apidocs/src/Form/ApiDocForm.php
```diff
@@ -0,0 +1,106 @@
...
+      $container->get('entity_type.bundle.info'),
+      $container->get('datetime.time'),
+      $container->get('messenger')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
```
mxr576: Unnecessary method override.

modules/apigee_edge_apidocs/src/Form/ApiDocForm.php
```diff
@@ -0,0 +1,106 @@
...
+
+    switch ($status) {
+      case SAVED_NEW:
+        $this->messenger->addMessage($this->t('Created the %label API Doc.', [
+          '%label' => $entity->label(),
+        ]));
+        break;
+
+      default:
+        $this->messenger->addMessage($this->t('Saved the $label API Doc.', [
```
mxr576: Use` $entity->getEntityType()->getSingularLabel()` instead of "API doc". Second parameter of `t()` can be extracted to a shared variable.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Form submission handler.
+   *
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
```
mxr576, line 81: Empty method?

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
```diff
@@ -0,0 +1,71 @@
...
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   *
+   * @return array
+   *   Form definition array.
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $form['apidoc_settings']['#markup'] = 'Settings form for API Docs. Manage field settings here.';
```
mxr576: Missing t().

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Defines the settings form for API Doc entities.
+   *
+   * @param array $form
+   *   An associative array containing the structure of the form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The current state of the form.
+   *
+   * @return array
```
mxr576: Missing inheritdoc.

modules/apigee_edge_apidocs/tests/src/Functional/ApiDocsAdminTest.php
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * Simple test to ensure that main page loads with module enabled.
+ *
+ * @group apigee_edge
+ * @group apigee_edge_apidocs
+ */
+class ApiDocsAdminTest extends BrowserTestBase {
+
+  use TestFileCreationTrait;
+
```
mxr576, line 36: Unnecessary new lines.

modules/apigee_edge_apidocs/templates/apidoc.html.twig
```diff
@@ -0,0 +1,22 @@
...
+ * Available variables:
+ * - content: A list of content items. Use 'content' to print all content, or
+ * - attributes: HTML attributes for the container element.
+ *
+ * @see template_preprocess_apidoc()
+ *
+ * @ingroup themeable
+ */
+#}
+<div{{ attributes.addClass('apidoc') }}>
```
mxr576: Not sure whether the custom template is necessary for this entity. If Drupal Console generated it I'd get rid of this until it is actually not needed.

modules/apigee_edge_apidocs/tests/src/Kernel/ApiDocTest.php
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Basic CRUD operations on a ApiDoc entity.
+   */
+  public function testEntity() {
+    $this->installEntitySchema('apidoc');
+    $entity = ApiDoc::create([
+      'name' => 'API 1',
+      'description' => 'Test API 1',
+      'spec' => NULL,
```
mxr576, line 44: Why an API Documentation can be created without a specification?

mxr576 (changes requested):

mxr576:
Missing JIRA number from the title of the PR.

cnovak:
@mxr576 this is going into 8.x-1.x-apidocs feature branch and will not be merged until we add more functionality, so no, it does not need to be merged now.

cnovak:
CLAs verified

modules/apigee_edge_apidocs/apigee_edge_apidocs.links.task.yml
cnovak: I moved it under Apigee Edge, good idea.  It does not yet depend on it, but being a submodule that seems to make more sense.

modules/apigee_edge_apidocs/apigee_edge_apidocs.permissions.yml
cnovak: I am not sure what needs to be done at this point, it this more of a TODO for later functionality integration?  We will want to reference API products in the future but do not have that functionality yet.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
cnovak: Admin role can create API Doc entities, am I missing something?  I do need to add tests however like you mentioned

cnovak: Ignore that comment above, I think I understand your reasoning now.  Changning access control handler to give 'administer apidoc entities' permissions to do CRUD

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
cnovak: Not sure what you mean, can you clarify?  This is supposed to be a simple add of the base functionality.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
cnovak: Added DRUP-585 for this

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
cnovak: Created DRUP-586 for this

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
cnovak: Function implements FormInterface function

modules/apigee_edge_apidocs/tests/src/Kernel/ApiDocTest.php
cnovak: An API Doc may not need a spec when setting up a doc that starts only with a summary when first configuring.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
cnovak: Added test cases for access

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
cnovak: added

modules/apigee_edge_apidocs/apigee_edge_apidocs.links.task.yml
mxr576: Did you move it? I do not see any reference to `system.admin_config_edge` as "parent" in menu.yml. 
e.g.: https://github.com/apigee/apigee-edge-drupal/blob/375850f6d8261dbb409cb0b388e55a6b91ac884f/apigee_edge.links.menu.yml#L16

modules/apigee_edge_apidocs/apigee_edge_apidocs.permissions.yml
mxr576: If you would like to leave it as-it-is for now, that is fine, but please create a new user story about considering replacing this with Entity module's EntityPermissionProviderBase before beta/stable.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
```diff
@@ -0,0 +1,70 @@
...
+   */
+  protected function checkAccess(EntityInterface $entity, $operation, AccountInterface $account) {
+    /** @var \Drupal\apigee_edge_apidocs\Entity\ApiDocInterface $entity */
+    switch ($operation) {
+      case 'view':
+        if (!$entity->isPublished()) {
+          return AccessResult::allowedIfHasPermission($account, 'view unpublished apidoc entities')
+            ->orIf(AccessResult::allowedIfHasPermission($account, 'administer apidoc entities'));
+        }
+        return AccessResult::allowedIfHasPermission($account, 'view published apidoc entities')
```
mxr576: I'd suggest to use `allowedIfHasPermissions($account, ['foo', 'bar'], 'OR')` when multiple permissions needs to be checked.

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function getRoutes(EntityTypeInterface $entity_type) {
+    $collection = parent::getRoutes($entity_type);
+
+    $entity_type_id = $entity_type->id();
+
+    if ($settings_form_route = $this->getSettingsFormRoute($entity_type)) {
+      $collection->add("$entity_type_id.settings", $settings_form_route);
```
mxr576, line 45: Always use curly braces syntax if you can http://php.net/manual/en/language.types.string.php#language.types.string.parsing.complex, `"{$entity_type_id}.settings"`.

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
mxr576: Nitpicking, use the same semantic as for settings_form_route:

```php
if ($apidoc_collection_route = $collection->get('entity.apidoc.collection')) {
 ....
}
```

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
mxr576: TL;DR; I'd set this to "API Doc settings" and close DRUP-586.


OK, would it be just easier to use `API Doc settings` till 586 gets fixed? That string would be translatable although it would not respect if the entity type's label would change. Which I am quite sure is unnecessary, because admin UI page titles do not need to respect entity title alterations/aliases. They do not respect in case of the main module either, "Developer apps" are always "Developer apps" on the admin settings UI no matter how developer apps called for end-users (ex.: Apps).

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -0,0 +1,228 @@
...
+ *   id = "apidoc",
+ *   label = @Translation("API Doc"),
+ *   label_singular = @Translation("API Doc"),
+ *   label_plural = @Translation("API Docs"),
+ *   handlers = {
+ *     "view_builder" = "Drupal\Core\Entity\EntityViewBuilder",
+ *     "list_builder" = "Drupal\apigee_edge_apidocs\ApiDocListBuilder",
+ *     "views_data" = "Drupal\views\EntityViewsData",
+ *     "translation" = "Drupal\content_translation\ContentTranslationHandler",
+ *
```
mxr576, line 43: Unnecessary empty line.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -0,0 +1,228 @@
...
+ *   admin_permission = "administer apidoc entities",
+ *   entity_keys = {
+ *     "id" = "id",
+ *     "label" = "name",
+ *     "uuid" = "uuid",
+ *     "langcode" = "langcode",
+ *     "status" = "status",
+ *   },
+ *   links = {
+ *     "canonical" = "/apidoc/{apidoc}",
```
mxr576, line 67: Should not this be `/api-doc/{apidoc}` or even `/api/{apidoc}`. Either of these would look better in Google results.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -0,0 +1,228 @@
...
+   */
+  public function setDescription($description) {
+    $this->set('description', $description);
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCreatedTime() {
```
mxr576, line 113: Nice to have, add return types to our own methods, just like in the main module. (Create  JIRA issue about this.)

modules/apigee_edge_apidocs/src/Form/ApiDocForm.php
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * The messenger service.
+   *
+   * @var \Drupal\Core\Messenger\MessengerInterface
+   */
+  protected $messenger;
+
+  /**
+   * Constructs a new  API Doc edit form.
```
mxr576, line 46: `s/a new  /a/g`

modules/apigee_edge_apidocs/src/Form/ApiDocForm.php
```diff
@@ -0,0 +1,99 @@
...
+    switch ($status) {
+      case SAVED_NEW:
+        $this->messenger->addMessage($this->t('Created the %label @entity_type_label.', [
+          '%label' => $entity->label(),
+          '@entity_type_label' => $singular_label,
+        ]));
+        break;
+
+      default:
+        $this->messenger->addMessage($this->t('Saved the $label @entity_type_label.', [
```
mxr576, line 91: This is probably not the best default behavior because if $status is not `SAVED_NEW` or `SAVED_UPDATED` then it could be an indicator of an error, e.g.: `0`.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
mxr576: If this admin form does not contain any configuration at this moment it should not be committed yet or if you think that there will be configuration options _in the near future_ here a TODO should be left with a warning to finalize the implementation.

modules/apigee_edge_apidocs/tests/src/Functional/ApiDocsAccessTest.php
```diff
@@ -0,0 +1,418 @@
...
+
+    // Test the 'administer apidoc entities' permission.
+    $this->drupalLogin($this->drupalCreateUser([
+      'administer apidoc entities',
+      'administer apidoc display',
+      'administer apidoc fields',
+      'administer apidoc form display',
+    ]));
+
+    $this->drupalGet('apidoc/' . $this->apidocPublished->id());
```
mxr576, line 111: Always build paths from routes, even in tests.
Ex.: https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/tests/src/Functional/DeveloperAppUITest.php#L547

mxr576 (changes requested):

modules/apigee_edge_apidocs/tests/src/Kernel/ApiDocTest.php
mxr576: This seems to something that should be solved by using Workflow module. The Workflow states would specify whether a field is required or not. So you could create new API documentation with a "draft" state without attaching an API documentation file but you can not change its state to "published" without attaching the actual the actual API specification to the entity. 
We have built plenty of portals with an implementation like this to customers.

https://www.drupal.org/docs/8/core/modules/workflows/overview

modules/apigee_edge_apidocs/tests/src/Functional/ApiDocsAccessTest.php
```diff
@@ -0,0 +1,418 @@
...
+use Drupal\apigee_edge_apidocs\Entity\ApiDoc;
+use Drupal\Tests\BrowserTestBase;
+
+/**
+ * Tests the ApiDoc term access permissions.
+ *
+ * @group apigee_edge
+ * @group apigee_edge_apidocs
+ */
+class ApiDocsAccessTest extends BrowserTestBase {
```
mxr576, line 31: @peterserfozo
* Please describe why fewer test methods within a test class are better in Drupal 8 than more. 
* Please demonstrate to Chris how entity access can be tested in a better way in Drupal 8.

peterserfozo: In our case, it's better to use fewer test methods in a test class because Drupal is reinstalled before every test method execution in functional tests. Of course, creating well-separated test cases is a PHPUnit best practice but in Drupal 8 we should create fewer because of the heavy system setup. And if there are time-consuming operations (for example API calls) in `setUp()` then it's better to call it once because `setUp()` is called before executing test methods. On the other hand, our module is really complex, we use different static and persistent cache layers, persistent data, etc. It means if we create separated test cases then we may hide bugs because of the purged DB. Anyway, writing well-structured code is important, you can create a similar solution like this one: https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/tests/src/Functional/QueryTest.php#L126

The best approach to test permissions and access system is using a permission matrix: a two-dimensional array is enough to define which routes are accessible to the different permissions/roles then the test loops through this array, sets the roles of the user and validates whether the user should have access to a route or not in every step. To test entity access you can use the entity's route provider for collecting the routes defined by the entity. A good example: https://github.com/apigee/apigee-edge-drupal/blob/237ff6fbd5c83dd3b6efffafff82ff20f7dd26da/modules/apigee_edge_teams/tests/src/Functional/AccessTest.php

peterserfozo (changes requested):

modules/apigee_edge_apidocs/src/Form/ApiDocForm.php
```diff
@@ -0,0 +1,99 @@
...
+   * @param \Drupal\Core\Entity\EntityTypeBundleInfoInterface $entity_type_bundle_info
+   *   The entity type bundle service.
+   * @param \Drupal\Component\Datetime\TimeInterface $time
+   *   The time service.
+   * @param \Drupal\Core\Messenger\MessengerInterface $messenger
+   *   The messenger service.
+   */
+  public function __construct(EntityRepositoryInterface $entity_repository, EntityTypeBundleInfoInterface $entity_type_bundle_info, TimeInterface $time, MessengerInterface $messenger) {
+    parent::__construct($entity_repository, $entity_type_bundle_info, $time);
+    $this->messenger = $messenger;
```
peterserfozo, line 59: It's not necessary to inject the messenger service because the class `FormBase` uses the trait `MessengerTrait` so you can simply call `$this->messenger()->addMessage()` in your code.

peterserfozo (changes requested):

modules/apigee_edge_apidocs/config/optional/views.view.apigee_edge_api_docs.yml
```diff
@@ -0,0 +1,289 @@
...
+base_field: id
+core: 8.x
+display:
+  default:
+    display_plugin: default
+    id: default
+    display_title: Master
+    position: 0
+    display_options:
+      access:
```
dawehner, line 24: I would have expected that this view checks for the "view published apidoc entities" permission.

modules/apigee_edge_apidocs/apigee_edge_apidocs.links.menu.yml
```diff
@@ -0,0 +1,13 @@
...
+  title: 'API Docs'
+  route_name: entity.apidoc.collection
+  description: 'List API Docs'
+  parent: system.admin_structure
+  weight: 100
+
+apidoc.admin.structure.settings:
+  title: 'API Docs'
+  description: 'Configure API Doc settings'
+  route_name: apidoc.settings
```
dawehner: Are you sure this route name is right? I would have rather expected something like `apigee_edge_apidocs.settings` as name.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
dawehner: Given you have set an 'admin_permission' in the entity annotation you could call to the parent implementation first and then combine that result with ```->orIf```. This way you don't have to check for 'administer apidoc entities' all over the place.

modules/apigee_edge_apidocs/apigee_edge_apidocs.info.yml
```diff
@@ -0,0 +1,8 @@
+name: 'Apigee Edge: API Docs'
+type: module
+description: 'Display OpenAPI documentation of your APIs to your developers.'
+core: 8.x
+package: 'Apigee (Experimental)'
```
dawehner, line 5: Suggestion: You can provide a ```configure: apidocs.settings```key which adds a nice little configure icon on the admin -> extend page.

modules/apigee_edge_apidocs/src/Form/ApiDocForm.php
dawehner: Node module checks for ```    $insert = $node->isNew();``` instead to decide whether to show the created a saved message.

dawehner: See ```\Drupal\node\NodeForm::save```

modules/apigee_edge_apidocs/src/Form/ApiDocDeleteForm.php
```diff
@@ -0,0 +1,74 @@
...
+use Drupal\Core\Entity\ContentEntityConfirmFormBase;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Url;
+
+/**
+ * Provides a form for deleting API Doc entities.
+ *
+ * @ingroup apigee_edge_apidocs
+ */
+class ApiDocDeleteForm extends ContentEntityConfirmFormBase {
```
dawehner, line 32: I'm curious, did you consider using ```\Drupal\Core\Entity\ContentEntityDeleteForm``` instead?

modules/apigee_edge_apidocs/src/Entity/ApiDocInterface.php
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * Returns the API Doc published status indicator.
+   *
+   * Unpublished API Doc are only visible to restricted users.
+   *
+   * @return bool
+   *   TRUE if the API Doc is published.
+   */
+  public function isPublished();
```
dawehner, line 98: Suggestion: Instead of adding isPublished and setPublished here you could extend ```\Drupal\Core\Entity\EntityPublishedInterface``` which provides these methods already.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -0,0 +1,228 @@
...
+      ))
+      ->setDisplayOptions('form', array(
+        'label' => 'hidden',
+        'type' => 'file_generic',
+        'weight' => 0,
+      ))
+      ->setDisplayConfigurable('form', TRUE)
+      ->setDisplayConfigurable('view', TRUE);
+
+    $fields['status'] = BaseFieldDefinition::create('boolean')
```
dawehner, line 208: Note: If you use the ```EntityPublishedInterface```, see below, you can also leverage ```\Drupal\Core\Entity\EntityPublishedTrait``` to define this base field

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -0,0 +1,228 @@
...
+      ->setDisplayOptions('form', [
+        'type' => 'boolean_checkbox',
+        'weight' => 1,
+      ]);
+
+    $fields['created'] = BaseFieldDefinition::create('created')
+      ->setLabel(t('Created'))
+      ->setDescription(t('The time that the entity was created.'));
+
+    $fields['changed'] = BaseFieldDefinition::create('changed')
```
dawehner, line 221: Did you considered adding the ```\Drupal\Core\Entity\EntityChangedInterface``` interface?

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -0,0 +1,228 @@
...
+use Drupal\Core\Entity\EntityChangedTrait;
+use Drupal\Core\Entity\EntityTypeInterface;
+
+/**
+ * Defines the API Doc entity.
+ *
+ * @ingroup apigee_edge_apidocs
+ *
+ * @ContentEntityType(
+ *   id = "apidoc",
```
dawehner, line 34: I am wondering whether it would be sensible to prefix the entity type ID by apigee, so 'apigee_apidocs'. This could avoid some potential conflict, even I guess the risk would be quite low. Sadly this would require a lot of changes all over the place.

modules/apigee_edge_apidocs/src/ApiDocListBuilder.php
```diff
@@ -0,0 +1,87 @@
...
+    $build['table'] = parent::render();
+    return $build;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildHeader() {
+    $header['id'] = $this->t('ID');
+    $header['name'] = $this->t('Name');
```
dawehner, line 69: Would it be worth exposing the description as additional column?

modules/apigee_edge_apidocs/apigee_edge_apidocs.links.menu.yml
```diff
@@ -0,0 +1,13 @@
+# API Doc menu items definition
+entity.apidoc.collection:
+  title: 'API Docs'
+  route_name: entity.apidoc.collection
+  description: 'List API Docs'
+  parent: system.admin_structure
```
dawehner, line 6: Would it be sensible to move API docs to content? It feels for me more like content, but maybe that's just me

modules/apigee_edge_apidocs/apigee_edge_apidocs.links.task.yml
dawehner: I moved it over

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
dawehner: I fixed that. I don't understand how it is possible that this route doesn't exist, but ¯\_(ツ)_/¯ :)

modules/apigee_edge_apidocs/apigee_edge_apidocs.permissions.yml
dawehner: I created a task for that: https://apigeesc.atlassian.net/browse/DRUP-617

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
dawehner: I went ahead and solved DRUP-586 as part of this PR.

modules/apigee_edge_apidocs/src/ApiDocListBuilder.php
dawehner: I created an extra ticket for that: https://apigeesc.atlassian.net/browse/DRUP-618

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
dawehner: I created another task for that: https://apigeesc.atlassian.net/browse/DRUP-619

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
dawehner: Resolved

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
dawehner: I created another task for that: https://apigeesc.atlassian.net/browse/DRUP-620

modules/apigee_edge_apidocs/src/Entity/ApiDocInterface.php
dawehner: I realized this interface isn't available yet in Drupal 8.6. I was looking at Drupal 8.7-dev locally.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
dawehner: I realized this interface isn't available yet in Drupal 8.6. I was looking at Drupal 8.7-dev locally.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
dawehner: Nevermind the interface is used, but the base fields don't exist on the trait yet

modules/apigee_edge_apidocs/src/Form/ApiDocForm.php
dawehner: Good point!

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
dawehner: I added a quick todo.

modules/apigee_edge_apidocs/tests/src/Kernel/ApiDocTest.php
dawehner: It seems like this isn't really something we can solve right now. If a field is required, you cannot save an entity, but as described by @cnovak there is a valid usecase for it. Given that I mark this as resolved. I think it could be up to the individual site owners to implement custom workflows if they need some.

PR #137, opened 2019-02-15
DRUP-581: Private file path issues

cnovak:
Display if private file path needs to be configured, or if it is a permissions problem on the file path. Tests created.

### DIFF
--- a/src/Form/AuthenticationForm.php
+++ b/src/Form/AuthenticationForm.php
@@ -216,10 +216,10 @@ class AuthenticationForm extends ConfigFormBase {
 
     if (!$private_path_is_writable || !$private_path_is_directory) {
       if (!$private_path_is_directory) {
-        $error = t('The private file path %path does not exist', ['%path' => $file_private_path]);
+        $error = t('The private file path %path does not exist.', ['%path' => $file_private_path]);
       }
       else {
-        $error = t('The private file path %path is not writable', ['%path' => $file_private_path]);
+        $error = t('The private file path %path is not writable.', ['%path' => $file_private_path]);
       }
       $this->disableForm($form, $error);
       return $form;
--- a/src/Form/AuthenticationForm.php
+++ b/src/Form/AuthenticationForm.php
@@ -37,6 +37,8 @@ use Drupal\Core\Form\ConfigFormBase;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Render\Element\StatusMessages;
 use Drupal\Core\Render\RendererInterface;
+use Drupal\Core\Site\Settings;
+use Drupal\Core\StreamWrapper\PrivateStream;
 use Drupal\key\Entity\Key;
 use Drupal\key\KeyInterface;
 use Drupal\key\KeyRepositoryInterface;
@@ -107,6 +109,13 @@ class AuthenticationForm extends ConfigFormBase {
   protected $fileSystem;
 
   /**
+   * Site settings.
+   *
+   * @var \Drupal\Core\Site\Settings
+   */
+  protected $settings;
+
+  /**
    * Constructs a new AuthenticationForm.
    *
    * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
@@ -123,8 +132,10 @@ class AuthenticationForm extends ConfigFormBase {
    *   The `renderer` service.
    * @param \Drupal\Core\File\FileSystemInterface $file_system
    *   The `file_system` service.
+   * @param \Drupal\Core\Site\Settings $settings
+   *   The `settings` service.
    */
-  public function __construct(ConfigFactoryInterface $config_factory, KeyRepositoryInterface $key_repository, SDKConnectorInterface $sdk_connector, ModuleHandlerInterface $module_handler, OauthTokenStorageInterface $oauth_token_storage, RendererInterface $renderer, FileSystemInterface $file_system) {
+  public function __construct(ConfigFactoryInterface $config_factory, KeyRepositoryInterface $key_repository, SDKConnectorInterface $sdk_connector, ModuleHandlerInterface $module_handler, OauthTokenStorageInterface $oauth_token_storage, RendererInterface $renderer, FileSystemInterface $file_system, Settings $settings) {
     parent::__construct($config_factory);
     $this->keyRepository = $key_repository;
     $this->sdkConnector = $sdk_connector;
@@ -132,6 +143,7 @@ class AuthenticationForm extends ConfigFormBase {
     $this->oauthTokenStorage = $oauth_token_storage;
     $this->renderer = $renderer;
     $this->fileSystem = $file_system;
+    $this->settings = $settings;
   }
 
   /**
@@ -145,7 +157,8 @@ class AuthenticationForm extends ConfigFormBase {
       $container->get('module_handler'),
       $container->get('apigee_edge.authentication.oauth_token_storage'),
       $container->get('renderer'),
-      $container->get('file_system')
+      $container->get('file_system'),
+      $container->get('settings')
     );
   }
 
@@ -190,26 +203,31 @@ class AuthenticationForm extends ConfigFormBase {
       '#tree' => TRUE,
     ];
 
+    // Validate private file path is set.
+    if (!PrivateStream::basePath()) {
+      $this->disableForm($form, $this->t('The Drupal private file setting has not been configured.'));
+      return $form;
+    }
+
+    // Validate private file path is a directory and is writable.
+    $file_private_path = Settings::get('file_private_path');
+    $private_path_is_writable = is_writable($file_private_path);
+    $private_path_is_directory = is_dir($file_private_path);
+
+    if (!$private_path_is_writable || !$private_path_is_directory) {
+      if (!$private_path_is_directory) {
+        $error = t('The private file path %path does not exist', ['%path' => $file_private_path]);
+      }
+      else {
+        $error = t('The private file path %path is not writable', ['%path' => $file_private_path]);
+      }
+      $this->disableForm($form, $error);
+      return $form;
+    }
+
     // Save the key for later use.
     if (!($this->activeKey = $active_key = $this->getActiveKey())) {
-      $form['actions']['#disabled'] = TRUE;
-      $form['connection_settings']['unconfigurable'] = [
-        '#type' => 'container',
-        '#attributes' => ['class' => ['form-item']],
-      ];
-      $form['connection_settings']['unconfigurable']['label'] = [
-        '#type' => 'html_tag',
-        '#tag' => 'label',
-        '#value' => $this->t('Unable to initialize configuration.'),
-      ];
-      $form['connection_settings']['unconfigurable']['description'] = [
-        '#type' => 'html_tag',
-        '#tag' => 'div',
-        '#value' => $this->t("The Apigee Edge connection settings are not configured and we weren't able to automatically provision connection settings. Make sure the <a href=\":file_docs_uri\" target=\"_blank\">file_private_path</a> is configured and try again.", [
-          ':file_docs_uri' => 'https://www.drupal.org/docs/8/core/modules/file/overview',
-        ]),
-      ];
-
+      $this->disableForm($form, $this->t('Unable to initialize configuration.'));
       return $form;
     }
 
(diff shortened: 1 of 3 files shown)


src/Form/AuthenticationForm.php
```diff
@@ -190,26 +203,31 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
     ];
 
+    // Validate private file path is set.
+    if (!PrivateStream::basePath()) {
+      $this->disableForm($form, $this->t('The Drupal private file setting has not been configured.'));
+      return $form;
+    }
+
+    // Validate private file path is a directory and is writable.
+    $file_private_path = Settings::get('file_private_path');
```
mxr576: @cnovak Probably the easiest way to validate whether the Private FS is configured and it is configured _properly_ is what we do in the OauthTokenFileStorage. The `private://` stream wrapper is only registered when file_private_path is configured and the `realpath()` can only resolve the path if the file_private_path is correct. Not sure about whether the `is_writable()` will be still necessary you should try that out. (We also tried this Settings based approach before in the OauthTokenFileStorage and it was not stable enough.) 
https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/src/OauthTokenFileStorage.php#L280

mxr576: Core does perform an `$is_writable `check, but it also tries to create directories first, so maybe it only does it because of that.
https://github.com/drupal/core/blob/8bd590f7aec78b8d273187ec2d1adab262e2e1d7/modules/system/system.install#L670

src/Form/AuthenticationForm.php
```diff
@@ -415,6 +433,37 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
+   *
+   * If there are issues with the private file path, we want to show an error
+   * to the user instead of the form.
+   *
+   * @param array $form
+   *   The form to disable.
+   * @param string $error
+   *   The error to display.
+   */
+  protected function disableForm(array &$form, string $error) {
```
mxr576: The second parameter should be a `TranslatableMarkup` object because this is what `$this->t()` returns, you should not cast that to a string here.

src/Form/AuthenticationForm.php
```diff
@@ -106,6 +108,13 @@ class AuthenticationForm extends ConfigFormBase {
    */
   protected $fileSystem;
```
mxr576: `$fileSystem` has become an unused variable unless you change your implementation as I suggested below.

tests/src/Kernel/Form/AuthenticationFormTest.php
```diff
@@ -103,6 +97,65 @@ public function testGenerateNewAuthKey() {
...
+    $this->assertEquals('The Drupal private file setting has not been configured.', $form['connection_settings']['unconfigurable']['label']['#value']);
+
+    // Add file_private_path setting, but not the private file dir.
+    $private_directory = DrupalKernel::findSitePath(Request::create('/')) . '/private';
+    $this->setSetting('file_private_path', $private_directory);
+
+    // Rebuild the form.
+    $form_state = new FormState();
+    $form = \Drupal::formBuilder()->buildForm(AuthenticationForm::class, $form_state);
+    $this->assertInstanceOf(AuthenticationForm::class, $form_state->getFormObject());
```
mxr576: Why this check is necessary? We know for sure what the form builder returned, aren't we?

tests/src/Kernel/Form/AuthenticationFormTest.php
```diff
@@ -103,6 +97,65 @@ public function testGenerateNewAuthKey() {
...
+
+    // Private file system is not configured.
+    $this->assertNotEmpty($form['connection_settings']['unconfigurable']['description']);
+    $this->assertTrue($form['actions']['#disabled']);
+    $this->assertEquals('The Drupal private file setting has not been configured.', $form['connection_settings']['unconfigurable']['label']['#value']);
+
+    // Add file_private_path setting, but not the private file dir.
+    $private_directory = DrupalKernel::findSitePath(Request::create('/')) . '/private';
+    $this->setSetting('file_private_path', $private_directory);
```
mxr576: Could not we use a path in the virtual fs for testing this instead of actually creating dirs and files on the hard drive?
See: `\Drupal\Tests\apigee_edge\Kernel\OauthTokenFileStorageTest` and `\Drupal\KernelTests\KernelTestBase::setUpFilesystem()` and `\Drupal\KernelTests\Core\File\FileTestBase::setUpFilesystem()`.

tests/src/Kernel/Form/AuthenticationFormTest.php
```diff
@@ -103,6 +97,65 @@ public function testGenerateNewAuthKey() {
...
+    $form = \Drupal::formBuilder()->buildForm(AuthenticationForm::class, $form_state);
+    $this->assertInstanceOf(AuthenticationForm::class, $form_state->getFormObject());
+
+    // Make sure error is shown on page, and form is not available.
+    $this->assertNotEmpty($form['connection_settings']['unconfigurable']['description']);
+    $this->assertTrue($form['actions']['#disabled']);
+    $this->assertStringEndsWith('is not writable.', $form['connection_settings']['unconfigurable']['label']['#value']);
+
+    // Make private file path writable.
+    \Drupal::service('file_system')->chmod($private_directory, 755);
```
mxr576: Please use `$this->container->get()` wherever it is possible instead of `\Drupal::`, especially in kernel tests.

mxr576 (changes requested):

mxr576:
@cnovak Please add a link to the description with the latest Travis build related to this PR that confirms all tests pass with a real Apigee Edge connection with your changes.

src/Form/AuthenticationForm.php
cnovak: This was the way the original code was written, but @llynch00 and I both saw a few issues with that approach: 1) The error message was vague and would say to configure the private file path even when it was set in settings.php and 2) still came back null after making a change to the settings.php file and/or creating the directory.  Seemed to need a cache refresh for the error to go away.  #1 problem was also why we are checking if the path is writable. 

I updated to use `PrivateStream::basePath()` just like `system_install()` does.

src/Form/AuthenticationForm.php
cnovak: Good point, updated code and tests to use `TranslatableMarkup`

src/Form/AuthenticationForm.php
cnovak: removed

tests/src/Kernel/Form/AuthenticationFormTest.php
cnovak: This is from the earlier commited code, did not seem like a bad idea. Removed

cnovak:
@mxr576 here are the tests, not all tests pass due to other errors: https://travis-ci.org/cnovak/apigee-edge-drupal/builds/493595603

tests/src/Kernel/Form/AuthenticationFormTest.php
cnovak: Nice, thanks for info on virtual fs, updated.

tests/src/Kernel/Form/AuthenticationFormTest.php
cnovak: fixed

cnovak:
All items noted above should be fixed now

cnovak:
One more commit to fix code style issues, now done for real

src/Form/AuthenticationForm.php
mxr576: > still came back null after making a change to the settings.php file and/or creating the directory. Seemed to need a cache refresh for the error to go away.

Yes, you should always do a cache clear if you modify its content. So this is part of the workflow.

src/Form/AuthenticationForm.php
```diff
@@ -190,26 +192,31 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       '#tree' => TRUE,
     ];
 
+    // Validate private file path is set.
+    $file_private_path = PrivateStream::basePath();
+    if (!$file_private_path) {
+      $this->disableForm($form, $this->t('The Drupal private file setting has not been configured.'));
```
mxr576, line 198: * I do not think "Drupal" is necessary here.

src/Form/AuthenticationForm.php
```diff
@@ -190,26 +192,31 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+      return $form;
+    }
+
+    // Validate private file path is a directory and is writable.
+    $private_path_is_writable = is_writable($file_private_path);
+    $private_path_is_directory = is_dir($file_private_path);
+
+    if (!$private_path_is_writable || !$private_path_is_directory) {
+      if (!$private_path_is_directory) {
+        $error = t('The private file path %path does not exist.', ['%path' => $file_private_path]);
```
mxr576, line 208: Use `$this->t()`.

tests/src/Kernel/Form/AuthenticationFormTest.php
```diff
@@ -103,6 +106,60 @@ public function testGenerateNewAuthKey() {
...
 
+  /**
+   * Test private file path errors.
+   *
+   * @throws \Drupal\Core\Form\EnforcedResponseException
+   * @throws \Drupal\Core\Form\FormAjaxException
+   */
+  public function testPrivateFileSystemPathErrors() {
+    $form_state = new FormState();
+    $form = \Drupal::formBuilder()->buildForm(AuthenticationForm::class, $form_state);
```
mxr576, line 117: Use `$this->container->get('form_builder')`.

tests/src/Kernel/Form/AuthenticationFormTest.php
```diff
@@ -103,6 +106,60 @@ public function testGenerateNewAuthKey() {
...
+    $this->assertInstanceOf(AuthenticationForm::class, $form_state->getFormObject());
```
mxr576, line 118: This assertion still does not provide any value.

tests/src/Kernel/Form/AuthenticationFormTest.php
```diff
@@ -103,6 +106,60 @@ public function testGenerateNewAuthKey() {
...
+    $this->assertDirectoryExists($this->privateFileDirectoryUri());
+
+    // Rebuild the form.
+    $form_state = new FormState();
+    $form = \Drupal::formBuilder()->buildForm(AuthenticationForm::class, $form_state);
+
+    // Make sure error is shown on page, and form is not available.
+    $this->assertNotEmpty($form['connection_settings']['unconfigurable']['description']);
+    $this->assertTrue($form['actions']['#disabled']);
+    $this->assertEquals(new TranslatableMarkup('The private file path %path is not writable.', ['%path' => $this->privateFileDirectoryUri()]), $form['connection_settings']['unconfigurable']['label']['#value']);
```
mxr576, line 149: You do not need to use a TranslatableMarkup in the assertion, comparison happens between plain English strings in tests.

mxr576:
@cnovak Peter Serföző (@peterserfozo) highlighted an important problem when we had a chat about this PR. This sanity check for private FS should only be performed in this page IF
* there is no key provider
* OR the selected key provider is the private filesystem
* OR the authentication method is Oauth - because the OAuth token is stored in the private FS.

If either the environment variable or Lockr or any other key storages are being used this sanity check should not be executed on this form. 

Just think about the following workflow:
* Somone installs the site by using config_installer.
* They active key gets imported by the config_installer and it uses the environment key storage, the authentication method is "HTTP basic".
* This form should not warn the user that the private FS is not configured properly because it is not being used.

Using the private file storage is highly not recommended in a production environment still so building a robust logic around the expectation that is being used may not be the best solution.

Also, maybe all this validation that we do in the authentication form should be decoupled and performed by the private key provider plugin and the authentication form should call the "requirements" method on the key provider to decide whether the system meets with the requirements of the plugin. If not it should raise the warning that it does now.

UPDATE: `\Drupal\apigee_edge\Plugin\KeyProvider\PrivateFileKeyProvider::validateConfigurationForm()` already tries to validate whether the private FS is configured properly or not. Either this method or a decoupled method should be called by the authentication form to validate the requirements. 
The initial idea with this authentication form was to provide a simplified UI for configuring the Apigee Edge credentials but still leverage as much as possible from what the Key module and its plugins ca provide us. Meaning:
* The key module (plugin) should provide the configuration form for the key.
* The key module (plugin) should validate the input data.
* The key module (plugin) should save the key configuration.

mxr576:
Closed in favor of #152.

PR #138, opened 2019-02-15
DRUP-406 Team roles and new team permissions feature

mxr576:
Latest Travis build: https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/499202151


modules/apigee_edge_teams/js/apigee_edge_teams.team_permissions.js
```diff
@@ -0,0 +1,63 @@
+/*
```
Jaesin, line 1: It would be nice to have some comments in this file.

modules/apigee_edge_teams/apigee_edge_teams.module
```diff
@@ -158,13 +182,22 @@ function apigee_edge_teams_api_product_access(EntityInterface $entity, $operatio
...
+  /** @var \Drupal\apigee_edge_teams\TeamMemberApiProductAccessHandlerInterface $access_checker */
+  $access_checker = \Drupal::service('apigee_edge_teams.team_member_api_product_access_handler');
   /** @var \Drupal\apigee_edge_teams\TeamMembershipManagerInterface $team_membership_manager */
   $team_membership_manager = \Drupal::service('apigee_edge_teams.team_membership_manager');
-  $team_ids = $team_membership_manager->getTeams($account->getEmail());
+  try {
+    $developer_team_ids = $team_membership_manager->getTeams($account->getEmail());
+  }
+  catch (Exception $e) {
+    $developer_team_ids = [];
```
Jaesin: I'm wondering why we are ignoring `DeveloperDoesNotExistException` here.

We could return early with something like `return AccessResult::neutral("{$account->getEmail()} was not found.");`

modules/apigee_edge_teams/apigee_edge_teams.module
```diff
@@ -179,9 +212,9 @@ function apigee_edge_teams_api_product_access(EntityInterface $entity, $operatio
     /** @var \Drupal\apigee_edge_teams\Entity\Storage\TeamStorageInterface $team_storage */
     $team_storage = \Drupal::entityTypeManager()->getStorage('team');
     /** @var \Drupal\apigee_edge_teams\Entity\TeamInterface $team */
-    $teams = $team_storage->loadMultiple($team_ids);
+    $teams = $team_storage->loadMultiple($developer_team_ids);
```
Jaesin, line 216: `Team::loadMultiple()` could be a bit nicer here. The same with `Developer::load()`.

modules/apigee_edge_teams/apigee_edge_teams.services.yml
```diff
@@ -55,15 +49,21 @@ services:
...
+    arguments: ['@apigee_edge_teams.team_membership_manager', '@apigee_edge_teams.team_permissions', '@module_handler', '@current_user']
 
-  apigee_edge_teams.event_subscriber.team_api_product_access_manager_cache_reset:
-    class: Drupal\apigee_edge_teams\EventSubscriber\TeamApiProductAccessManagerCacheReset
-    arguments: ['@apigee_edge_teams.team_api_product_access_manager']
+  apigee_edge_teams.event_subscriber.team_member_api_product_access_handler_cache_reset:
+    class: Drupal\apigee_edge_teams\EventSubscriber\TeamMemberApiProductAccessHandlerCacheReset
+    arguments: ['@apigee_edge_teams.team_member_api_product_access_handler']
     tags:
       - { name: event_subscriber }
```
Jaesin, line 69: Why not add a general purpose team permission checker like: 

https://github.com/Jaesin/apigee-m10n-drupal/commit/2858cbf1f69618fbe9a1909cbf62d1fdd5ec83d1

modules/apigee_edge_teams/src/Controller/TeamMembersList.php
```diff
@@ -45,17 +48,21 @@ class TeamMembersList extends ControllerBase {
    *
    * @param \Drupal\apigee_edge_teams\TeamMembershipManagerInterface $team_membership_manager
    *   The team membership manager service.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
    */
-  public function __construct(TeamMembershipManagerInterface $team_membership_manager) {
+  public function __construct(TeamMembershipManagerInterface $team_membership_manager, EntityTypeManagerInterface $entity_type_manager) {
     $this->teamMembershipManager = $team_membership_manager;
+    $this->entityTypeManager = $entity_type_manager;
```
Jaesin, line 56: This file is now using a mix of `$this->entityTypeManager` and `$this->entityTypeManager()`.

modules/apigee_edge_teams/src/Controller/TeamMembersList.php
```diff
@@ -181,9 +217,12 @@ protected function buildOperations(string $member, TeamInterface $team): array {
   protected function getOperations(string $member, TeamInterface $team) {
     $operations = [];
     $operations['edit'] = [
+      'title' => $this->t('Edit'),
```
Jaesin, line 220: "Manage Roles" might be better here.

modules/apigee_edge_teams/src/Entity/Form/TeamAppCreateFormBase.php
```diff
@@ -24,7 +24,7 @@
 use Drupal\apigee_edge\Entity\Controller\AppCredentialControllerInterface;
```
Jaesin, line 24: Usually entity create forms are just controllers that instantiate the entity and return the edit form.
See: `\Drupal\node\Controller\NodeController::add()`,

modules/apigee_edge_teams/src/TeamPermissionHandler.php
```diff
@@ -0,0 +1,300 @@
...
+ *   # DynamicTeamPermissionProviderInterface interface.
+ *   - Drupal\apigee_edge_teams\DefaultTeamPermissionsProvider
+ * @endcode
+ *
+ * Based on Drupal core's PermissionHandler with a bunch of improvements.
+ *
+ * @see \Drupal\user\PermissionHandler
+ * @see \Drupal\apigee_edge_teams\DynamicTeamPermissionProviderInterface
+ */
+final class TeamPermissionHandler implements TeamPermissionHandlerInterface {
```
Jaesin, line 72: I wonder if this differs from a plugin manager so much that it can't extend `\Drupal\Component\Plugin\PluginManagerBase`.

modules/apigee_edge_teams/js/apigee_edge_teams.team_permissions.js
mxr576: I think it is quite straightforward what it does, especially because I mentioned it is a modified version of a well-known Drupal file.

modules/apigee_edge_teams/apigee_edge_teams.module
mxr576: I do not like static method calls on entity classes. You should not use that in classes and it only takes one more line to eliminate static method calls procedural functions as well.
I think it is a bad design in Drupal 8 that an entity object - that should be only a data object - can perform CRUDL operations itself and it also knows and has a dependency to its entity storage handlers.

Related problem in Symfony: https://www.tomasvotruba.cz/blog/2017/10/16/how-to-use-repository-with-doctrine-as-service-in-symfony/

modules/apigee_edge_teams/src/Controller/TeamMembersList.php
mxr576: We may like to allow later to modify additional information related to a team member on this form, this is the reason why I used this general label.

modules/apigee_edge_teams/src/Controller/TeamMembersList.php
mxr576: https://github.com/apigee/apigee-edge-drupal/pull/139/commits/c1e5a4c54e276a1390b624fdb6a73a51332160d8 commit in #139 fixes this.

modules/apigee_edge_teams/src/Entity/Form/TeamAppCreateFormBase.php
mxr576: Yeah, but node add and node edit form always look like the same by default, always. Team/developer app add and edit form looks completely differently, moreover and app form looks differently on different routes.

(@Jaesin Please resolve this comment if I manager to answer your question.)

modules/apigee_edge_teams/src/TeamPermissionHandler.php
mxr576: This could be plugin manager, but it could be an overkill for this, just like the user permission handler in Drupal 8 core, this is more lightweight and task-oriented but still allows to solve every _valid_ use cases that you may want to implement in a custom module. (Maybe it is even faster than if it would be a plugin manager.)

Let me know if you see any use case that you can not implement with this approach.

modules/apigee_edge_teams/apigee_edge_teams.services.yml
mxr576: I thought about that but we did not need that yet in the core module. Feel free to contribute it back to the core module.
PS.: I left a comment in your commit.

modules/apigee_edge_teams/apigee_edge_teams.module
mxr576: Good catch! TBH, I have just run a quick check to ensure `->getTeams()` is in trying to catch but haven't run a check to verify whether what it is in the `catch` is the optimal solution every time.

modules/apigee_edge_teams/src/Form/AddTeamMembersForm.php
```diff
@@ -73,7 +82,8 @@ class AddTeamMembersForm extends FormBase {
   public function __construct(TeamMembershipManagerInterface $team_membership_manager, EntityTypeManagerInterface $entity_type_manager) {
     $this->teamMembershipManager = $team_membership_manager;
     $this->userStorage = $entity_type_manager->getStorage('user');
-    $this->teamEntityType = $entity_type_manager->getDefinition('team');
+    $this->teamRoleStorage = $entity_type_manager->getStorage('team_role');
```
Jaesin, line 85: There is a problem with loading entity handlers during a service constructor.

The this loads entity type definitions and if there are any services used during `hook_entity_type_alter`, you can get a circular dependency error.

This effectively only happens on cache rebuild but it will prevent the container from being rebuilt.

mxr576: This is a common pattern in Drupal core as well, therefore, I believe that there is no problem with retrieving saving entity storage handlers from the entity type manager in the constructor.
So you either found [a Drupal core bug](https://www.drupal.org/project/issues/drupal?text=circular+reference+dependency&status=Open&priorities=All&categories=All&version=All&component=All) or you are somehow misusing the combinations of services and `hook_entity_type_alter()` hook implementations.
I'll check a [MINT module](https://github.com/apigee/apigee-m10n-drupal), let's move this conversation to another channel.

Jaesin: I've seen this in plugins but never in a service constructor.

src/ParamConverter/DeveloperWithUser.php
```diff
@@ -0,0 +1,74 @@
...
+
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\ParamConverter\ParamConverterInterface;
+use Symfony\Component\Routing\Route;
+
+/**
+ * Resolves "developer_with_user" type parameters in routes.
+ *
+ * Use this parameter converter instead of paramconverter.entity if you would
+ * like to ensure that a developer (by UUID or email address) from Apigee Edge
```
cnovak, line 31: We should not be using UUIDs or supporting them, should only use email address in code and comments

src/ParamConverter/DeveloperWithUser.php
```diff
@@ -0,0 +1,74 @@
...
+  private $developerStorage;
+
+  /**
+   * DeveloperWithUser constructor.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
+   */
+  public function __construct(EntityTypeManagerInterface $entity_type_manager) {
+    $this->developerStorage = $entity_type_manager->getStorage('developer');
```
Jaesin: This causes all of the functional tests in the `apigee_m10n` module to fail with the following error:

```
Drupal\Component\Plugin\Exception\PluginNotFoundException : The "developer" entity type does not exist.

.../core/lib/Drupal/Core/Entity/EntityTypeManager.php:133
.../core/lib/Drupal/Core/Entity/EntityTypeManager.php:234
.../core/lib/Drupal/Core/Entity/EntityTypeManager.php:170
.../modules/contrib/apigee_edge/src/ParamConverter/DeveloperWithUser.php:50
.../core/lib/Drupal/Component/DependencyInjection/Container.php:262
```

Are all of the tests in the `apigee_edge` module still passing with this addition?

src/ParamConverter/DeveloperWithUser.php
mxr576: @cnovak I thought we have closed this topic with the two last comments in DRUP-489. We have to support them, but we can enforce usage of email address where it is possible.

src/ParamConverter/DeveloperWithUser.php
mxr576: No, they do not, but Peter has fixed this problem already in his machine and it will be pushed today. I did not expect that you are using a draft branch in M10N module.
Fun fact, the fix is simple. the `$entity_type_manager->getStorage()` can not be used in param converter constructors because entity type manager service collects the storages later than param converters are being built. This may work if I would convert the param converter to a lazy service, not sure.

modules/apigee_edge_teams/src/Form/AddTeamMembersForm.php
mxr576: This is a form constructor, this is not the first place where this "pattern" is being used. So if it compiles, not like the param converter plugins (see below) I think there is no problem with that. The `hook_entity_type_alter()` issue that you have described is an edge case that should be handled in place. Multiple projects suffers from circular reference errors, sometimes only using `\Drupal::service()` instead of DI solves the problem as I could see:
https://www.google.hu/search?q=drupal+circular+reference+detected+for+service

Jaesin: Yes, there shouldn't be any issue with this form.  The service that is causing us to use a workaround is `\Drupal\apigee_edge_teams\TeamPermissionHandler::__construct()`.

modules/apigee_edge_teams/src/TeamPermissionHandler.php
```diff
@@ -0,0 +1,300 @@
...
+   * @param \Drupal\apigee_edge_teams\TeamMembershipManagerInterface $team_membership_manager
+   *   The team membership manager service.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
+   */
+  public function __construct(ModuleHandlerInterface $module_handler, ClassResolverInterface $class_resolver, TeamMembershipManagerInterface $team_membership_manager, EntityTypeManagerInterface $entity_type_manager) {
+    $this->moduleHandler = $module_handler;
+    $this->classResolver = $class_resolver;
+    $this->teamMembershipManager = $team_membership_manager;
+    $this->teamRoleStorage = $entity_type_manager->getStorage('team_role');
```
Jaesin, line 138: I marked the wrong file for this issue but see: https://github.com/apigee/apigee-edge-drupal/pull/138#discussion_r260523610

peterserfozo (approved):

mxr576:
Double checked tests that failed on Travis on my local dev env and except the known issues (ex.: authentication form) all tests passed so I am merging this PR.

PR #144, opened 2019-03-03
Inject product access manager via setter injection

dawehner:
Alternative PR to https://github.com/apigee/apigee-edge-drupal/pull/141


mxr576:
@dawehner I tried to use setter injection before here to solve this but it did not solve all issues. Also, this PR does not cover #140 issue completely, where @Jaesin reported an issue related to the team permission handler service primarily.

dawehner:
No worries :)

PR #146, opened 2019-03-06
[DRUP-565] Add a team context switcher

shadcn:
This PR adds a team context switcher block and a form. This is wip for DRUP-565.


modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -0,0 +1,123 @@
...
+  public function getFormId() {
+    return 'team_context_switcher_form';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state, array $teams = []) {
+    /** @var \Drupal\apigee_edge_teams\Entity\TeamInterface $team */
+    $options = ['' => $this->t('Select a team')];
```
mxr576: Apigee Edge entities can be aliased, retrieve the "team" string's current alias from the team entity type.
Ex.: https://github.com/apigee/apigee-edge-drupal/blob/7eff57bc7ac2a9ea777a54826a509ea61aeb952f/modules/apigee_edge_teams/src/Entity/Form/TeamDeleteForm.php#L56

mxr576: Use #empty_option instead of `['' => $this->t('Select a team')]`.
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Select.php/class/Select/8.2.x

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -0,0 +1,123 @@
...
+      '#type' => 'container',
+      '#attributes' => [
+        'class' => [
+          'container-inline',
+        ],
+      ],
+    ];
+
+    $form['wrapper']['context'] = [
+      '#title' => $this->t('Select a team'),
```
mxr576: The same comment as above regarding "team".

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -0,0 +1,123 @@
...
+    ];
+
+    $form['wrapper']['actions'] = [
+      '#type' => 'actions',
+    ];
+
+    $form['wrapper']['actions']['submit'] = [
+      '#type' => 'submit',
+      '#button_type' => 'primary',
+      '#value' => $this->t('Go'),
```
mxr576: Should not it be "Switch" or even better, configurable on the block level?

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -0,0 +1,123 @@
...
+    return new static(
+      $container->get('current_route_match')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getFormId() {
+    return 'team_context_switcher_form';
```
mxr576: Use module name as a prefix in form ids.

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -0,0 +1,123 @@
...
+    ];
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    if ($context = $form_state->getValue('context')) {
```
mxr576: If the form element is required, why do you need to check whether there is a form value? Ideally, you should not be able to get here if you submitted the form with the empty value.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,145 @@
...
+   * @param \Drupal\Core\Entity\EntityStorageInterface $storage
+   *   The team entity storage.
+   * @param \Drupal\Core\Form\FormBuilderInterface $form_builder
+   *   The form builder.
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, AccountInterface $current_user, TeamMembershipManagerInterface $team_membership_manager, EntityStorageInterface $storage, FormBuilderInterface $form_builder) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->currentUser = $current_user;
+    $this->teamMembershipManager = $team_membership_manager;
+    $this->storage = $storage;
```
mxr576: Call it entity storage just like everywhere else.

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -0,0 +1,123 @@
...
+    return 'team_context_switcher_form';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state, array $teams = []) {
+    /** @var \Drupal\apigee_edge_teams\Entity\TeamInterface $team */
+    $options = ['' => $this->t('Select a team')];
+    foreach ($teams as $team) {
```
mxr576: Options should only contain those teams that a user is actually an active member OR if it has "Manage teams" site-wide permission it should contain all.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,145 @@
...
+  protected function blockAccess(AccountInterface $account) {
+    return AccessResult::allowedIf($account->isAuthenticated());
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    // If the user can administer team, show all teams.
+    if ($this->currentUser->hasPermission('administer team')) {
```
mxr576: No access check should happen here. The form builder should check the available teams for a user and display only those as an option. (See my comment above).

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,145 @@
...
+use Drupal\Core\Session\AccountInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a block for switching team context.
+ *
+ * @Block(
+ *   id = "apigee_edge_team_switcher",
+ *   admin_label = @Translation("Select a team"),
+ *   category = @Translation("Apigee")
```
mxr576: Apigee Edge?

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,145 @@
...
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Session\AccountInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a block for switching team context.
+ *
+ * @Block(
+ *   id = "apigee_edge_team_switcher",
+ *   admin_label = @Translation("Select a team"),
```
mxr576: Team switcher? Go to team?

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,145 @@
...
+    // If the user can administer team, show all teams.
+    if ($this->currentUser->hasPermission('administer team')) {
+      $teams = NULL;
+    }
+    // Otherwise show teams the user belongs to.
+    elseif (!($teams = $this->teamMembershipManager->getTeams($this->currentUser->getEmail()))) {
+      return NULL;
+    }
+
+    return $this->formBuilder->getForm(TeamContextSwitcherForm::class, $this->storage->loadMultiple($teams));
```
mxr576: What should be displayed to a user if it does not have access to any team? The form with the empty "Select a team" select list?

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
```
mxr576, line 223: I would also override the `getCacheTags()` method and add the "team_list" tag to the cache tags to ensure this block gets invalidated when a developer's team membership changes.
See `\Drupal\apigee_edge_teams\Controller\TeamMembersList::overview()`

mxr576 (changes requested):

mxr576:
@arshad Please do not forget to add a test that ensures only those Teams are visible in the block that the current user has access.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,145 @@
...
+use Drupal\Core\Form\FormBuilderInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Session\AccountInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a block for switching team context.
+ *
+ * @Block(
+ *   id = "apigee_edge_team_switcher",
```
mxr576: apigee_edge_teams_team_switcher (Module name as a prefix)

mxr576 (changes requested):

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
shadcn: Done. I didn't know about this one. Thanks. 👍

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
shadcn: Done.

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
shadcn: Done.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
shadcn: Done.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
shadcn: No block is displayed if the user does not have access to any team.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
shadcn: Done.

shadcn:
> @arshad Please do not forget to add a test that ensures only those Teams are visible in the block that the current user has access.

Yeah. I'm going to add this next. Wanted to make sure about the implementation first.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
shadcn: Renamed to Team switcher.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
shadcn: hmm so when a user has permission to "administer team", the user can still view all teams on the `/teams` overview page. Wouldn't it be inconsistent to be able to view all teams but not have them in the switcher?

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
shadcn: the `team_list` tag?

```
public function getCacheTags() {
    return Cache::mergeTags(parent::getCacheTags(), ['team_list']);
  }
```

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
Jaesin: I would rather have this as a jump menu or a drop button. Using links would let the routing system handle access.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
Jaesin: That could end up being a long list. Should there be a limit on the number of teams in the list?

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state, array $teams = []) {
+    $form_state->set('teams', $teams);
+
+    /** @var \Drupal\apigee_edge_teams\Entity\TeamInterface $team */
+    $options = [];
+    foreach ($teams as $id => $team) {
+      $options[$id] = $team->label();
```
mxr576: Things are just getting more completeced here :-)
What is no question that list of visible Teams must be limited here and not elsewhere. This is missing at this moment.

Normally, if we would like to implement a "Go to team" feature which does not keep the original "context" when it redirect the user I would suggest to call `$team->access('view')` to display only those teams that the current user has access. (Keeping the context means that it redirects a user from Team A's create team app form to Team B's create team app instead of Team B's canonical page.)
But if we would like to keep the context here and implement a "Switch team page" functionality that keeps the context, this from should display only those teams as an available option that the **route access handler service** returns access granted. Based on the example above, it could happen that I have access to view both Team A's, Team B's and Team C's canonical pages and their team app list, but if I am on the Team A's create team app page, I could only see Team C as an available "Go to" option because I do not have "Create team apps" permission in Team B. Although, when I am on the team app list page of Team B, I could see Team A and Team C as an option here.

What do you think @Jaesin @cnovak ?

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    // Get the current team from the route to use as default value.
+    /** @var \Drupal\apigee_edge_teams\Entity\TeamInterface $current_team */
+    $current_team = $this->routeMatch->getParameter('team') ?? NULL;
+
+    $title = (string) $this->t('Select @name', [
```
mxr576: Do not cast a translatable markup to string, leave it translateable markup object.

mxr576: Use `@team` instead of `@name` to make translators life easier.

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
mxr576: This list also should not display the current team (from the route) as an option because that does not make any sense.

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -0,0 +1,155 @@
...
+
+    return $form;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    /** @var \Drupal\apigee_edge_teams\Entity\TeamInterface $team */
+    if (($team_id = $form_state->getValue('team_id'))
```
mxr576: The indention of this if condition looks inappropriate, probably caused all these conditions checks...

Why *all* these sanity checks are needed? A form in Drupal cannot be submitted with a value that was not displayed there originally so `(isset($teams[$team_id]) && ($team = $teams[$team_id])` is unnnessary and form field is required therefore `$team_id = $form_state->getValue('team_id'))` is also redundant.

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -0,0 +1,155 @@
...
+      && ($teams = $form_state->get('teams'))
+      && (isset($teams[$team_id]))
+      && ($team = $teams[$team_id])) {
+      // Default to the canonical url.
+      $url = $team->toUrl();
+
+      // If there is team parameter in route, redirect to corresponding route.
```
mxr576: Otherwise, redirect the user to the selected team's canonical page.

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -0,0 +1,155 @@
...
+   */
+  public function getFormId() {
+    return 'apigee_edge_teams_context_switcher_form';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state, array $teams = []) {
+    $form_state->set('teams', $teams);
```
mxr576: Do not pass teams from the block. Collect teams in the form. (Somone may like to display this form elsewhere as well).

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
mxr576: @arshad See my comments above regarding access checks, what is sure, that you should not check permission in the block level. All users should see this block but with different options which are handled in the form level.

@Jaesin  I do not think that a user would be a member of more than 10 groups (teams) in general which still could be displayed properly within a select list. If a customer has any problem with the default select list, s/he can install Chosen.js (to make the list searchable) or replace the form element in a form alter hook.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
mxr576: Yes, this is the default team entity listing render cache tag that gets automatically cleared when a developer's membership state changes (becomes a member of a team, or gets removed from a team.)

mxr576 (changes requested):

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
shadcn: Yep. This is already in place.

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
shadcn: @mxr576 do teams have granular access control? How can I give permission to a user to create app in Team A and not Team B?

@mxr576 Also, I'd argue we need to keep the current team as an option. It gives the user context from navigating. Maybe the current team could be disabled in the select but it should be there.

mxr576: * How do you disable an option in a select list? :) I mean, sure, you can display the current team somehow in the form, but not in the select list, otherwise the user could select it, click on "Go" which simply just reloads the page - maybe s/he is going to lose some form data because of that if s/he does that on a team related form. But this is just a small details compared with the who should see what question.
* The possibility is there, but by default all teams (all team members) share the same settings (permissions), meaning that, if a team member can create apps then it can create team apps in all teams where s/he is a member. Check [apigee_edge_teams.api.php](https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/modules/apigee_edge_teams/apigee_edge_teams.api.php#L63) and team member permission related classes. Even a user who is not member of a team can get access to team related UIs via site-wide permissions or custom implements of `hook_apigee_edge_teams_developer_permissions_by_team_alter()` alter hook.

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -79,7 +103,13 @@ public function getFormId() {
   /**
    * {@inheritdoc}
    */
-  public function buildForm(array $form, FormStateInterface $form_state, array $teams = []) {
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    // Get the teams for the user.
+    if (!($team_ids = $this->teamMembershipManager->getTeams($this->currentUser->getEmail()))) {
+      return NULL;
```
mxr576: Form builder should always return a (render) array.

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
```diff
@@ -134,10 +164,10 @@ public function buildForm(array $form, FormStateInterface $form_state, array $te
    */
   public function submitForm(array &$form, FormStateInterface $form_state) {
     /** @var \Drupal\apigee_edge_teams\Entity\TeamInterface $team */
-    if (($team_id = $form_state->getValue('team_id'))
-      && ($teams = $form_state->get('teams'))
-      && (isset($teams[$team_id]))
-      && ($team = $teams[$team_id])) {
+    if ($teams = $form_state->get('teams')) {
```
mxr576: Still these sanity checks are unnecessary. The submitted team id always exists and valid and form state storage always contains the `teams` because you passed that from the form builder.

mxr576 (changes requested):

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
shadcn: What should we display if the user has no team?

mxr576: This was a question that I also asked yesterday, I think we should wait for @cnovak 's answer here. I think the empty option list with only "Select team" should be a sane default as the "Teams" link is also always available in the user menu.

Jaesin:
I talked with Arun and @cnovak about this today. They want this to be more like what was in the  d7 dev portal. 
image

This looks very similar to a drop button which only takes two clicks to switch (opposed to 3 with a form). I think making these links will simplify things and can be handled in the block.

The equivalent of the "Manage Companies" link at the bottom of the list would be a link to `/teams`.

Even if you are a team administrator, the switcher should only show the teams in which you are a member. This is how it works in D7 and they don't see the need to change that.

The current context shows at the top only.

I hope this helps.

mxr576:
TL;DR; I am fine with this simplification.

1. Just to confirm. Does it mean that the switcher would not keep the "context" and it would always redirect the user to the front page (canonical page) of the team? If the answer is yes, we only need to check whether the user has "view" access to a team entity to figure out which teams should we display in this form.
2. It would be great to find a built-in render element in Drupal 8 core that works somehow similar to what you have showcased from Drupal 7. Custom render elements in the Drupal 7 version always complicated a front-ender's life. Therefore we tried to use only Drupal 8 core render elements wherever it was possible because that way we do not need to worry about how they look with different themes. Even the worst custom themes usually provide a quite OK looking overrides to core render elements.
Maybe dropbutton could be used for this? https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Dropbutton.php/class/Dropbutton/8.6.x
3. _"Even if you are a team administrator, the switcher should only show the teams in which you are a member. This is how it works in D7 and they don't see the need to change that."_ This simplification or limitation of the switcher must be documented. I'd also display it on the block configuration form to make sure everybody understands how this work.

mxr576:
Maybe dropbutton could be used for this?
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Render%21Element%21Dropbutton.php/class/Dropbutton/8.6.x

mxr576:
Also, auto-submit can be enabled on a simple select list as well - just like as it would be a Views exposed filter - so I would suggest to implement a simple and stable solution in the module and make it fancy in the Kickstart theme. Because the look and feel that @Jaesin shared from Drupal 7 is typically something that should not be implemented in a module.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,187 @@
...
+
+      $build['#links'] = [
+        'current' => [
+          'title' => $current_route_entity->label(),
+          'attributes' => ['class' => ['apigee-team-context-current']],
+        ],
+      ];
+      if ($current_route_entity_type === 'team') {
+        $build['#links']['developers'] = [
+          'title' => t('Developers:'),
```
mxr576: Use `StringTranslationTrait` and `$this->t()`.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,187 @@
...
+          'title' => $this->current_user->getDisplayName(),
+          'url' => Url::fromRoute($user_route, ['user' => $this->current_user->id()]),
+        ];
+      }
+      $build['#links']['teams'] = [
+        'title' => t('Teams:'),
+        'attributes' => ['class' => ['apigee-team-context-group-label']],
+      ];
+
+      $teams = $this->team_membership_manager->getTeams($this->current_user->getEmail());
```
mxr576: `getTeams()` returns an array of **team IDs (names)** and not display names, therefore, it should not be used as title below.
Also, it would be easier to load Team entities from Apigee Edge and call `access()` on entity objects, for examplel `$this->teamStorage->loadMultiple($this->team_membership_manager->getTeams($this->current_user->getEmail()))`.

modules/apigee_edge_teams/src/Routing/TeamAppByNameRouteAlterSubscriber.php
```diff
@@ -51,6 +52,14 @@ protected function alterRoutes(RouteCollection $collection) {
         NestedArray::setValue($params, ['app', 'type'], 'team_app_by_name');
         $route->setOption('parameters', $params);
       }
+      // Add the team route ID to any developer (user) routes as corresponding
+      // to a team route. We add the team route to the developer route to take
+      // advantage of route caching.
+      if (($developer_route_id = $route->getOption(TeamContextManagerInterface::DEVELOPER_ROUTE_OPTION_NAME))
```
mxr576: 1. Can you describe in details why it is necessary?
2. Even if it is necessary it should not be implemented in a **TeamApp** related class.
3. It is hard to read this condition because of how it is indented. Maybe you should not assign values to variables in conditions.

modules/apigee_edge_teams/src/TeamContextManager.php
```diff
@@ -0,0 +1,62 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+namespace Drupal\apigee_edge_teams;
+
+use Drupal\Core\Routing\RouteMatchInterface;
+
+/**
+ * Service that makes easier to work with company (team) memberships.
```
mxr576: This is what TeamMembershipManager does :)
Please describe the actual problem that you were tried to solve with this service in details.

modules/apigee_edge_teams/src/TeamContextManager.php
```diff
@@ -0,0 +1,62 @@
...
+    $this->current_route_match = $current_route_match;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function correspondingRouteName() {
+    // Get the current route object.
+    if ($current_route_object = $this->current_route_match->getRouteObject()) {
+      // The corresponding team, developer route or null.
```
mxr576: What does it do with the developer? Based on its name it should not do anything with the developer in the route. If it would only handle team related routes then tagging routes with custom options would not be necessary. It could figure out the context from the `{team}` in the route object.
The `DEVELOPER_ROUTE_OPTION_NAME` name is also confusing because I see above that you assigned `entity.user.canonical` to this route option.

modules/apigee_edge_teams/src/TeamContextManagerInterface.php
```diff
@@ -0,0 +1,42 @@
...
+  const DEVELOPER_ROUTE_OPTION_NAME = '_apigee_developer_route';
+  const TEAM_ROUTE_OPTION_NAME = '_apigee_team_route';
+
+  /**
+   * Gets the name for the route that corresponds to the current route.
+   *
+   * @return null|string
+   *   The corresponding route name if one is detected.
+   */
+  public function correspondingRouteName();
```
mxr576: Please use PHP 7 syntax in files/methods that related to the Apigee Edge module and not inherited from Drupal core.

modules/apigee_edge_teams/src/TeamContextManager.php
```diff
@@ -0,0 +1,62 @@
...
+ * It also handles cache invalidation.
+ */
+class TeamContextManager implements TeamContextManagerInterface {
+
+  /**
+   * The current route match.
+   *
+   * @var \Drupal\Core\Routing\RouteMatchInterface
+   */
+  protected $current_route_match;
```
mxr576: lowerCamelCasePropertyName

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,187 @@
...
+   * @var \Drupal\Core\Session\AccountInterface
+   */
+  protected $current_user;
+
+  /**
+   * The team context manager.
+   *
+   * @var \Drupal\apigee_edge_teams\TeamContextManagerInterface
+   */
+  protected $team_context_manager;
```
mxr576: lowerCamelCasePropertyNames

mxr576 (changes requested):

mxr576:
@Jaesin Please also attach a GIF animation about the swither changes that you implemented. Now it displays the developer as well?

Jaesin:
Sorry, that rough draft was accidentally merged into Arshad's branch.

modules/apigee_edge_teams/css/apigee_edge_teams.switcher.css
```diff
@@ -0,0 +1,45 @@
...
+.team-switcher .dropbutton .dropbutton-action span {
+  display: block;
+}
+
+.team-switcher .dropbutton-wrapper .dropbutton-widget {
+  text-align: left;
+}
+
+/* Style the dropbutton-no-action as a dropbutton divider. */
+.team-switcher .dropbutton-action > span.dropbutton-no-action {
```
mxr576: > I would suggest to implement a simple and stable solution in the module and make it fancy in the Kickstart theme. Because the look and feel that @Jaesin shared from Drupal 7 is typically something that should not be implemented in a module.
https://github.com/apigee/apigee-edge-drupal/pull/146#issuecomment-470956425

> Therefore we tried to use only Drupal 8 core render elements wherever it was possible because that way we do not need to worry about how they look with different themes.
https://github.com/apigee/apigee-edge-drupal/pull/146#issuecomment-470836413

If you start adding custom styles to Drupal core form elements, you should start worrying about again how these tweaks look like in different themes. Can't we avoid this in the module and use the dropdown render element as it looks like by default? (Make it fancy in the Kickstart theme as I suggested earlier.)

@Jaesin @cnovak

mxr576:
@arshad Let me know when this PR is ready for review again. (I see code style issues already, it would be great if you could keep conditions and chained method calls in one line until PHPCS does not complain about the length.)

modules/apigee_edge_teams/css/apigee_edge_teams.switcher.css
Jaesin: I don't see any issue with this CSS since it's prefixed with a class and wouldn't affect anything except the switcher. I would prefer to have the component provider present the component in a sane way. If someone installs the module without the kickstart, the feature should be usable as it was installer. It shouldn't require a custom theme to make it usable.

mxr576: Are you saying that dropdown render element without CSS is not usable? Do you have a screenshot about how it looks like now?
>  usable as it was installer

Usable where? On desktop? On mobile? Because as far as I know, core renders elements work fine on mobile devices, for example, tables, buttons, navigation... Thanks for this My apps just looks fine on mobile devices in Drupal 8. (Compared with how it did not work in Drupal thanks to the custom UI.)

shadcn: @mxr576 the first element from the dropbutton is a non-link. So we added minimum styling to remove the hover effect from the first element. That's it.

modules/apigee_edge_teams/src/Form/TeamContextSwitcherForm.php
shadcn: Closing this since this is now outdated. We are now using a dropbutton instead of a form with a select.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,273 @@
...
+   * Returns an array of additional links for the switcher.
+   *
+   * @return array
+   *   An array of additional links keyed with the route name.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  protected function getAdditionalLinks() {
+    return [
+      'entity.team.collection' => $this->t('Manage @teams', [
```
Jaesin: Since this will be seen by any authenticated user (not just users with the `administer team` permission), I think this should just say "Teams" like the main menu link to the same URL. Possibly "My teams".

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,273 @@
...
+   */
+  public function build() {
+    // Do not show a block if we do not have a context.
+    if (!($current_context = $this->getCurrentContext())) {
+      return [];
+    }
+
+    // Add a link for the developer account.
+    $entities = [
+      $this->entityTypeManager->getStorage('user')->load($this->account->id()),
```
Jaesin, line 145: Does this entity really need to be loaded? The `$this->account` will contain the user email?

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,273 @@
...
+    ];
+
+    // Add links for teams.
+    if ($team_ids = $this->teamMembershipManager->getTeams($this->account->getEmail())) {
+      $entities = array_merge($entities, $this->entityTypeManager->getStorage('team')
```
Jaesin, line 150: Do these teams really need to be loaded? The team membership manager will return their IDs. Loading these means that any route with team context will load all teams on every page. Loading would hopefully come from cache but if don't need any additional information from those entities, we should avoid loading them.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,273 @@
...
+  protected function getCurrentContext() {
+    // Get the current context from the team or the user in the route.
+    if (($current_context = $this->routeMatch->getParameter('team'))
+      || ($current_context = $this->routeMatch->getParameter('user'))
+    ) {
+      return $current_context;
+    }
+
+    // Get the current context from additional links.
+    $additional_links = $this->getAdditionalLinks();
```
Jaesin: I don't think we should care about current context of additional links. The only additional link planned is for `/teams` and it is a listing page that shows all the  links to teams (same as the switcher) making the switcher redundant. There is no `developers` listing for corresponding context so the switcher doesn't provide any additional value on this page.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,273 @@
...
+        $links[] = [
+          'title' => $title,
+          'url' => $url,
+          'is_current' => (is_string($current_context) && $current_context === $route_name),
+        ];
+      }
+    }
+
+    // Move the current link to the top and render it as a non link.
+    foreach ($links as $index => $link) {
```
Jaesin: Re-processing this list shouldn't really be necessary. We can start the list with the current context and then skip adding it if it is the same as the current route. #optimization.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,273 @@
...
+   * Helper to find the current context from the route.
+   *
+   * @return mixed
+   *   The current context from the route or the route name or null.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  protected function getCurrentContext() {
+    // Get the current context from the team or the user in the route.
+    if (($current_context = $this->routeMatch->getParameter('team'))
```
Jaesin: `routeMatch->getParameter()` Does not ensure the return type.

modules/apigee_edge_teams/src/TeamContextManager.php
```diff
@@ -0,0 +1,86 @@
...
+  public function __construct(RouteMatchInterface $route_match) {
+    $this->routeMatch = $route_match;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getDestinationUrlForEntity(EntityInterface $entity): ?Url {
+    if ($corresponding_route_name = $this->getCorrespondingRouteNameForEntity($entity)) {
+      $parameters = $this->routeMatch->getParameters();
```
Jaesin: `routeMatch->getParameters()` returns a parameter bag but doesn't ensure the return type for parameters. Using `getRawParameters` should return only scalars. I believe `Url::fromRoute` requires parameters as scalars or objects that can be casted to a string.

modules/apigee_edge_teams/src/TeamContextManager.php
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCorrespondingRouteNameForEntity(EntityInterface $entity): ?string {
+    if ($current_route_object = $this->routeMatch->getRouteObject()) {
+      // If the route has only one and the same parameter, return the current
+      // route.
+      $parameters = $this->routeMatch->getRawParameters();
+      if ($parameters->has($entity->getEntityTypeId()) && $parameters->count() === 1) {
```
Jaesin: This is interesting. I see the problem with the corresponding route option is that it doesn't tell you if you should be showing the switcher to switch teams when there is no corresponding route to change to a developer route.
Some routes do have the `team` parameter but aren't compatible with other teams (such as `entity.subscription.team`). But that route has more than one parameter. I think this will work given the current state of things but I do worry about the switcher showing up where it isn't wanted because it is using the `{team}` parameter as it's only parameter in a custom implementation. I would almost rather have a static list of routes that is source of truth as to where the switcher shows up. for instance, I see the switcher shows up on `user/{uid}/orders` when the commerce module is installed. The team links fall back to their canonical links. We should see if this fallback is needed.

modules/apigee_edge_teams/css/apigee_edge_teams.switcher.css
Jaesin: image

Jaesin:
Functionally, the only thing I have questions about is if we need to fall back to using canonical urls for routes that don't have a corresponding team/developer link. For example, if you have the commerce module enabled, you will have the following path `user/1/orders`, I would expect that since this path isn't specifically an Apigee route, the switcher block would not be displayed at all. The way it is now, we see the switcher block with links to my team URLs i.e. `/team/foo`.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
shadcn: We need to get the entity label and `\Drupal\apigee_edge_teams\TeamContextManager::getDestinationUrlForEntity` needs an `EntityInterface`

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
shadcn: Same here. We need to get the entity label and `\Drupal\apigee_edge_teams\TeamContextManager::getDestinationUrlForEntity` needs an `EntityInterface`

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
shadcn: So I think this would be a UX issue. The user sees the switcher block, clicks on it to navigate to the `/teams` page which then has no switcher. Wouldn't this break the workflow?

modules/apigee_edge_teams/src/TeamContextManager.php
shadcn: `$parameters->all()` should handle this, right?

modules/apigee_edge_teams/src/TeamContextManager.php
shadcn: Yeah. Right now if we can build a URL from existing parameters and user has access to this URL, the switcher will be displayed and other entities fallback to their canonical URL. This is more a UX thing. On the legacy site, the switcher is available on all pages. It seems weird (bad UX) to have it on some pages only.

Let me know how to proceed. @Jaesin @cnovak

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
shadcn: I have a check for `instanceof EntityInterface`.

shadcn:
> Functionally, the only thing I have questions about is if we need to fall back to using canonical urls for routes that don't have a corresponding team/developer link. For example, if you have the commerce module enabled, you will have the following path `user/1/orders`, I would expect that since this path isn't specifically an Apigee route, the switcher block would not be displayed at all. The way it is now, we see the switcher block with links to my team URLs i.e. `/team/foo`.

Yeah. Right now if we can build a URL from existing parameters and user has access to this URL, the switcher will be displayed and other entities fallback to their canonical URL. This is more a UX thing. On the legacy site, the switcher is available on all pages. It seems weird (bad UX) to have it on some pages only.

Let me know how to proceed. @Jaesin @cnovak

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
Jaesin: OK. That makes sense.

modules/apigee_edge_teams/src/TeamContextManager.php
Jaesin: From what I understand, parameters->all() just gets you an array from the parameter bag but the values can still be either strings or object that were converted via a param converter.

Jaesin:
I did a quick demo and talked this over with cnovak today. For now, we shouldn't be concerned with the switcher block appearing and disappearing depending on the page you are on. Some dev portal users found this to be confusing because for some routes it didn't seem to do anything.

Another thing that was found to be confusing is the fallback to the canonical route. The links in the switcher sometimes take to you to a corresponding page and sometimes to the team route. As a casual user, you can't really be sure what's gong to happen if you click on a link in the selector.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
shadcn: Fixed.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
shadcn: Outdated. We are now showing the block on routes with corresponding route only.

shadcn:
@Jaesin I've implemented the suggestions. The block is now displayed if we have a corresponding route only. Ready for review.

mxr576:
I asked this before several times but I would like to ask this again because of the UX issues @Jaesin has described. Are you sure that this switcher is useful and necessary for the majority of the customers @cnovak? Because if only 1-2 customers would like a feature like this then it should be a custom solution for them and not part of the Apigee Edge module that we have to maintain and support.
What I see is as backend developer in this PR is an additional technical dept. If there is an already known UX issue with this feature should we enforce this?

I think the problem is the purpose of the team switcher is not clearly defined. (Or I do not know what you have discussed with @Jaesin.)
Mapping developer related routes to team-related routes is definitely a major complexity that is not easy to solve because there is not always a match. Like in case of the Commerce order routes.
If the purpose of this block to speed up the switching between teams then maybe this block should be only visible on team-related pages. This can be easily solved.
I would not try to include developer related routes to the switching, because then you have to deal with problems like what if there is no match for the current team or developer route.

modules/apigee_edge_teams/src/TeamContextManager.php
```diff
@@ -0,0 +1,104 @@
...
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Url;
+
+/**
+ * Describes the `apigee_edge_teams.context_manager` service.
+ *
+ * This service is responsible for understanding the context of the current
+ * route. Context will either be developer context or team context.
+ */
+class TeamContextManager implements TeamContextManagerInterface {
```
mxr576, line 33: Make this final to ensure nobody tries to extend something that is 100% internal. They can still encapsulate it if needed.

modules/apigee_edge_teams/src/Routing/TeamContextSwitcherRouteAlterSubscriber.php
```diff
@@ -0,0 +1,47 @@
...
+final class TeamContextSwitcherRouteAlterSubscriber extends RouteSubscriberBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function alterRoutes(RouteCollection $collection) {
+    foreach ($collection as $id => $route) {
+      // Add a corresponding team route if the team route defines a
+      // corresponding developer route.
+      if (($developer_route_id = $route->getOption(TeamContextManagerInterface::DEVELOPER_ROUTE_OPTION_NAME)) && ($developer_route = $collection->get($developer_route_id)) && empty($developer_route->getOption(TeamContextManagerInterface::TEAM_ROUTE_OPTION_NAME))) {
```
mxr576, line 41: `$route->getOption(TeamContextManagerInterface::DEVELOPER_ROUTE_OPTION_NAME)` only returns "entity.user.canonical" if it is not null which is kinda a contant in Drupal. That route always exists.
So I am not sure what the extra DEVELOPER_ROUTE_OPTION_NAME option on one canonical route tries to solve.
Please explain what you are trying to solve with this extra option.

mxr576: Okay, so you are trying to map user related routes to team-related routes with this magic I guess.
I think it would be easier to understand and more maintainable if you would just build a map in the team context switcher service from the currently known route matches and allow 3rd party modules (including the MINT module) to alter and extend this mapping with their own route. Have you considered this option?

My problem is with the current solution that it is hard to understand what it does and why it does it. Why XY routes are tagged and why others are not.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,240 @@
...
+      return [];
+    }
+
+    // Add a link for the developer account.
+    $entities = [
+      $this->entityTypeManager->getStorage('user')->load($this->account->id()),
+    ];
+
+    // Add links for teams.
+    if ($team_ids = $this->teamMembershipManager->getTeams($this->account->getEmail())) {
```
mxr576, line 149: This approach ignores proper access check on team-related routes, is it intentional? As a user, even if I am admin, I can only see those teams in this list that I am actually a member on Apigee Edge. So this invalidates the purpose of "Manage teams" and similar admin permissions and the entity access that can grant access to a team (view) even if you are not a member of the team (company) in Apigee Edge.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
```diff
@@ -0,0 +1,240 @@
...
+      $entities = array_merge($entities, $this->entityTypeManager->getStorage('team')
+        ->loadMultiple($team_ids));
+    }
+
+    $links = [];
+
+    /** @var \Drupal\Core\Entity\EntityInterface $entity */
+    foreach ($entities as $entity) {
+      // No link if we are on the current context route.
+      if ($current_context instanceof EntityInterface && $current_context->getEntityTypeId() === $entity->getEntityTypeId() && $current_context->id() === $entity->id()) {
```
mxr576, line 159: You could get rid of this condition in a loop if you would remove the current team's entity id from the ids that getTeams() returns. You could add this placeholder link to $links before the loop as the first element so this condition check would not run in every iteration.

modules/apigee_edge_teams/src/TeamContextManagerInterface.php
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Drupal\apigee_edge_teams;
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Url;
+
+/**
+ * Base definition of the team context manager service.
+ */
+interface TeamContextManagerInterface {
```
mxr576, line 29: I may rename this and the related class to TeamContextSwitcherManager or "Helper" instead of "Manager" because it is clearly can be only used with the switcher.

mxr576 (dismissed):

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
Jaesin: This doesn't bypass access but does limit to the teams the user is a member of (even admin). That is by design. Access is handled at the individual URL level.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
Jaesin: I agree. #optimization

modules/apigee_edge_teams/src/Routing/TeamContextSwitcherRouteAlterSubscriber.php
Jaesin: We did consider implementing a simple lookup table but the table would have to be updated and a hook would have to be created. Then that would all have to be cached.

The route option is already very well cached and it's exposed in the routing file so it is clear that a team route relates to a developer route in some way.

Maybe this could be more clear.

> // Add a corresponding team route if the team route defines a
> // corresponding developer route.

The context switcher is a teams feature and there is a 1-to-1 relationship between  a team route and a developer route and so it doesn't we shouldn't have to specify the developer route has a corresponding team route. This adds the corresponding team route to developer routes so they are easily accessible from the route object. Without this, we would have to reverse lookup the related team route from the developer route at run-time.

modules/apigee_edge_teams/src/TeamContextManager.php
Jaesin: The only thing that does is forces a decorator pattern over inheritance. Why is it important to do that?

modules/apigee_edge_teams/src/TeamContextManagerInterface.php
Jaesin: I was thinking this too since this implementation is pretty specific to providing for the team switcher but at some point I am hoping to get authoritative source for the question. "Is the route I'm on for developers or is it for teams". This can be determined by using `getCurrentContext` and checking the entity type. Maybe in the future we have `getCurrentContext` return `static::CONTEXT_TYPE_TEAM` or `CONTEXT_TYPE_DEVELOPER` so I think it makes sense to rename `getCurrentContext` to `getCurrentContextEntity` or something like that.

Jaesin (approved):
This has been tested manually and is working as expected.

modules/apigee_edge_teams/src/Plugin/Block/TeamContextSwitcherBlock.php
mxr576: Yeah, but for admins - I mean users with "Manage teams" permissions - this is misleading and rather a UX issue than a feature. "Why I can see XY team on the Teams page but I can not see it in the Team switcher?" They will ask...

modules/apigee_edge_teams/src/TeamContextManager.php
mxr576: 1. https://matthiasnoback.nl/2018/09/final-classes-by-default-why/
2. https://matthiasnoback.nl/2018/08/when-to-add-an-interface-to-a-class/
3. https://codeburst.io/inheritance-is-evil-stop-using-it-6c4f1caf5117

How this all applies to Symfony and Drupal 8 service classes?

You create an interface for a service in Drupal 8/Symfony because you and external developers should (must) depend on only public methods exposed on a service interface. Which means, everybody should only depend on the expected behavior of service methods instead of their concrete implementation (that you may ship in a class with your module).

Thanks for this approach, external developers can easily replace completely or just decorate the service implementations (classes) that you ship with your module with their own, they only need to create something which is compatible with the service interface.

Services usually have external dependencies that they get in constructor parameters. Drupal 8 BC policy does states that [constructor parameters may change between _minor_ versions](https://www.drupal.org/core/d8-bc-policy#constructors). _(Not sure how many Drupal developers knows what they should extend/override and what they should based on the BC policy. Developers do not read doc usually...)_

What does it mean for external developers?
If they still prefer inheritance over encapsulation then their code gets broken when a service's constructor gets changed.

How can we help them?
This is easy, we mark service classes as final so they do not try to extend it. They do not need to worry about constructor parameter changes because they cannot extend our concrete service implementations. Although they can always [inject the original service with "inner"](https://www.blueoakinteractive.com/blog/service-decorators-drupal-8) and leverage the actually re-usable components from the service - the public methods.

Just think about it, this is not just about constructors, the same problem occurs for public and protected methods and properties. As a maintainer, you are going to have bad days if you try to change these on your non-final classes because you may introduce BC breaking changes with those changes.

modules/apigee_edge_teams/src/Routing/TeamContextSwitcherRouteAlterSubscriber.php
mxr576: I wrote this because I also considered that what you described in the comment above and I saw better options for doing this. As you have already merged this PR and I stop thinking about how this can be done differently.

PR #147, opened 2019-03-06
[DRUP-609] Add API Product reference to API Docs.

arlina-espinoza:
API Docs should have an entity reference to API Products, so that views can be created to map API products to specs.
Note: By default the API product reference field is hidden on the "view", because we don't have renderer for the product yet (DRUP-400).


arlina-espinoza:
I signed it!

Cheers,
Arlina Espinoza

mxr576 (approved):

dawehner (approved):

PR #149, opened 2019-03-07
Fix errors regarding the API Docs entity settings route.

arlina-espinoza:
Fixes the following error regarding the "apidoc.settings" route:

`Symfony\Component\Routing\Exception\RouteNotFoundException: Route "apidoc.settings" does not exist. in Drupal\Core\Routing\RouteProvider->getRouteByName()...`


mxr576:
@arlina-espinoza Please add related JIRA issue number as a suffix to your PR title if there is any.

mxr576 (approved):

PR #150, opened 2019-03-07
Correctly show the entity label when the API Doc entity gets saved.

arlina-espinoza:
When saving an API Doc entity, the message displayed to the user is _"Saved the $label API Doc."_. This PR fixes the message to correctly show the entity name/label instead of $label.


mxr576:
@arlina-espinoza Please add related JIRA issue number as a suffix to your PR title if there is any.

mxr576 (approved):

PR #152, opened 2019-03-08
DRUP-581, DRUP-582: Authentication, key provider improvements

peterserfozo:
Latest Travis build: https://travis-ci.org/peterserfozo/apigee-edge-drupal/builds/505247582


src/Form/AuthenticationForm.php
```diff
@@ -167,13 +134,27 @@ protected function getEditableConfigNames() {
...
+    // Save the key for later use.
+    $this->activeKey = $this->getActiveKey();
 
+    /** @var \Drupal\apigee_edge\Plugin\KeyProviderRequirementsInterface $key_provider */
+    if (($key_provider = $this->activeKey->getKeyProvider()) instanceof KeyProviderRequirementsInterface) {
+      try {
+        $key_provider->checkRequirements($this->activeKey);
+      }
+      catch (KeyProviderRequirementsException $exception) {
+        $this->messenger()->addError($this->t("The requirements of the selected key provider (@key_provider) are not fulfilled. Fix the errors described below or <a href=':key_config_uri' target='_blank'>change the active key's provider</a>.", [
```
mxr576: s/the errors/errors/g

src/Form/AuthenticationForm.php
```diff
@@ -167,13 +134,27 @@ protected function getEditableConfigNames() {
...
+          '@key_provider' => $this->activeKey->getKeyProvider()->getPluginDefinition()['label'],
```
mxr576, line 160: Meh, `getKeyProvider()` returns a `\Drupal\key\Plugin\KeyProviderInterface` according to its documentation. `getPluginDefinition()` is only available if the returned key provider extends `\Drupal\key\Plugin\KeyProviderBase`. :-/

src/Form/AuthenticationForm.php
```diff
@@ -167,13 +134,27 @@ protected function getEditableConfigNames() {
...
+          ':key_config_uri' => Url::fromRoute('entity.key.edit_form', ['key' => $this->activeKey->id()])->toString(),
+        ]));
+        $this->messenger()->addError($exception->getTranslatableMarkupMessage());
```
mxr576, line 163: `getTranslatableMarkupMessage()` can return null, you should check whether the return value is null and in that case return a "Something went wrong..." message, in that case, you could display the non-translateable exception message maybe.

src/Exception/KeyProviderRequirementsException.php
```diff
@@ -0,0 +1,65 @@
...
+   *   The Exception message.
+   * @param \Drupal\Core\StringTranslation\TranslatableMarkup $translatable_markup_message
+   *   The translatable markup object of the exception to display on the pages
+   *   where the exception is caught.
+   * @param int|null $code
+   *   The error code.
+   * @param \Throwable|null $previous
+   *   The previous throwable used for the exception chaining.
+   */
+  public function __construct(string $message = '', TranslatableMarkup $translatable_markup_message = NULL, int $code = 0, \Throwable $previous = NULL) {
```
mxr576: I would make the `$message` at least mandatory here, because otherwise how else anybody could know why this exception was being thrown?

src/Form/AuthenticationForm.php
```diff
@@ -167,13 +134,27 @@ protected function getEditableConfigNames() {
...
+        return $form;
+      }
+    }
+
+    $form = parent::buildForm($form, $form_state);
```
mxr576: I think parent call could still happen at the beginning of the method. (Unless it adds something unwanted.)

src/Form/AuthenticationForm.php
```diff
@@ -167,13 +134,27 @@ protected function getEditableConfigNames() {
...
   /**
    * {@inheritdoc}
-   *
-   * @throws \Drupal\Core\Entity\EntityStorageException
-   * @throws \Drupal\key\Exception\KeyValueNotSetException
    */
   public function buildForm(array $form, FormStateInterface $form_state) {
-    $form = parent::buildForm($form, $form_state);
+    // Save the key for later use.
+    $this->activeKey = $this->getActiveKey();
```
mxr576: Why this is not called in the constructor?

src/Form/AuthenticationForm.php
```diff
@@ -266,7 +224,9 @@ public function buildForm(array $form, FormStateInterface $form_state) {
     $form['test_connection'] = [
       '#type' => 'details',
       '#title' => $this->t('Test connection'),
-      '#description' => 'Send request using the selected authentication key.',
+      '#description' => $this->keyIsWritable($this->activeKey) ? $this->t('Send request using the given API credentials.') : $this->t("Send request using the <a href=':key_config_uri' target='_blank'>active authentication key</a>.", [
+        ':key_config_uri' => Url::fromRoute('entity.key.edit_form', ['key' => $this->activeKey->id()])->toString(),
```
mxr576: Use `$this->activeKey->toUrl()->toString()` instead.

src/Form/AuthenticationForm.php
```diff
@@ -167,13 +134,27 @@ protected function getEditableConfigNames() {
...
 
+    /** @var \Drupal\apigee_edge\Plugin\KeyProviderRequirementsInterface $key_provider */
+    if (($key_provider = $this->activeKey->getKeyProvider()) instanceof KeyProviderRequirementsInterface) {
+      try {
+        $key_provider->checkRequirements($this->activeKey);
+      }
+      catch (KeyProviderRequirementsException $exception) {
+        $this->messenger()->addError($this->t("The requirements of the selected key provider (@key_provider) are not fulfilled. Fix the errors described below or <a href=':key_config_uri' target='_blank'>change the active key's provider</a>.", [
+          '@key_provider' => $this->activeKey->getKeyProvider()->getPluginDefinition()['label'],
+          ':key_config_uri' => Url::fromRoute('entity.key.edit_form', ['key' => $this->activeKey->id()])->toString(),
```
mxr576: Use `$this->activeKey->toUrl()->toString()` instead.

src/Form/AuthenticationForm.php
```diff
@@ -302,12 +262,15 @@ public function buildForm(array $form, FormStateInterface $form_state) {
   public function validateForm(array &$form, FormStateInterface $form_state) {
     parent::validateForm($form, $form_state);
 
-    $active_key = $this->activeKey;
+    // Throw an error if the key is not writable but the form is submitted.
```
mxr576: How this is possible? If actions are disabled this should not happen.

src/Form/AuthenticationForm.php
```diff
@@ -637,26 +591,19 @@ protected function getActiveKey(): ?KeyInterface {
...
-    // Create a new key name.
+  protected function generateNewAuthKey(): KeyInterface {
     $new_key_id = 'apigee_edge_connection_default';
-    $file_path = "private://.apigee_edge/{$new_key_id}.json";
-
-    // Make sure the key and the associated key file do not exist.
-    for ($i = 1; (Key::load($new_key_id) || file_exists($file_path)); $i++) {
-      $new_key_id = "apigee_edge_connection_default_{$i}";
-      $file_path = "private://.apigee_edge/{$new_key_id}.json";
+    if (Key::load($new_key_id) !== NULL) {
```
mxr576: I would inject the entity type manager - or the key entity storage - and use that to CRUDL keys because it would allow mocking CRUDL related operations in tests.

src/Form/AuthenticationForm.php
```diff
@@ -670,15 +617,22 @@ protected function generateNewAuthKey(): ?KeyInterface {
...
-    $provider->setKeyValue($new_key, Json::encode((object) ['auth_type' => EdgeKeyTypeInterface::EDGE_AUTH_TYPE_BASIC]));
+    try {
+      /** @var \Drupal\apigee_edge\Plugin\KeyProviderRequirementsInterface $key_provider */
+      if (($key_provider = $new_key->getKeyProvider()) instanceof KeyProviderRequirementsInterface) {
+        $key_provider->checkRequirements($new_key);
+        // Write out an empty key.
+        $key_provider->setKeyValue($new_key, Json::encode((object) ['auth_type' => EdgeKeyTypeInterface::EDGE_AUTH_TYPE_BASIC]));
+      }
+    }
+    catch (\Exception $exception) {
```
mxr576: Probably only key requirements exceptions should be caught here.

src/Plugin/KeyProvider/EnvironmentVariablesKeyProvider.php
```diff
@@ -105,36 +140,42 @@ public function getKeyValue(KeyInterface $key) {
...
+      }
+    }
 
-  /**
-   * {@inheritdoc}
-   */
-  public static function obscureKeyValue($key_value, array $options = []) {
-    return $key_value;
+    if (!empty($missing_env_variables)) {
+      throw new KeyProviderRequirementsException($this->t('The following environment variables are not set: @missing_env_variables.', [
```
mxr576: The second parameter is the translatable string. :)

src/Plugin/KeyProvider/EnvironmentVariablesKeyProvider.php
```diff
@@ -65,17 +97,11 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
...
+    try {
+      $this->checkRequirements($form_state->getFormObject()->getEntity());
     }
-
-    if (!empty($missing_env_variables)) {
-      $form_state->setError($form, $this->t('The following environment variables are not set: @missing_env_variables.', [
-        '@missing_env_variables' => implode(', ', $missing_env_variables),
-      ]));
+    catch (KeyProviderRequirementsException $exception) {
+      $form_state->setError($form, $exception->getTranslatableMarkupMessage());
```
mxr576, line 104: Defensive programming, the `getTranslatableMarkupMessage()` can still be null. At least I would use this simple pattern `$form_state->setError($form, $exception->getTranslatableMarkupMessage() ?? $this->t('Key provider error: @error', ['@error' => $e->getMessage()]))`

Hm, maybe even better. Use ` $this->t('Key provider error: @error', ['@error' => $e->getMessage()])` as default fallback return value in `getTranslatableMarkupMessage()` instead of null. That would simplify usage of this exception.

src/Plugin/KeyProvider/PrivateFileKeyProvider.php
```diff
@@ -102,7 +124,15 @@ public function submitConfigurationForm(array &$form, FormStateInterface $form_s
...
+    try {
+      $this->checkRequirements($key);
+    }
+    catch (KeyProviderRequirementsException $exception) {
+      $context = [
+        '@message' => (string) $exception,
+      ];
+      $context += Error::decodeException($exception);
+      $this->logger->error('Could not retrieve Apigee Edge authentication key value from the private file storage: @message %function (line %line of %file). <pre>@backtrace_string</pre>', $context);
       return NULL;
```
mxr576, line 139: Please add a note that maybe returning `null` is not the best possible option but key's design does not allow to throw an exception and place a reference to this D.o issue: https://www.drupal.org/project/key/issues/3038212

src/Plugin/KeyProvider/PrivateFileKeyProvider.php
```diff
@@ -111,20 +141,26 @@ public function getKeyValue(KeyInterface $key) {
...
+    try {
+      $this->checkRequirements($key);
+    }
+    catch (KeyProviderRequirementsException $exception) {
+      $context = [
+        '@message' => (string) $exception,
+      ];
+      $context += Error::decodeException($exception);
+      $this->logger->error('Could not save Apigee Edge authentication key value in the private file storage: @message %function (line %line of %file). <pre>@backtrace_string</pre>', $context);
+      return FALSE;
```
mxr576, line 161: Please add the same note as I suggested above.

src/Plugin/KeyProvider/PrivateFileKeyProvider.php
```diff
@@ -111,20 +141,26 @@ public function getKeyValue(KeyInterface $key) {
...
     $file_uri = $this->getFileUri($key);
     $file_path = dirname($file_uri);
     // Make sure the folder exists.
     file_prepare_directory($file_path, FILE_CREATE_DIRECTORY | FILE_MODIFY_PERMISSIONS);
     // Save the token data.
-    if (!file_unmanaged_save_data($key_value, $file_uri, FILE_EXISTS_REPLACE)) {
-      $file_full_path = realpath($file_uri);
-      throw new KeyValueNotSetException("Error saving connection settings to file ({$file_full_path}). Make sure the directory is writable.");
-    }
+    return file_unmanaged_save_data($key_value, $file_uri, FILE_EXISTS_REPLACE);
```
mxr576, line 171: Procedural functions from file.inc will be deprecated in Drupal 8.7.
Let's use successor file system service method calls instead of them everywhere it is possible.

https://github.com/drupal/core/blob/8.7.x/includes/file.inc#L921

src/Plugin/KeyProvider/PrivateFileKeyProvider.php
```diff
@@ -111,20 +141,26 @@ public function getKeyValue(KeyInterface $key) {
...
+      ];
+      $context += Error::decodeException($exception);
+      $this->logger->error('Could not save Apigee Edge authentication key value in the private file storage: @message %function (line %line of %file). <pre>@backtrace_string</pre>', $context);
+      return FALSE;
+    }
+
     $file_uri = $this->getFileUri($key);
     $file_path = dirname($file_uri);
     // Make sure the folder exists.
     file_prepare_directory($file_path, FILE_CREATE_DIRECTORY | FILE_MODIFY_PERMISSIONS);
```
mxr576, line 168: Procedural functions from file.inc will be deprecated in Drupal 8.7.
Let's use successor file system service method calls instead of them everywhere it is possible.

https://github.com/drupal/core/blob/8.7.x/includes/file.inc#L317

src/Plugin/KeyProvider/PrivateFileKeyProvider.php
```diff
@@ -111,20 +141,26 @@ public function getKeyValue(KeyInterface $key) {
...
+      $context = [
+        '@message' => (string) $exception,
+      ];
+      $context += Error::decodeException($exception);
+      $this->logger->error('Could not save Apigee Edge authentication key value in the private file storage: @message %function (line %line of %file). <pre>@backtrace_string</pre>', $context);
+      return FALSE;
+    }
+
     $file_uri = $this->getFileUri($key);
     $file_path = dirname($file_uri);
```
mxr576: Procedural functions from file.inc will be deprecated in Drupal 8.7.
Let's use successor file system service method calls instead of them everywhere it is possible.

https://github.com/drupal/core/blob/8.7.x/includes/file.inc#L1107

src/Plugin/KeyProvider/PrivateFileKeyProvider.php
```diff
@@ -134,6 +170,27 @@ public function deleteKeyValue(KeyInterface $key) {
     return file_unmanaged_delete($this->getFileUri($key));
```
mxr576, line 179: https://github.com/drupal/core/blob/8.7.x/includes/file.inc#L828

src/Plugin/KeyProviderRequirementsInterface.php
```diff
@@ -0,0 +1,41 @@
...
+   * Checks the requirements of the key provider.
+   *
+   * @param \Drupal\key\KeyInterface $key
+   *   The key entity.
+   *
+   * @throws \Drupal\apigee_edge\Exception\KeyProviderRequirementsException
+   *   Exception thrown when the requirements of the key provider are not
+   *   fulfilled.
+   */
+  public function checkRequirements(KeyInterface $key);
```
mxr576: Missing void return type.

modules/apigee_edge_teams/src/Entity/Storage/TeamMemberRoleStorage.php
```diff
@@ -258,8 +258,10 @@ protected function doSave($id, EntityInterface $entity) {
   protected function doDelete($entities) {
     /** @var \Drupal\apigee_edge_teams\Entity\TeamMemberRoleInterface $entity */
     foreach ($entities as $entity) {
-      // See explanation in doSave().
-      Cache::invalidateTags($entity->getTeam()->getCacheTags());
+      if (($team = $entity->getTeam()) !== NULL) {
```
mxr576, line 263: Please document that this small sanity check allows uninstalling the module if the Apigee Edge connection would be broken.

mxr576 (changes requested):

mxr576:
@peterserfozo Please also address the following small issues in this PR:
* Form id should be `apigee_edge_authentication_form` not `apigee_edge_authentication`.
* Use either `json_encode()` or `Json::encode()` in the AuthenticationForm but not both.

src/Form/AuthenticationForm.php
peterserfozo: I've moved the parent call after the key provider requirement check because that adds the `actions` container (and we should hide that if the key provider requirements are not fulfilled).

mxr576: I think it would be better to hide that with `#access => FALSE` instead of completely removing the array elements that should be there in the `$form` render array.

mxr576 (approved):

PR #159, opened 2019-03-14
Add missing @return tag in function comment

heykarthikwithu:


src/Entity/Form/AppCreateForm.php
```diff
@@ -110,6 +110,10 @@ protected function alterFormWithApiProductElement(array &$form, FormStateInterfa
    * @param \Drupal\Core\Form\FormStateInterface $form_state
    *   The form state object.
    *
+   * @return array $elements
```
mxr576: Weird that PHPCS did not complain about this in tests, it should have. But if you add the `@return` tag, always describe for humans what a function/method returns. Here it contains the "The API product render element."

mxr576 (changes requested):

mxr576:
https://travis-ci.org/apigee/apigee-edge-drupal/jobs/506176707 tests are failing

src/Entity/Form/AppCreateForm.php
heykarthikwithu: Added desription to the @return tag in function comment

heykarthikwithu:
@mxr576 fixed now..

mxr576 (approved):

mxr576:
Thanks!

PR #161, opened 2019-03-20
DRUP-624 Add revisions to API Docs

arlina-espinoza:
This will add revisioning to the API Docs entities:

- A log message when editing.
- Revision history for the entity.
- Viewing past revisions.
- Reverting to past revisions.
- Granular permissions to view and revert revisions.
- Tests.

### DIFF
--- a/modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
+++ b/modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
@@ -22,6 +22,7 @@ namespace Drupal\apigee_edge_apidocs;
 
 use Drupal\Core\Entity\EntityAccessControlHandler;
 use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Access\AccessResult;
 
@@ -33,6 +34,20 @@ use Drupal\Core\Access\AccessResult;
 class ApiDocAccessControlHandler extends EntityAccessControlHandler {
 
   /**
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(EntityTypeInterface $entity_type) {
+    parent::__construct($entity_type);
+
+    $this->entityTypeManager = \Drupal::entityTypeManager();
+  }
+
+  /**
    * {@inheritdoc}
    */
   protected function checkAccess(EntityInterface $entity, $operation, AccountInterface $account) {
@@ -40,6 +55,12 @@ class ApiDocAccessControlHandler extends EntityAccessControlHandler {
 
     if (!$parent_access->isAllowed()) {
       /** @var \Drupal\apigee_edge_apidocs\Entity\ApiDocInterface $entity */
+
+      // Access control for revisions.
+      if (!$entity->isDefaultRevision()) {
+        return $this->checkAccessRevisions($entity, $operation, $account);
+      }
+
       switch ($operation) {
         case 'view':
           if (!$entity->isPublished()) {
@@ -57,6 +78,60 @@ class ApiDocAccessControlHandler extends EntityAccessControlHandler {
 
     // Unknown operation, no opinion.
     return $parent_access;
+  }
+
+  /**
+   * Additional access control for revisions.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface   $entity
+   * @param                                       $operation
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *
+   * @return bool
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  protected function checkAccessRevisions(EntityInterface $entity, $operation, AccountInterface $account) {
+    $entity_access = $this->entityTypeManager->getAccessControlHandler($this->entityTypeId);
+
+    /** @var \Drupal\Core\Entity\EntityStorageInterface $entity_storage */
+    $entity_storage = $this->entityTypeManager->getStorage($this->entityTypeId);
+
+    // Must have access to the same operation on the default revision.
+    $default_revision = $entity_storage->load($entity->id());
+    $entity_access_default = $default_revision->access($operation, $account);
+    if (!$entity_access_default) {
+      return AccessResult::forbidden();
+    }
+
+    $map = [
+      'view' => "view all {$this->entityTypeId} revisions",
+      'update' => "revert all {$this->entityTypeId} revisions",
+      'delete' => "delete all {$this->entityTypeId} revisions",
+    ];
+    $bundle = $entity->bundle();
+    $type_map = [
+      'view' => "view {$this->entityTypeId} $bundle revisions",
+      'update' => "revert {$this->entityTypeId} $bundle revisions",
+      'delete' => "delete {$this->entityTypeId} $bundle revisions",
+    ];
+
+    if (!$entity || !isset($map[$operation]) || !isset($type_map[$operation])) {
+      // If there was no entity to check against, or the $op was not one of the
+      // supported ones, we return access denied.
+      return AccessResult::forbidden();
+    }
+
+    $admin_permission = $this->entityType->getAdminPermission();
+
+    // Perform basic permission checks first.
+    if ($account->hasPermission($map[$operation]) ||
+      $account->hasPermission($type_map[$operation]) ||
+      ($admin_permission && $account->hasPermission($admin_permission))) {
+      return AccessResult::allowed();
+    }
+
+    return AccessResult::forbidden();
   }
 
   /**
--- a/modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
+++ b/modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
@@ -34,6 +34,8 @@ use Drupal\Core\Access\AccessResult;
 class ApiDocAccessControlHandler extends EntityAccessControlHandler {
 
   /**
+   * The entity type manager.
+   *
    * @var \Drupal\Core\Entity\EntityTypeManagerInterface
    */
   protected $entityTypeManager;
@@ -83,17 +85,20 @@ class ApiDocAccessControlHandler extends EntityAccessControlHandler {
   /**
    * Additional access control for revisions.
    *
-   * @param \Drupal\Core\Entity\EntityInterface   $entity
-   * @param                                       $operation
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity for which to check access.
+   * @param string $operation
+   *   The entity operation.
    * @param \Drupal\Core\Session\AccountInterface $account
+   *   The user for which to check access.
    *
-   * @return bool
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   *
    * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
    * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
    */
   protected function checkAccessRevisions(EntityInterface $entity, $operation, AccountInterface $account) {
-    $entity_access = $this->entityTypeManager->getAccessControlHandler($this->entityTypeId);
-
     /** @var \Drupal\Core\Entity\EntityStorageInterface $entity_storage */
     $entity_storage = $this->entityTypeManager->getStorage($this->entityTypeId);
 
--- a/modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
+++ b/modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
@@ -21,17 +21,20 @@
 namespace Drupal\apigee_edge_apidocs;
 
 use Drupal\Core\Entity\EntityAccessControlHandler;
+use Drupal\Core\Entity\EntityHandlerInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Access\AccessResult;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Access controller for the API Doc entity.
  *
  * @see \Drupal\apigee_edge_apidocs\Entity\ApiDoc.
  */
-class ApiDocAccessControlHandler extends EntityAccessControlHandler {
+class ApiDocAccessControlHandler extends EntityAccessControlHandler implements EntityHandlerInterface {
 
   /**
    * The entity type manager.
@@ -41,12 +44,26 @@ class ApiDocAccessControlHandler extends EntityAccessControlHandler {
   protected $entityTypeManager;
 
   /**
+   * Constructs an access control handler instance.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
+   *   The entity type definition.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entityTypeManager
+   *   The entity type manager.
+   */
+  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entityTypeManager) {
+    parent::__construct($entity_type);
+    $this->entityTypeManager = $entityTypeManager;
+  }
+
+  /**
    * {@inheritdoc}
    */
-  public function __construct(EntityTypeInterface $entity_type) {
-    parent::__construct($entity_type);
-
-    $this->entityTypeManager = \Drupal::entityTypeManager();
+  public static function createInstance(ContainerInterface $container, EntityTypeInterface $entity_type) {
+    return new static(
+      $entity_type,
+      $container->get('entity_type.manager')
+    );
   }
 
   /**
--- a/modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
+++ b/modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
@@ -111,9 +111,6 @@ class ApiDocAccessControlHandler extends EntityAccessControlHandler implements E
    *
    * @return \Drupal\Core\Access\AccessResultInterface
    *   The access result.
-   *
-   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
-   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
    */
   protected function checkAccessRevisions(EntityInterface $entity, $operation, AccountInterface $account) {
     /** @var \Drupal\Core\Entity\EntityStorageInterface $entity_storage */
@@ -121,8 +118,8 @@ class ApiDocAccessControlHandler extends EntityAccessControlHandler implements E
 
     // Must have access to the same operation on the default revision.
     $default_revision = $entity_storage->load($entity->id());
-    $entity_access_default = $default_revision->access($operation, $account);
-    if (!$entity_access_default) {
+    $has_default_entity_rev_access = $default_revision->access($operation, $account);
+    if (!$has_default_entity_rev_access) {
       return AccessResult::forbidden();
     }
 
--- a/modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
+++ b/modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
@@ -139,13 +139,6 @@ class ApiDoc extends EditorialContentEntityBase implements ApiDocInterface {
   /**
    * {@inheritdoc}
    */
-  public function getRevisionUser() {
-    return $this->get('revision_user')->entity;
-  }
-
-  /**
-   * {@inheritdoc}
-   */
   public function preSave(EntityStorageInterface $storage) {
     parent::preSave($storage);
 
@@ -164,9 +157,9 @@ class ApiDoc extends EditorialContentEntityBase implements ApiDocInterface {
 
     if (!$this->isNewRevision() && isset($this->original) && empty($record->revision_log_message)) {
       // If we are updating an existing entity without adding a new revision, we
-      // need to make sure $entity->revision_log is reset whenever it is empty.
-      // Therefore, this code allows us to avoid clobbering an existing log
-      // entry with an empty one.
+      // need to make sure $entity->revision_log_message is reset whenever it is
+      // empty. Therefore, this code allows us to avoid clobbering an existing
+      // log  entry with an empty one.
       $record->revision_log_message = $this->original->revision_log_message->value;
     }
   }
--- a/modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
+++ b/modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
@@ -20,8 +20,10 @@
 
 namespace Drupal\apigee_edge_apidocs\Entity;
 
+use Drupal\Core\Entity\EditorialContentEntityBase;
+use Drupal\Core\Entity\EntityStorageInterface;
+use Drupal\Core\Entity\RevisionableInterface;
 use Drupal\Core\Field\BaseFieldDefinition;
-use Drupal\Core\Entity\ContentEntityBase;
 use Drupal\Core\Entity\EntityChangedTrait;
 use Drupal\Core\Entity\EntityTypeInterface;
 
@@ -34,6 +36,8 @@ use Drupal\Core\Entity\EntityTypeInterface;
  *   label_singular = @Translation("API Doc"),
  *   label_plural = @Translation("API Docs"),
  *   handlers = {
+ *
+ *     "storage" = "Drupal\Core\Entity\Sql\SqlContentEntityStorage",
  *     "view_builder" = "Drupal\Core\Entity\EntityViewBuilder",
  *     "list_builder" = "Drupal\apigee_edge_apidocs\ApiDocListBuilder",
  *     "views_data" = "Drupal\views\EntityViewsData",
@@ -47,10 +51,14 @@ use Drupal\Core\Entity\EntityTypeInterface;
  *     "access" = "Drupal\apigee_edge_apidocs\ApiDocAccessControlHandler",
  *     "route_provider" = {
  *       "html" = "Drupal\apigee_edge_apidocs\ApiDocHtmlRouteProvider",
+ *       "revision" = "Drupal\entity\Routing\RevisionRouteProvider",
  *     },
  *   },
  *   base_table = "apidoc",
  *   data_table = "apidoc_field_data",
+ *   revision_table = "apidoc_revision",
+ *   revision_data_table = "apidoc_field_revision",
+ *   show_revision_ui = TRUE,
  *   translatable = TRUE,
  *   admin_permission = "administer apidoc entities",
  *   entity_keys = {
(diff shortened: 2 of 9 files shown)


arlina-espinoza:
Added ApiDoc settings for configuring default revision options.
API Doc settings   Drupal 8 m10n

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
```diff
@@ -32,6 +33,22 @@
...
 
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * {@inheritdoc}
```
mxr576: Do not use `{@inheritdoc}` because it does not explain the actual dependencies.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
```diff
@@ -32,6 +33,22 @@
...
+   */
+  public function __construct(EntityTypeInterface $entity_type) {
+    parent::__construct($entity_type);
+
+    $this->entityTypeManager = \Drupal::entityTypeManager();
```
mxr576: Implement the `\Drupal\Core\Entity\EntityHandlerInterface`, that should allow to dependency inject the entity type manager service.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
```diff
@@ -59,6 +82,63 @@ protected function checkAccess(EntityInterface $entity, $operation, AccountInter
...
+   *   The entity for which to check access.
+   * @param string $operation
+   *   The entity operation.
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The user for which to check access.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
```
mxr576: In _this_ context, I'd not document these exceptions because an entity with `$this->entityTypeId` should always exist therefore exceptions like these should not be thrown.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
```diff
@@ -59,6 +82,63 @@ protected function checkAccess(EntityInterface $entity, $operation, AccountInter
...
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  protected function checkAccessRevisions(EntityInterface $entity, $operation, AccountInterface $account) {
+    /** @var \Drupal\Core\Entity\EntityStorageInterface $entity_storage */
+    $entity_storage = $this->entityTypeManager->getStorage($this->entityTypeId);
+
+    // Must have access to the same operation on the default revision.
+    $default_revision = $entity_storage->load($entity->id());
+    $entity_access_default = $default_revision->access($operation, $account);
```
mxr576: I'd call it `has_default_entity_rev_access` or something like that because `has` and `is` indicates that it is a boolean value.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
```diff
@@ -59,6 +82,63 @@ protected function checkAccess(EntityInterface $entity, $operation, AccountInter
...
+    if (!$entity_access_default) {
+      return AccessResult::forbidden();
+    }
+
+    $map = [
+      'view' => "view all {$this->entityTypeId} revisions",
```
mxr576, line 127: I think you are unintentionally mixing what is the scope that an entity access handler should handle and what is the scope of a route access handler.
https://github.com/fago/entity/blob/85f76b56f25d7c3f5bc466f94db30a9b63d97fc5/src/Access/EntityRevisionRouteAccessChecker.php#L86 This is a route access handler and you can see the similarity between your implementation and this class.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
```diff
@@ -59,6 +82,63 @@ protected function checkAccess(EntityInterface $entity, $operation, AccountInter
...
+   */
+  protected function checkAccessRevisions(EntityInterface $entity, $operation, AccountInterface $account) {
+    /** @var \Drupal\Core\Entity\EntityStorageInterface $entity_storage */
+    $entity_storage = $this->entityTypeManager->getStorage($this->entityTypeId);
+
+    // Must have access to the same operation on the default revision.
+    $default_revision = $entity_storage->load($entity->id());
+    $entity_access_default = $default_revision->access($operation, $account);
+    if (!$entity_access_default) {
+      return AccessResult::forbidden();
```
mxr576, line 123: Results should be cached in a static storage.

For example: https://github.com/fago/entity/blob/85f76b56f25d7c3f5bc466f94db30a9b63d97fc5/src/Access/EntityRevisionRouteAccessChecker.php#L124

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
mxr576: So probably this is the right place to ask again, can we get rid of the custom entity access and entity route handlers that we implemented and rely on what the Entity module provides? 
I would feel much safer if access handlers from a well-tested framework would ensure the entity access in our system.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -122,16 +139,48 @@ public function setCreatedTime(int $timestamp) : ApiDocInterface {
   /**
    * {@inheritdoc}
    */
-  public function isPublished() : bool {
-    return (bool) $this->getEntityKey('status');
+  public function getRevisionUser() {
```
mxr576: Why did you override the inherited implementation from `\Drupal\Core\Entity\RevisionLogEntityTrait::getRevisionUser()`?

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -122,16 +139,48 @@ public function setCreatedTime(int $timestamp) : ApiDocInterface {
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function preSaveRevision(EntityStorageInterface $storage, \stdClass $record) {
+    parent::preSaveRevision($storage, $record);
+
+    if (!$this->isNewRevision() && isset($this->original) && empty($record->revision_log_message)) {
+      // If we are updating an existing entity without adding a new revision, we
+      // need to make sure $entity->revision_log is reset whenever it is empty.
```
mxr576: I think you have copy-pasted this from `\Drupal\node\Entity\Node::preSaveRevision()` but you changed the of the class property to `revision_log_message` so either the comment or the property name is incorrect in this context.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -47,30 +51,43 @@
...
+ *     "version-history" = "/admin/structure/apidoc/{apidoc}/revisions",
+ *     "revision" = "/admin/structure/apidoc/{apidoc}/revisions/{apidoc_revision}/view",
+ *     "revision-revert-form" = "/admin/structure/apidoc/{apidoc}/revisions/{apidoc_revision}/revert",
  *     "collection" = "/admin/structure/apidoc",
  *   },
  *   field_ui_base_route = "apigee_edge_apidocs.settings"
  * )
  */
-class ApiDoc extends ContentEntityBase implements ApiDocInterface {
+class ApiDoc extends EditorialContentEntityBase implements ApiDocInterface {
```
mxr576, line 90: After I saw copy-pasting from `\Drupal\node\Entity\Node` several places here I need to ask, is there any reason why this class does not extend Node? That way even the node access would work on the API documentation (content) entity type, which is a requirement if we would like to support any kind of search module.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -122,16 +139,48 @@ public function setCreatedTime(int $timestamp) : ApiDocInterface {
...
    */
-  public function setPublished(bool $published) : ApiDocInterface {
-    $this->set('status', $published);
-    return $this;
+  public function preSave(EntityStorageInterface $storage) {
+    parent::preSave($storage);
+
+    // If no revision author has been set explicitly, make the current user
+    // the revision author.
+    if (!$this->getRevisionUser()) {
```
mxr576, line 147: I like better what Node does.

```
  // If no revision author has been set explicitly, make the node owner the
    // revision author.
    if (!$this->getRevisionUser()) {
      $this->setRevisionUserId($this->getOwnerId());
    }
```

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -214,24 +267,38 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
     return $fields;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  protected function urlRouteParameters($rel) {
+    $uri_route_parameters = parent::urlRouteParameters($rel);
+
+    if ($rel === 'revision-revert-form' && $this instanceof RevisionableInterface) {
```
mxr576, line 290: I think this is something that the entity route provider should handle.
https://github.com/fago/entity/blob/062c7c23900d0e5ef32060c2500143fcdeecbff5/src/Routing/RevisionRouteProvider.php#L76

modules/apigee_edge_apidocs/src/Entity/ApiDocInterface.php
```diff
@@ -85,25 +85,4 @@ public function getCreatedTime() : int;
...
 
-  /**
-   * Returns the API Doc published status indicator.
-   *
-   * Unpublished API Doc are only visible to restricted users.
-   *
-   * @return bool
-   *   TRUE if the API Doc is published.
-   */
-  public function isPublished() : bool;
```
mxr576, line 96: Why did you remove this? Isn't it a requirement for an API documentation entity have published/unpublished state?
You probably wanted to extend the `\Drupal\Core\Entity\EntityPublishedInterface` here.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
```diff
@@ -41,13 +42,55 @@ public function getFormId() {
    * {@inheritdoc}
    */
   public function submitForm(array &$form, FormStateInterface $form_state) {
+    $entity_type = \Drupal::entityTypeManager()
```
mxr576: Inject the entity type manager service.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
```diff
@@ -41,13 +42,55 @@ public function getFormId() {
...
+      ->getEntityType();
+    $config = $this->configFactory()->getEditable('apigee_edge_apidocs.settings');
+
+    $options = $form_state->getValue('options');
+    $config->set('default_revision', (bool) $options['new_revision'])->save();
+
+    $args = [
+      '@type' => $entity_type->getLabel(),
+    ];
+    drupal_set_message($this->t('@type settings have been updated.', $args));
```
mxr576: `$this->messenger()`

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
```diff
@@ -41,13 +42,55 @@ public function getFormId() {
...
   public function buildForm(array $form, FormStateInterface $form_state) {
     $form['apidoc_settings']['#markup'] = $this->t('Settings for API Docs. Manage field settings using the tabs above.');
+
+    $form['additional_settings'] = [
+      '#type' => 'vertical_tabs',
+    ];
+
+    $form['workflow'] = [
+      '#type' => 'details',
+      '#title' => t('Publishing options'),
```
mxr576: `$this->t()`

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
```diff
@@ -41,13 +42,55 @@ public function getFormId() {
...
   public function submitForm(array &$form, FormStateInterface $form_state) {
+    $entity_type = \Drupal::entityTypeManager()
+      ->getStorage('apidoc')
+      ->getEntityType();
+    $config = $this->configFactory()->getEditable('apigee_edge_apidocs.settings');
+
+    $options = $form_state->getValue('options');
+    $config->set('default_revision', (bool) $options['new_revision'])->save();
+
+    $args = [
```
mxr576: An unnecessary new variable, you can pass this directly to `$this->messenger()`

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -122,16 +139,48 @@ public function setCreatedTime(int $timestamp) : ApiDocInterface {
...
+    }
+  }
+
+  /**
+   * Gets whether a new revision should be created by default.
+   *
+   * @return bool
+   *   TRUE if a new revision should be created by default.
+   */
+  public static function shouldCreateNewRevision() {
```
mxr576: Is it really necessary to be static? `@see \Drupal\node\Entity\NodeType::shouldCreateNewRevision()`

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
```diff
@@ -41,13 +42,55 @@ public function getFormId() {
...
+      '#type' => 'vertical_tabs',
+    ];
+
+    $form['workflow'] = [
+      '#type' => 'details',
+      '#title' => t('Publishing options'),
+      '#group' => 'additional_settings',
+    ];
+    $workflow_options = [
+      'new_revision' => ApiDoc::shouldCreateNewRevision(),
```
mxr576: I would avoid calling static methods (See my comment above) and hardcoding our ApiDoc class to the code. Default entity classes can be swapped by custom modules.
The current entity type object that belongs to the API documentation entity type should be retrieved from the entity type manager and the `shouldCreateNewRevision()` method should be called on that object.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
```diff
@@ -41,13 +42,55 @@ public function getFormId() {
...
+    ];
+    // Prepare workflow options to be used for 'checkboxes' form element.
+    $keys = array_keys(array_filter($workflow_options));
+    $workflow_options = array_combine($keys, $keys);
+    $form['workflow']['options'] = [
+      '#type' => 'checkboxes',
+      '#title' => t('Default options'),
```
mxr576: $this->t()....

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
```diff
@@ -41,13 +42,55 @@ public function getFormId() {
...
+      '#default_value' => $workflow_options,
+      '#options' => [
+        'new_revision' => t('Create new revision'),
```
mxr576: $this->t()....

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
```diff
@@ -41,13 +42,55 @@ public function getFormId() {
...
+    $form['workflow'] = [
+      '#type' => 'details',
+      '#title' => t('Publishing options'),
+      '#group' => 'additional_settings',
+    ];
+    $workflow_options = [
+      'new_revision' => ApiDoc::shouldCreateNewRevision(),
+    ];
+    // Prepare workflow options to be used for 'checkboxes' form element.
+    $keys = array_keys(array_filter($workflow_options));
```
mxr576: Use a more meaningful variable name.

modules/apigee_edge_apidocs/tests/src/Kernel/ApidocEntityTest.php
```diff
@@ -60,4 +78,56 @@ public function testEntity() {
...
+    $entity->setNewRevision();
+    $entity->setRevisionLogMessage($new_log);
+    $entity->save();
+    $v2_id = $entity->getRevisionId();
+    $this->assertTrue($v2_id > $v1_id);
+
+    // Test saving without a new revision.
+    $entity->setDescription('Test API v3');
+    $entity->save();
+    $this->assertTrue($v2_id === $entity->getRevisionId());
```
mxr576: assertEquals()

modules/apigee_edge_apidocs/tests/src/Kernel/ApidocEntityTest.php
```diff
@@ -60,4 +78,56 @@ public function testEntity() {
...
+    $this->assertNotNull($v1_id);
+
+    // Test saving a new revision.
+    $new_log = 'v2';
+    $entity->setDescription('Test API v2');
+    $entity->setNewRevision();
+    $entity->setRevisionLogMessage($new_log);
+    $entity->save();
+    $v2_id = $entity->getRevisionId();
+    $this->assertTrue($v2_id > $v1_id);
```
mxr576: assertGreaterThan()

modules/apigee_edge_apidocs/tests/src/Kernel/ApidocEntityRevisionsAccessTest.php
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  /**
+   * Test ApiDocs revision access as a logged in user with admin permissions.
+   */
+  public function testApiDocRevisionsAccessAdmin() {
+    $user = $this->createUser([
+      'administer apidoc entities',
+    ]);
+    \Drupal::currentUser()->setAccount($user);
```
mxr576: Inject and use the current_user service or use the account_switcher current. I do not think that this is the right way to switch user in tests.

modules/apigee_edge_apidocs/tests/src/Kernel/ApidocEntityRevisionsAccessTest.php
```diff
@@ -0,0 +1,200 @@
...
+
+    $this->installEntitySchema('user');
+    $this->installEntitySchema('apidoc');
+    $this->installSchema('system', ['sequences']);
+
+    $this->entityTypeManager = $this->container->get('entity_type.manager');
+    $this->entityTypeStorage = $this->entityTypeManager->getStorage('apidoc');
+
+    // Create a published apidoc.
+    $apidoc = ApiDoc::create([
```
mxr576: Avoid calling methods directly on the ApiDoc class/object, access methods via the apidoc storage. `$this->entityTypeStorage->create()`

mxr576 (changes requested):

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
arlina-espinoza: Fixed.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
arlina-espinoza: Implemented the suggested interface. :)

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
arlina-espinoza: Agreed, fixed.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
arlina-espinoza: Agreed, fixed.

modules/apigee_edge_apidocs/tests/src/Kernel/ApidocEntityTest.php
arlina-espinoza: Fixed.

modules/apigee_edge_apidocs/tests/src/Kernel/ApidocEntityRevisionsAccessTest.php
arlina-espinoza: Cool service, I hadn't run into it before. Code has been updated to use `$this->container->get('account_switcher')->switchTo($user);`

modules/apigee_edge_apidocs/tests/src/Kernel/ApidocEntityTest.php
arlina-espinoza: Done.

modules/apigee_edge_apidocs/tests/src/Kernel/ApidocEntityRevisionsAccessTest.php
arlina-espinoza: Fixed!

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: My bad, I have now removed the overriden method.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: Fixed comment.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
arlina-espinoza: Added injecting the entity type manager service in the settings form.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
arlina-espinoza: Fixed, using `$this->messenger()->addStatus()`.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
arlina-espinoza: Fixed.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
arlina-espinoza: Fixed.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
arlina-espinoza: Fixed.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
arlina-espinoza: Fixed.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: In Node, the owner id is the _uid_ field, which we don't have in the ApiDoc entity. And Node's implementation is a workaround to what we are already doing:
From Node:
```
$fields['uid'] = BaseFieldDefinition::create('entity_reference')
     // ...
      ->setDefaultValueCallback('Drupal\node\Entity\Node::getCurrentUserId')
    // ...
```
and 
```
public static function getCurrentUserId() {
  return [\Drupal::currentUser()->id()];
}
```
So I think the current approach works just as well:
```
if (!$this->getRevisionUser()) {
  $this->setRevisionUserId(\Drupal::currentUser()->id());
}
```

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
mxr576: @Jaesin @cnovak Please share your thoughts about this ^ We are still maintaining and testing our own solution instead of relying on an OOTB solution by Entity module - which has its own test coverage.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
mxr576: > In Node, the owner id is the uid field, which we don't have in the ApiDoc entity.

Should not we have an owner on ApiDoc entities? :) This is kinda an OOTB feature on content entities (nodes) that now we are intentionally removing from API doc entities - although it could be useful in the content moderation workflows, ex.: the owner (creator) of an API documentation cannot publish the API documentation only an other evaluated user.

mxr576 (changes requested):

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: The _RevisionRouteProvider_ does provide the route, but the actual URL parameters passed are defined in the core Entity class (https://git.drupalcode.org/project/drupal/blob/8.6.x/core/lib/Drupal/Core/Entity/Entity.php#L314), and that doesn't pass the _apidoc_revision_ parameter for _'revision-revert-form',_ hence the need to add it here. Otherwise it throws this error:

`Symfony\Component\Routing\Exception\MissingMandatoryParametersException: Some mandatory parameters are missing ("apidoc_revision") to generate a URL for route "entity.apidoc.revision_revert_form". in Drupal\Core\Routing\UrlGenerator->doGenerate() (line 182 of /var/www/html/apigee_kickstarter_d8/web/core/lib/Drupal/Core/Routing/UrlGenerator.php).`

modules/apigee_edge_apidocs/src/Entity/ApiDocInterface.php
arlina-espinoza: Yes, added EntityPublishedInterface.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: Yes, that works, I have changed the static method and updated references.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
arlina-espinoza: Fixed.

modules/apigee_edge_apidocs/src/Form/ApiDocSettingsForm.php
arlina-espinoza: Fixed.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
arlina-espinoza: The results are already being cached by Drupal\Core\Entity\EntityAccessControlHandler->access() (https://git.drupalcode.org/project/drupal/blob/8.6.x/core/lib/Drupal/Core/Entity/EntityAccessControlHandler.php#L78)

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: It sounds like a good idea, but not strictly related to this PR. Could we revisit it as another ticket?

mxr576: Since there is no BC guaranteed for experimental modules, we can revisit this later - together with replacing the custom entity access on this entity type with what Entity module provides.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
mxr576: If you extend this class then yes it is cached, but this comment is in connection with my next comment
https://github.com/apigee/apigee-edge-drupal/pull/161/files#r271145040 in which I described that I think you are mixing what an entity access handler and an entity _route_ access handler should do. Consequently, this class should not extend this EntityAccessControlHandler base class.

If you do not believe the Entity (contrib) module, this is how it is implemented for the Node entity in Drupal core:
This is the route access handler for node revision access - which cares about revision elated permissions.
https://github.com/drupal/core/blob/17557c97a816bc78e70989a86064c562bc27027f/modules/node/node.services.yml#L11
https://github.com/drupal/core/blob/17557c97a816bc78e70989a86064c562bc27027f/modules/node/src/Access/NodeRevisionAccessCheck.php#L93-L103
and this is the entity access handler on the Node entity that does not care about revisioning related permissions...
https://github.com/drupal/core/blob/17557c97a816bc78e70989a86064c562bc27027f/modules/node/src/Entity/Node.php#L31
https://github.com/drupal/core/blob/c784c6a8e416505fdbd7f8ab1a6608ad58d37dd9/modules/node/src/NodeAccessControlHandler.php

Also, there is this major difference between entity and route access handlers: https://www.drupal.org/node/2122195#multiple

Please create a follow-up task about cleaning up the access check on the API documentation entity type.

mxr576 (approved):
As @arunz6161 requested I approved this PR, but please create follow up tasks for all remaining, unresolved comments in this PR, especially cleaning up the entity/route access checks on the API documentation entity type.

modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php
mxr576: Yet an other reason why the custom implementation should go:
https://github.com/apigee/apigee-edge-drupal/pull/161/files/901a1e5c817a9447528c55d1cfe1524151f7e7ad#r274771293

Jaesin:
This PR has been moved to the API Catalog module: https://github.com/cnovak/apigee-api-catalog-drupal/pull/7

Issue #72, opened 2018-10-04
Allow ApiProducts entities to be rendered.


Jaesin:
| Q                           | A
| --------------------------- | -----------------------------------------------
| Bug?                        | no
| New Feature?                | yes
| Module Version              | 8.x-1.x
| PHP Client version          | N/A
| Requires PHP client fix     | no

#### Actual Behavior

ApiProduct entities are not renderable because they are not fieldable.

#### Expected Behavior

ApiProduct entities would be renderable.

```
\Drupal::entityTypeManager()
        ->getViewBuilder('api_product')
        ->view($product)
```

#### Steps to Reproduce the behavior

Not a bug.

#### Proposed Solution

Either create a `\Drupal\apigee_edge\Entity\ViewBuilder\EdgeEntityViewBuilder` class that can handle displaying an edge entity without making it fieldable or make the ApiProduct entity fieldable so it can be rendered. The latter is probably a better solution.

mxr576:
Indeed, ideally they should be fielable and they should support view modes where you should be able to change the visibility of base fields (properties) and custom fields (custom attribute representations) for a view mode. Also they should have a list builder - so basically they should support CRUDL in the developer portal - which would also require some additional Edge permissions from the org user used by the portal of course.

In the Drupal 7 devportal this feature has not been supported, there were no default API product listing and viewing UIs so the creation of these UIs were always a custom requirement. We focused mainly on porting default D7 devportal features to Drupal 8 - in a better way where it was possible. This is the reason why API products do not have this capability at this moment. Although, as in Drupal 7, these UIs - if they are needed - can be provided by 3rd party modules. (Listing builder can be added in `hook_entity_type_alter()` and the missing view controller can be defined in a custom module.)

Besides, based on our experiences, the majority of customers does not need these UIs, but if they do they want something really custom. So even if these UIs would exist they would be either disabled and hidden from end users or they would be overridden in custom modules.

If you opened this issue because of the Monetization integration - based on the above described reasons - I think you should/could create the listing builder and view controller for API products in the Monetization integration module, because I assume there are some Monetization (only) information that users may like to see in Drupal about API products. So even if the base apigee_edge module would/will provide these UIs for API products the Monetization (sub)module should/would override these default UIs.

@cnovak What do you think?

Jaesin:
[...]

PR #163, opened 2019-03-25
DRUP-530 Fix error when viewing an API product.

arlina-espinoza:
When rendering an API product (such as setting a view mode other than label on the API docs entity), the system returns the error:

> Error: Call to undefined method Drupal\apigee_edge\Entity\ApiProduct::isDefaultRevision() in Drupal\Core\Entity\EntityViewBuilder->getBuildDefaults() (line 181 of /var/www/html/apigee_kickstarter_d8/web/core/lib/Drupal/Core/Entity/EntityViewBuilder.php) 

This is due to a [core bug](https://www.drupal.org/project/drupal/issues/2951487). This PR adds the fix with a patch to composer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,11 @@
             "services": {
                 "drush.services.yml": "^9"
             }
+        },
+        "patches": {
+            "drupal/core": {
+                "EntityViewBuilder::getBuildDefaults doesn't respect the EntityInterface of the $entity parameter": "https://www.drupal.org/files/issues/2019-01-10/2951487_15_no-tests.patch"
+            }
         }
     }
 }


mxr576:
@arlina-espinoza This bug only occurs when the Monetization module is enabled, isn't it? If this is true, this patch should be required by the Monetization module and not this one.

arlina-espinoza:
@mxr576 No, this was actually reported from the kickstarter when trying to render the api_product on the API docs (from apigee_edge_apidocs module). By default the apidocs module will not show the api_product, but a user might change that on the apidocs entity display settings, so we would have two options:

- Not allowing showing the api_product entity reference field on the API docs - but if a user somehow uses an api_product reference field elsewhere then this bug will show up.
- Fixing it on this module, as the _apigee_edge_ module is the one that defines the api_product entity (`apigee_edge/src/Entity/ApiProduct.php`).

Let me know your thoughts.

mxr576:
Thanks for the clarification @arlina-espinoza, I remembered the same issue with the Monetization module.

I was quite sure that referencing Apigee Edge entities - including API products - works fine with the Apigee Edge. We even have a test case for that. I was partially right about this.

Did a quick check and as I can see the problem that this patch fixes can **only** occur if someone tries to use "Rendered entity" field formatter on an API page (or any content entity with this entity reference field). Of course, this does not work and it is expected to fail because the Apigee Edge does not provide an entity view builder for API products at this point, see #72 for more details.

Manage_display___Apigee_D8_Devportal

Although when "Label" field formatter is being used on the API product field the page works fine without this patch (even if "Link to the referenced entity" is enabled, which is not going to render a link because of the above-mentioned reasons).

Petstore___Apigee_D8_Devportal

* Can you confirm that selecting "Label" or "id" as a field formatter also works for you?
* Do you have a use case why using "Rendered entity" would be necessary?

Based on this I would maybe we should document in the Known issues section of README.md that selecting "Rendered entity" as an option for an API product entity reference field causes WSOD unless a custom module does not define an entity view builder for the API product entity.

arlina-espinoza:
@mxr576:

- _Can you confirm that selecting "Label" or "id" as a field formatter also works for you?_ Yes, the label formatter works fine.
- _Do you have a use case why using "Rendered entity" would be necessary?_ Not out of the box. The "api_product" entity reference field on the API docs is not shown on the display by default. So I guess there are two options:

1. Document this issue on the README, as you mentioned.
2. Or add this core patch.

What do you think?

mxr576:
In general, I would not enforce the installation of any Drupal core patches that do not address a critical issue. A consider an issue a critical one if it breaks a site entirely or only a partially when this module or its submodule is installed. The issue is mitigated by the fact that you have to configure a field formatter in a specific way to reproduce this bug and that is going to cause WSOD only on specific pages (nodes). A simple warning - that I added in #166 - can help externals developer and site-builders to understand how they can avoid to run to this problem.

The problem with Drupal core patches that they complicate the upgrade process between Drupal core releases. (A patch may needs to be rerolled, etc.) Because we suggest customers to allow contrib modules - including this module - to install patches on Drupal core automatically, therefore we should take responsibility and only require those Drupal core patches that are essential for our modules.

@arlina-espinoza if you agree, please close this PR and review mine.

arlina-espinoza:
@mxr576 I agree, your PR looks good and I have reviewed it. Closing this PR.

PR #171, opened 2019-04-03
DRUP-688 Get around core bug by implementing isDefaultRevision on edge entities

arlina-espinoza:
Drupal core assumes entities are revisionable on its EntityViewBuilder. Rather than requiring a core patch, this PR adds an `isDefaultRevision()` method to Edge entities to get around that.

Related:

- https://www.drupal.org/project/drupal/issues/2951487
- https://github.com/apigee/apigee-m10n-drupal/issues/79

**Notes:** when this gets merged, we'll also need to remove the patch from both the Kickstart install profile and apigee_m10n:

- https://github.com/apigee/apigee-m10n-drupal/pull/80
- https://github.com/apigee/apigee-devportal-kickstart-drupal/pull/55

### DIFF
--- a/src/Entity/EdgeEntityBase.php
+++ b/src/Entity/EdgeEntityBase.php
@@ -154,4 +154,24 @@ abstract class EdgeEntityBase extends Entity implements EdgeEntityInterface {
     return $label;
   }
 
+  /**
+   * This is a workaround to avoid a fatal error coming from core.
+   *
+   * Although not needed for Edge Entities, this method is used to get
+   * around core bug, because most of Drupal's core code assumes entities are
+   * either content or config.
+   *
+   * @param bool $new_value
+   *   (optional) This parameter is ignored.
+   *
+   * @return bool
+   *   For Edge Entities this will always return TRUE.
+   *
+   * @see https://www.drupal.org/project/drupal/issues/2951487
+   * @see https://github.com/apigee/apigee-m10n-drupal/issues/79
+   */
+  public function isDefaultRevision($new_value = NULL) {
+    return TRUE;
+  }
+
 }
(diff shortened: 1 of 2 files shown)


mxr576:
EdgeEntityBase is base class that only provides the very basics that are required to turn SDK entities to Drupal entities. It does not care about fields or revisioning and I would like to keep it that way.

FieldableEdgeEntityBase is the class that already contains workarounds for this kind of bugs.
https://github.com/apigee/apigee-edge-drupal/blob/c64b20c5b35774be578adb8a4ef95aa115830d74/src/Entity/FieldableEdgeEntityBase.php#L533-L544
https://github.com/apigee/apigee-edge-drupal/blob/c64b20c5b35774be578adb8a4ef95aa115830d74/src/Entity/FieldableEdgeEntityBase.php#L546-L562

I have already asked around whether we can push this Drupal core fix to the upcoming 8.7.0 release or what is the soonest release when we can get this merged. Having it fixed would be the best possible option for us because all other options are just workarounds. Patching Drupal core is also not a good idea as we discussed earlier.

Based on my findings this problem only affects the MINT module. I was not sure whether MINT entities could extend the FieldableEdgeEntityBase instead of EdgeEntityBase as an instant fix for the problem. So I decoupled and fine-tuned the revisioning "support" from FieldableEdgeEntityBase to a new trait in #172. You can use that add revisioning support to any entities that it is needed. (I would create a new MintEntityBase class in the MINT module that extends the EdgeEntityBase and uses the new trait.)

Please verify the suggested solution.

arlina-espinoza:
Closing per https://github.com/apigee/apigee-edge-drupal/pull/172.

PR #172, opened 2019-04-03
DRUP-688 Revisioning workaround

mxr576:
Contains a workaround for revisioning support that also can be used to address #171 in a better way.

This trait can be used in entity classes where the revisioning "support" is required for some reasons.

Other related PRs/issues from different projects:
* https://github.com/apigee/apigee-m10n-drupal/pull/80
* https://github.com/apigee/apigee-devportal-kickstart-drupal/pull/55

Latest Travis build: https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/515043910

PR #174, opened 2019-04-09
Add RevisioningWorkaroundTrait to ApiProduct

arlina-espinoza:
Adding the `RevisioningWorkaroundTrait` to _ApiProduct_ will fix the issue we were having when someone tried to display a product (other than the label). It will also allow removing the a core patch from the Kickstart profile (https://github.com/apigee/apigee-devportal-kickstart-drupal/pull/55).

Related:

- https://github.com/apigee/apigee-edge-drupal/pull/171
- https://github.com/apigee/apigee-edge-drupal/pull/172

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,8 +72,7 @@ module.**
   to make sure Drupal users and Apigee Edge developers are synchronized.
 * This module does not come with an entity view page for API products at this moment. (See
   [#72](https://github.com/apigee/apigee-edge-drupal/issues/72).) Unless you have a custom module enabled that provides
-  this functionality, do not use the "Rendered entity" field formatter on API product entity reference fields because
-  that could lead to a White Screen of Death (WSOD) error. (Related [issue](http://dgo.to/2951487) on Drupal.org)
+  this functionality, do not use the "Rendered entity" field formatter on API product entity reference fields.
 
 ## Requirements
 
--- a/src/Entity/ApiProduct.php
+++ b/src/Entity/ApiProduct.php
@@ -44,6 +44,12 @@ use Apigee\Edge\Structure\AttributesProperty;
  */
 class ApiProduct extends EdgeEntityBase implements ApiProductInterface {
 
+  // The majority of Drupal core & contrib assumes that an entity to be
+  // displayed is a content entity, and because it is a content entity it also
+  // must support revisioning. This incorrect assumption justifies the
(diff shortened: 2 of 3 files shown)


mxr576:
@arlina-espinoza I understand that adding this a workaround makes could make sense, but I thought we had a conclusion (which we documented in the README.md) that nobody should try to display anything else than the "ID" or the "Label" of an API product because it does not have an entity view builder, neither defined routes.

So I am still wondering, what do you do in the Kickstarter or the MINT module why you have this issue still? Can you guide me to some code that explains why you are having this problem still?

mxr576:
@arlina-espinoza Also, as a general request, please start adding Travis CI build links to your PR that confirms your changes does not break anything in 8.x-1.x after it gets merged.

For example https://github.com/apigee/apigee-edge-drupal/pull/172#issue-266915678

It is writting in the CONTRIBUTING.md how you can setup Travis: https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/CONTRIBUTING.md#to-start

Jaesin:
If there is an entity reference to an API Product, how can you ensure it's not configured to render as an entity? Why not allow entities to be rendered?

mxr576:
Do not expose it as a field that display/form display mode can be configured :) 
image

Or do not expose the data as a base field, expose it as a pseudo field.
https://github.com/apigee/apigee-edge-drupal/blob/2fdefd74f8ae75d2003a15b6652301023a3c1b14/apigee_edge.module#L216-L230
https://github.com/apigee/apigee-edge-drupal/blob/2fdefd74f8ae75d2003a15b6652301023a3c1b14/apigee_edge.module#L246-L276

arlina-espinoza:
We are trying to get rid of all core patches in the apigee modules and kickstart. I think that https://github.com/apigee/apigee-edge-drupal/pull/172 fixes all current issues on the edge and monetization modules, and kickstart profile, but it would be nice to have this in case someones adds an _api_product_ reference field somewhere else we don't have control over and it doesn't break their site, while still leaving it in the README that this module doesn't support any display modes for API product entities other than the label.

Jaesin (approved):
This allow the tests to pass for `apigee_m10n` without the need for a patch to Drupal core.

Tested against https://github.com/apigee/apigee-m10n-drupal/pull/80

PR #175, opened 2019-04-11
GH: Identify Drupal specific file types as PHP files

mxr576:
At this moment there is no syntax-highlighting on Drupal specific files. This small tweak ensures GH identifies these files as PHP files.

After this gets committed the syntax highlighting may not work until someone modifies a Drupal-specific file: https://github.com/github/linguist/issues/4491#issuecomment-481680110

Before:
image

After:
image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,3 +7,9 @@ CONTRIBUTING.md                 export-ignore
 .github                         export-ignore
 .travis                         export-ignore
 tests                           export-ignore
+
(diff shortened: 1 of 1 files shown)


cnovak (approved):
Good idea, do you want to add this to our m1on repo and kickstart as well, or should we add it as a TODO?

mxr576:
@cnovak Thanks, I'll create a PR with this later today for those projects too.

Issue #178, opened 2019-04-15
DRUP-735 apigee_auth key does not validate credentials


shadcn:
| Q                           | A
| --------------------------- | -----------------------------------------------
| Bug?                        | yes
| New Feature?                | no
| Module Version              | `8.x-1.x`
| PHP Client version          | `2.0.1`
| Requires PHP client fix     | no

#### Actual Behavior

Credentials submitted via the _admin/config/system/keys/manage/apigee_edge_connection_default_ are not validated against Apigee connection.

#### Expected Behavior

The connection for credentials provided should be validated.

#### Steps to Reproduce the behavior

1. Edit the auth key at _admin/config/system/keys/manage/apigee_edge_connection_default_
2. Select Configuration as the key provider.
3. Enter any random values.
4. Save.

The random values are saved.

#### Proposed Solution

Extend `\Drupal\apigee_edge\Plugin\KeyType\ApigeeAuthKeyType::validateKeyValue` to validate values provided for the key. Maybe extract `\Drupal\apigee_edge\Form\AuthenticationForm::validateForm` somewhere and reuse?

mxr576:
@arshad This works are expected, only Apigee prefixed key providers can do validation on the Key form because they are defined by the Apigee Edge module. Others can only validate on the Authentication form, which is the main configuration page for Apigee Edge authentication set up.

So you can change the Key provider on the Key configuration form to anything but then you have to go back the Authentication form to Send a request if the Key provider is not an Apigee prefixed one.

shadcn:
@mxr576 Sorry I don't follow.

`apigee_auth` is shipped with the Apigee Edge module. Can't we check if the values provided are valid credentials in `\Drupal\apigee_edge\Plugin\KeyType\ApigeeAuthKeyType::validateKeyValue`?

We are already doing some missing and empty field validations in there.

mxr576:
Yeah, I did some digging as well since my comment and I'll look into this whether it is possible to do that there. My gut says Peter must have had a reason for not doing this validation there but we'll see.

mxr576:
Okay, so probably the reason why Peter did not use this `validateKeyValue()` because it does not get called on the Authentication form. The reason for that is how the Key module's entity edit form got "included" in the Authentication form. Basically, it is a copy-paste from the Key module and it tries to replicate what Key module does when it processes a form and saves the submitted values. Calling validateKeyValue() is missing from the replicated behavior.

[...]

Issue #176, opened 2019-04-15
DRUP-734 \Drupal\apigee_edge\Form\AuthenticationForm does not update key on submit


shadcn:
| Q                           | A
| --------------------------- | -----------------------------------------------
| Bug?                        | yes
| New Feature?                | no
| Module Version              | `8.x-1.x`
| PHP Client version          | `2.0.1`
| Requires PHP client fix     | no

#### Actual Behavior

On submit, the `\Drupal\apigee_edge\Form\AuthenticationForm` does not update the active key.

#### Expected Behavior

The active key should be updated when the `\Drupal\apigee_edge\Form\AuthenticationForm` form is submitted.

#### Steps to Reproduce the behavior

1. Enable Apigee Edge.
2. Visit _/admin/config/system/keys_ and edit the default key.
3. Select *Configuration* as the key provider.
4. Add credentials and save. Credentials are saved successfully.
5. Visit _/admin/config/apigee-edge/settings_
6. Enter credentials and save.
7. Credentials are *NOT* updated.

#### Proposed Solution

Replace

https://github.com/apigee/apigee-edge-drupal/blob/9e753e8f8f390489dea2e2ec38571a13563bdb5f/src/Form/AuthenticationForm.php#L361-L364

With

```
$this->activeKey->setKeyValue($processed_submitted);
$this->activeKey->save();
```

`$this->activeKey->save()` will call the `keyProvider->setKeyValue` automatically. See `\Drupal\key\Entity\Key::preSave`

mxr576:
@arshad If you already have an idea for the fix, please create a PR. This issue surprises me because authentication form has not been touched since a couple of sprints.

mxr576:
Also, if this true it means we are missing something in the authentication form tests so please provide a fix or an additional test case that covers this issue.

mxr576:
PR merged.

PR #179, opened 2019-04-17
DRUP-734 & DRUP-735 Fix authentication form issues

mxr576:
Everything is working as it worked, except the "Save" and "Save configurations" buttons are not disabled on the Authentication and Key forms with States API as they were before. It would be too complicated to make it work with on the Key forms:

```
    // Disabling the Save button with States API would only work on Key edit
    // form and on Key add form after is submitted (and failed with validation
    // error). The reason is simple, the AJAXified select form elements only
    // rebuilds elements under $form['settings'].
    // Also we must not disable the Save button on not Apigee Edge related
    // key types, therefore, this feature is not implemented.
```
Also, it was not a really useful feature because HTML5 client-side validation handles required fields properly and even if someone disables it with a trick, the validation on the backend can also handle this.

**New features:**

Key provider finally can be changed on the Authentication form, although the default is still the private key provider.
new_auth_form

Key add/edit forms now validate the provided credentials if the key type is Apigee Edge Authentication. This alteration of the Key form can be disabled with a config.

validation_on_key_form

Latest Travis CI build: https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/521831379

Related issues: #176 #178


src/Form/AuthenticationForm.php
```diff
@@ -136,622 +105,88 @@ public function getFormId() {
...
-   * @param \Drupal\Core\Form\FormStateInterface $to
-   *   The form state object to move to.
-   */
-  protected function moveFormStateStorage(FormStateInterface $from, FormStateInterface $to) {
-    foreach ($from->getStorage() as $index => $value) {
-      $to->set($index, $value);
-    }
-  }
+    $form['settings']['provider_section']['#open'] = $key_provider_requirements_error || ($form_state->getTriggeringElement() && $form_state->getTriggeringElement()['#name'] === 'key_provider');
+    $form['settings']['provider_section']['#weight'] = $form['settings']['input_section']['#weight'] + 1;
```
shadcn, line 164: Is there a reason we are moving this under the `input_section`?

shadcn (changes requested):
This looks good. 

✅ I can confirm validation on both the authentication form and the key form. 
✅ This also fixes the issue with authentication form not saving the key.
❌ I found another issue though. You will need two sets of valid credentials to reproduce this. See below:

1. Fill in the authentication form with valid credentials for Org A and save.
2. Now edit and fill in valid credentials for Org B. Do not save.
3. Click Send request to test connection. You should see "Connection successful".
4. Save
5. You will see a connection error because the password field now holds the password from the previous Org.

src/Form/AuthenticationForm.php
mxr576: Apigee did not want to expose the Key provider selector on this form originally to keep it simple and speed up the first-time configuration for those who just would like to try to out the module. Although this implementation hid the fact that the Authentication form saves the Apigee Edge credentials by default to an unencrypted file, even if it is in the private file system. A different Key provider than private filesystem could have been configured only on the Key entity's edit form, which could have been found only by those ppl who know the module or read its documentation. I always felt this is a bad idea because ppl do not read documentations and they will use the private key provider in production sites as well.

Now the current UI combines the best from both. The credential configuration form still has the primary focus on the top of the page but the key provider selector is also visible, only in the bottom of the page. Any user can easily find it, they do not need to navigate to a different UI to select more secure credential storage.

mxr576:
@arshad Good catch! I managed to reproduce the issue, I already have a fix. I'll test is a little bit more tomorrow before I update this PR.

src/Form/AuthenticationForm.php
cnovak: I like the idea of getting the “best from both”!  Good solution to have advanced settings, this makes it more clear to the user that there are other options and to understand the tie to the Key module in a more clear way.

mxr576:
@arshad Implemented the fix, tested it manually several times. The previously problematic workflow that you described could be covered with an automated test only if
* we have two sets of valid Apigee Edge credentials
* we can return mock HTTP responses in the SDK connector - as I do it in the PHP API client - this is not yet implemented so I created a follow-up task for this: DRUP-747

shadcn (approved):
This looks good too. Confirmed it fixes all issues reported. Thanks.

tests/src/Traits/ApigeeEdgeFunctionalTestTrait.php
```diff
@@ -31,26 +31,24 @@
 /**
  * Provides common functionality for the Apigee Edge test classes.
  *
- * TODO Consider renaming this trait to ApigeeEdgeFunctionalTestTrait because
- * its setup() method is clearly works only in functional tests. Re-usable
- * methods in kernel/functional tests can be remain in this trait, ex.:
- * getRandomUniqueId(), logException(), etc.
+ * TODO Move methods from this trait that can be also used in kernel tests to a
```
Jaesin, line 34: Is there an issue to fix this?

mxr576: No, It is a low priority issue because it is in tests so I have created a ticket for that. I usually hunt down these kinds of issues when I need to modify tests.
Also, instead of a dedicated task for this TODO, I'd rather create a "hunt down TODOs" task because I guess there are more than one in the module.

PR #181, opened 2019-04-23
DRUP-687 Added default Apigee entity theme

### LINKED ISSUE #169: DRUP-687 Prefix CSS selectors in Apigee Edge
| Module Version              | 8.x-1.x
| PHP Client version          | 2.0.1

#### Actual Behavior

The Apigee Edge module adds CSS styles that are too general. Example:

https://github.com/apigee/apigee-edge-drupal/blob/ffeedfe54e7950260507c25e2ed50a36b0de2d98/css/apigee_edge.app_view.css#L18-L33

This breaks the layout of field on other pages (where the `apigee_edge.app_view` library is present) and makes it harder for themes to override.

#### Proposed Solution

Make this module/element-specific by using a selector prefix. Example: `.apigee-foo .field`.

mxr576:
New default HTML markup for Apigee Edge entities:

Developer app:
image

Team app:
image

Team:
image

Latest Travis build: https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/527644023

Resolves: #169
Related Drupal core issue: https://www.drupal.org/project/drupal/issues/2808481

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -120,9 +120,7 @@ function apigee_edge_theme() {
     ],
     'apigee_entity__app' => [
       'render element' => 'elements',
-      // DRY.
       'base hook' => 'apigee_entity',
-      'template' => 'apigee-entity',
     ],
     'app_credential' => [
       'render element' => 'elements',
@@ -176,21 +174,6 @@ function template_preprocess_apigee_entity(array &$variables) {
   foreach (Element::children($variables['elements']) as $key) {
     $variables['content'][$key] = $variables['elements'][$key];
   }
-}
-
-/**
- * Prepares variables for Apigee App entity templates.
- *
- * Base hook: apigee_entity
- * Default template: apigee-entity.html.twig.
- *
- * @param array $variables
- *   An associative array containing:
- *   - elements: An array of elements to display in view mode.
- */
-function template_preprocess_apigee_entity__app(array &$variables) {
-  $variables['attributes']['class'][] = 'apigee-entity--app';
-  $variables['attributes']['class'][] = 'apigee-entity--app--view-mode-' . $variables['elements']['#view_mode'];
 }
 
 /**
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -187,11 +187,11 @@ function apigee_edge_theme_suggestions_apigee_entity(array $variables) {
 
   if ($entity instanceof AppInterface) {
     $suggestions[] = 'apigee_entity__app';
-    $suggestions[] = 'apigee_entity__app__view_mode__' . $sanitized_view_mode;
+    $suggestions[] = 'apigee_entity__app__' . $sanitized_view_mode;
   }
 
   $suggestions[] = 'apigee_entity__' . $entity->getEntityTypeId();
-  $suggestions[] = 'apigee_entity__' . $entity->getEntityTypeId() . '__view_mode__' . $sanitized_view_mode;
+  $suggestions[] = 'apigee_entity__' . $entity->getEntityTypeId() . '__' . $sanitized_view_mode;
 
   return $suggestions;
 }
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -49,6 +49,7 @@ use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Field\BaseFieldDefinition;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Link;
+use Drupal\Core\Render\Element;
 use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Site\Settings;
@@ -114,6 +115,15 @@ function apigee_edge_form_key_form_alter(array &$form, FormStateInterface $form_
  */
 function apigee_edge_theme() {
   return [
+    'apigee_entity' => [
+      'render element' => 'elements',
+    ],
+    'apigee_entity__app' => [
+      'render element' => 'elements',
+      // DRY.
+      'base hook' => 'apigee_entity',
+      'template' => 'apigee-entity',
+    ],
     'app_credential' => [
       'render element' => 'elements',
     ],
@@ -139,6 +149,68 @@ function apigee_edge_user_format_name_alter(&$name, AccountInterface $account) {
       $name = $first_name_last_name;
     }
   }
+}
+
+/**
+ * Prepares variables for Apigee entity templates.
+ *
+ * Default template: apigee-entity.html.twig.
+ *
+ * @param array $variables
+ *   An associative array containing:
+ *   - elements: An array of elements to display in view mode.
+ */
+function template_preprocess_apigee_entity(array &$variables) {
+  $variables['view_mode'] = $variables['elements']['#view_mode'];
+  /** @var \Drupal\apigee_edge\Entity\EdgeEntityInterface $entity */
+  $entity = $variables['entity'] = $variables['elements']['#entity'];
+
+  $variables['label'] = $entity->label();
+
+  if (!$entity->isNew() && $entity->hasLinkTemplate('canonical')) {
+    $variables['url'] = $entity->toUrl('canonical', ['language' => $entity->language()])->toString();
+  }
+
+  // Helpful $content variable for templates.
+  $variables += ['content' => []];
+  foreach (Element::children($variables['elements']) as $key) {
+    $variables['content'][$key] = $variables['elements'][$key];
+  }
+}
+
+/**
+ * Prepares variables for Apigee App entity templates.
+ *
+ * Base hook: apigee_entity
+ * Default template: apigee-entity.html.twig.
+ *
+ * @param array $variables
+ *   An associative array containing:
+ *   - elements: An array of elements to display in view mode.
+ */
+function template_preprocess_apigee_entity__app(array &$variables) {
+  $variables['attributes']['class'][] = 'apigee-entity--app';
+  $variables['attributes']['class'][] = 'apigee-entity--app--view-mode-' . $variables['elements']['#view_mode'];
+}
+
+/**
+ * Implements hook_theme_suggestions_HOOK().
+ */
+function apigee_edge_theme_suggestions_apigee_entity(array $variables) {
+  $suggestions = [];
+  /** @var \Drupal\apigee_edge\Entity\EdgeEntityInterface $entity */
+  $entity = $variables['elements']['#entity'];
+  $sanitized_view_mode = str_replace('.', '_', $variables['elements']['#view_mode']);
+
+  if ($entity instanceof AppInterface) {
+    $suggestions[] = 'apigee_entity__app';
+    $suggestions[] = 'apigee_entity__app__view_mode__' . $sanitized_view_mode;
+  }
+
+  $suggestions[] = 'apigee_entity__' . $entity->getEntityTypeId();
+  $suggestions[] = 'apigee_entity__' . $entity->getEntityTypeId() . '__view_mode__' . $sanitized_view_mode;
+
+  return $suggestions;
 }
 
 /**
--- a/css/apigee_edge.app_view.css
+++ b/css/apigee_edge.app_view.css
@@ -15,13 +15,13 @@
  * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
-.field {
+.apigee-entity--app--view-mode-full .field {
   width: 48%;
   display: inline-block;
   vertical-align: top;
 }
 
-.field--label {
+.apigee-entity--app--view-mode-full .field--label {
   margin-bottom: 1rem;
   vertical-align: middle;
   font-weight: bold;
@@ -32,41 +32,37 @@
   min-width: 160px;
 }
 
-.panel-body .field .field--label::after {
-  content: '';
-}
-
-.wrapper--primary,
-.wrapper--secondary {
+.apigee-entity--app--view-mode-full .wrapper--primary,
+.apigee-entity--app--view-mode-full .wrapper--secondary {
   width: 48%;
   display: inline-block;
   vertical-align: top;
 }
 
-.wrapper--secondary {
+.apigee-entity--app--view-mode-full .wrapper--secondary {
   margin-left: 2%;
 }
 
-.wrapper--secondary .api-product-list-row {
+.apigee-entity--app--view-mode-full .wrapper--secondary .api-product-list-row {
   max-width: 300px;
   margin-bottom: 1rem;
 }
 
-.api-product-list-row .api-product-name {
+.apigee-entity--app--view-mode-full .api-product-list-row .api-product-name {
   float: left;
 }
 
-.api-product-list-row .wrapper--status {
+.apigee-entity--app--view-mode-full .api-product-list-row .wrapper--status {
   float: right;
 }
 
-.secret {
+.apigee-entity--app--view-mode-full .secret {
   display: none;
   position: relative;
   z-index: auto;
 }
 
-.secret:hover {
+.apigee-entity--app--view-mode-full .secret:hover {
   z-index: 999;
   background-color: rgba(255,255,255,0.8);
 }
--- a/src/Entity/EdgeEntityViewBuilder.php
+++ b/src/Entity/EdgeEntityViewBuilder.php
@@ -1,0 +1,56 @@
+<?php
+
+/**
+ * Copyright 2018 Google Inc.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * version 2 as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+namespace Drupal\apigee_edge\Entity;
+
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityViewBuilder;
+
+/**
+ * Default view builder class for Apigee Edge entities.
+ */
+class EdgeEntityViewBuilder extends EntityViewBuilder {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getBuildDefaults(EntityInterface $entity, $view_mode): array {
+    $defaults = parent::getBuildDefaults($entity, $view_mode);
+
+    // Simplified path to the entity.
+    $defaults['#entity'] = $entity;
+
+    // Use the generic apigee_entity theme if the entity type specific theme
+    // does not exist. Drupal core does not provide a default template for
+    // entities yet. This solution also allows to change the default, non-entity
+    // type specific theme that is being used if the entity type specific does
+    // not exist.
+    // When 2808481 gets in core, the default theme becomes the "entity".
+    // Because of BC we can not use that.
+    // @see apigee-entity.html.twig
+    // @see https://www.drupal.org/project/drupal/issues/2808481
+    if (!isset($defaults['#theme']) || $defaults['#theme'] === 'entity') {
+      $defaults['#theme'] = 'apigee_entity';
+    }
+
+    return $defaults;
+  }
+
+}
--- a/templates/apigee-entity--app.html.twig
+++ b/templates/apigee-entity--app.html.twig
@@ -1,13 +1,8 @@
 {#
 /**
  * @file
- * Generic theme implementation to display an Apigee entity.
- *
- * Drupal core does not yet provide a generic theme for entities therefore you
- * have to implement your own entity type specific theme. This will change
- * when http://dgo.to/2808481 gets fixed which will introduce a generic template
- * for entities with common preprocess functions and theme suggestions.
- *
+ * Generic theme implementation to display a developer- or team app entity.
+
  * Available variables:
  * - entity: The entity with limited access to object properties and methods.
  *   Only method names starting with "get", "has", or "is" and a few common
@@ -45,31 +40,11 @@
  */
 #}
 
-{%
-  set classes = [
-  'apigee-entity',
-  view_mode ? 'apigee-entity--view-mode-' ~ view_mode|clean_class,
-  entity.getEntityTypeId|clean_class,
-  view_mode ? entity.getEntityTypeId|clean_class ~ '--view-mode-' ~ view_mode|clean_class,
-]|merge(classes|default([]))
-%}
+{% if entity.getEntityTypeId in ['developer_app', 'team_app'] %}
+  {% set classes = [
+    'apigee-entity--app',
+    'apigee-entity--app--view-mode-' ~ view_mode|clean_class
+  ] %}
+{% endif %}
 
-{% block apigee_entity %}
-  <article role="article"{{ attributes|without('role').addClass(classes) }}>
-
-    {% block content %}
-      {{ title_prefix }}
-      {% if label and view_mode != 'full' %}
-        <h2{{ title_attributes }}>
-          {{ label }}
-        </h2>
-      {% endif %}
-      {{ title_suffix }}
-
-      <div{{ content_attributes }}>
-        {{ content }}
-      </div>
-    {% endblock %}
-
-  </article>
-{% endblock %}
+{% include 'apigee-entity.html.twig' %}
--- a/templates/apigee-entity.html.twig
+++ b/templates/apigee-entity.html.twig
@@ -1,0 +1,68 @@
+{#
+/**
+ * @file
+ * Generic theme implementation to display an Apigee entity.
+ *
+ * Drupal core does not yet provide a generic theme for entities therefore you
+ * have to implement your own entity type specific theme. This will change
+ * when http://dgo.to/2808481 gets fixed which will introduce a generic template
+ * for entities with common preprocess functions and theme suggestions.
+ *
+ * Available variables:
+ * - entity: The entity with limited access to object properties and methods.
+ *   Only method names starting with "get", "has", or "is" and a few common
+ *   methods such as "id", "label", and "bundle" are available. For example:
+ *   - entity.getEntityTypeId() will return the entity type ID.
+ *   - entity.hasField('field_example') returns TRUE if the entity includes
+ *     field_example. (This does not indicate the presence of a value in this
+ *     field.)
+ *   Calling other methods, such as entity.delete(), will result in an exception.
+ *   See \Drupal\apigee_edge\Entity\EdgeEntityInterface for a full list of
+ *   methods.
+ * - label: The title of the entity.
+ * - content: All rendered field items. Use {{ content }} to print them all,
+ *   or print a subset such as {{ content.field_example }}. Use
+ *   {{ content|without('field_example') }} to temporarily suppress the printing
+ *   of a given child element.
+ * - url: Direct URL of the current entity.
+ * - attributes: HTML attributes for the containing element.
+ * - title_attributes: Same as attributes, except applied to the main title
+ *   tag that appears in the template.
+ * - content_attributes: Same as attributes, except applied to the main
+ *   content tag that appears in the template.
+ * - title_prefix: Additional output populated by modules, intended to be
+ *   displayed in front of the main title tag that appears in the template.
+ * - title_suffix: Additional output populated by modules, intended to be
+ *   displayed after the main title tag that appears in the template.
+ * - view_mode: View mode; for example, "teaser" or "full".
+ *
+ * @see \Drupal\apigee_edge\Entity\EdgeEntityViewBuilder::getBuildDefaults()
+ * @see \Drupal\Core\Entity\EntityViewBuilder::getBuildDefaults()
+ * @see template_preprocess_apigee_entity()
+ * @see https://www.drupal.org/project/drupal/issues/2808481
+ *
+ * @ingroup themeable
+ */
+#}
+{% set classes = attributes['class']|merge([
+  'apigee-entity',
+  view_mode ? 'apigee-entity--view-mode-' ~ view_mode|clean_class,
+  entity.getEntityTypeId|clean_class,
+  view_mode ? entity.getEntityTypeId|clean_class ~ '--view-mode-' ~ view_mode|clean_class,
+]) %}
+
+<article role="article" {{ attributes|without('role').addClass(classes) }}>
+
+  {{ title_prefix }}
+  {% if label and view_mode != 'full' %}
+    <h2{{ title_attributes }}>
+      {{ label }}
+    </h2>
+  {% endif %}
+  {{ title_suffix }}
+
+  <div{{ content_attributes }}>
+    {{ content }}
+  </div>
+
+</article>
--- a/css/apigee_edge.app_listing.css
+++ b/css/apigee_edge.app_listing.css
@@ -15,9 +15,8 @@
  * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
-
 /* App listing components */
-.circle {
+.table--app-list .circle {
   border-radius: 50%;
   color: white;
   width: 14px;
@@ -30,22 +29,16 @@
   float: left;
 }
 
-.wrapper--status {
+.table--app-list .wrapper--status {
   float: left;
 }
 
-.row--warning {
+.table--app-list .row--warning {
   display: none;
 }
 
-/* Temporary zebra healing. */
-.table--app-list .odd,
-.table--app-list .even {
-  /* background-color: transparent; */
-}
-
 /* Collapsible functionality. */
-.toggle--warning {
+.table--app-list .toggle--warning {
   display: inline-block;
   overflow: hidden;
   border-bottom: none;
@@ -55,21 +48,13 @@
   font-size: 85%;
 }
 
-.toggle--warning:hover,
-.toggle--warning:active,
-.toggle--warning:visited,
-.toggle--warning:focus {
+.table--app-list .toggle--warning:hover,
+.table--app-list .toggle--warning:active,
+.table--app-list .toggle--warning:visited,
+.table--app-list .toggle--warning:focus {
   border-bottom: none;
 }
 
-.toggle--warning::after {
-
-}
-
-.toggle--warning.closed::after {
-
-}
-
-.toggle--warning.open span {
+.table--app-list .toggle--warning.open span {
   background-position: -65px -16px;
 }
--- a/src/Annotation/EdgeEntityType.php
+++ b/src/Annotation/EdgeEntityType.php
@@ -28,7 +28,8 @@ use Drupal\Core\StringTranslation\TranslatableMarkup;
  *
  * The annotation properties of entity types are found on
  * \Drupal\apigee_edge\Entity\EdgeEntityType and are accessed using
- * get/set methods defined in \Drupal\Core\Entity\EntityTypeInterface.
+ * get/set methods defined in
+ * \Drupal\apigee_edge\Entity\EdgeEntityTypeInterface.
  *
  * @Annotation
  */
--- a/src/Entity/AppViewBuilder.php
+++ b/src/Entity/AppViewBuilder.php
@@ -20,14 +20,13 @@
 
 namespace Drupal\apigee_edge\Entity;
 
-use Drupal\Core\Entity\EntityViewBuilder;
 use Drupal\Core\Render\Element;
 use Drupal\Core\Url;
 
 /**
  * Common app view builder for developer- and company (team) apps.
  */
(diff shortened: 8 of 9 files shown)


templates/apigee-entity-default.html.twig
```diff
@@ -0,0 +1,80 @@
...
+ * @see template_preprocess_apigee_entity_default()
+ *
+ * @ingroup themeable
+ */
+#}
+<article{{ attributes }}>
+
+  {{ title_prefix }}
+  {% if label and view_mode != 'full' %}
+    <h2{{ title_attributes }}>
```
kittiradovics: Can we make the element configurable from preprocess instead of hardcoding `<h2>`?

apigee_edge.module
```diff
@@ -108,6 +112,41 @@ function apigee_edge_user_format_name_alter(&$name, AccountInterface $account) {
   }
 }
 
+/**
+ * Prepares variables for Apigee entity templates.
+ *
+ * Default template: apigee-entity.html.twig.
```
kittiradovics, line 155: I can only see `apigee-entity-default.html.twig` template but I would prefer simply `apigee-entity.html.twig`.

css/apigee_edge.app_view.css
```diff
@@ -15,13 +15,13 @@
  * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
-.field {
+[class*="-app--view-mode-full"] .field {
```
kittiradovics: Wouldn't be better to add a common class to developer and teams apps? E.g. `apigee-entity--app` or something like that. That way we can avoid using attribute selectors in the future.

kittiradovics (changes requested):

apigee_edge.module
```diff
@@ -81,6 +82,9 @@ function apigee_edge_module_implements_alter(&$implementations, $hook) {
  */
 function apigee_edge_theme() {
   return [
+    'apigee_entity_default' => [
```
shadcn: Having one template for all entity types is weird. 

I see what you're doing here but I think we should have separate templates for entity types. 

Example: `developer_app`,  `team_app`, `api_product`.

apigee_edge.module
shadcn: If we are going to keep one template for all edge entities, yeah let's go with `apigee-entity.html.twig`.

css/apigee_edge.app_view.css
shadcn: @mxr576 We are doing something similar in Kickstart. See https://github.com/apigee/apigee-devportal-kickstart-drupal/blob/8.x-1.x/modules/custom/apigee_kickstart_enhancement/templates/app.html.twig

mxr576: I did not want to create a dedicated app (developer/team app) theme and template in the module, I wanted to keep it simple. Every Apigee entity uses the same ~default~ fallback Apigee entity template and if you would like to use a custom template, you just need to define a theme function with the name of the entity type - as the header in the apigee-entity-default.html.twig describes.

The header also describes that it is an intended behavior that there are entity type specific template suggestions for the apigee_entity_default theme. 

https://github.com/apigee/apigee-edge-drupal/blob/02a29d6f5082306bfa319eca57c5c157962ead1a/templates/apigee-entity-default.html.twig#L9-L31

Thanks for this approach we need to worry less about BC breaking issues in the theme layer. If someone starts customizing the Apigee entity view pages it starts with a clean sheet and it maintains its own customization.

templates/apigee-entity-default.html.twig
mxr576: This is a bare minimum, default template based on the node.html.twig. If you would like to make it customizable, create a new entity type specific theme and ENTITY_TYPE.htm.yml in your theme/module. :)

apigee_edge.module
mxr576: The "default" suffix only tries to emphasize what I described in my next comment and in the header of the apigee-entity-default.html.twig. An entity type specific theme should not be implemented by adding entity type specific CSS, classes and template suggestions to the "apigee_entity_default" theme. It should be achieved by defining a theme for the entity_type. The "apigee_entity_default" theme is only used as a fallback solution if a theme/profile/module would not define an entity type specific theme.

So this is not "one template for all edge entities", it is a fallback template for Apigee entities that do not have their own template defined. It just ensures that we write more specific CSS in the Apigee Edge module than without this.

mxr576: Should I rename it to `apigee_edge_fallback`? :)

apigee_edge.module
mxr576: See my comment above.
If you still disagree please describe what we would win by defining the entity type specific themes/temples with the same content? (It would mean that the MINT module also must define its own entity type specific themes and templates as well instead of using the EdgeEntityViewBuilder as a view builder and automatically using the fallback theme "apigee_edge_default" if no customization is needed in the HTML markup.)

css/apigee_edge.app_view.css
mxr576: @arshad What you referenced from the kickstarter_enhancements module is what I am actually trying to achieve here. Modules and themes that would like to customize a display of an Apigee Edge entity view should implement and maintain its own theme.

Edit:
And this is also a good example. In a custom module, if you need more control over a layout, you can replace the view/list builder of an entity and do whatever you just want. This does not cause any extra maintenance cost for the maintainers of the Apigee Edge module :+1: .

https://github.com/apigee/apigee-devportal-kickstart-drupal/blob/8.x-1.x/modules/custom/apigee_kickstart_enhancement/src/Entity/AppViewBuilder.php#L37-L41

apigee_edge.module
jacine: Can we please KISS and implement each entity with its own named hook, and template? It would help toward solving a lot of issues happening in this PR, and make the end result easier for anyone working with it to understand. 

The first thing someone looking to customize one of these templates will do is look for the a template named after the entity. Let's just give them that.  It's well worth having multiple templates, even if they all have the same, basic markup because this is about enabling the _ability_ to customize these entities just like anyone doing theme development would for any other entity: node, media, paragraph, or otherwise. Anything else is an anti-pattern that is bound to confuse people. 

Another negative side effect of doing it this way is `apigee_entity_default` is the hook that whole pipeline is affected (preprocess, suggestions, etc), and will have to be dealt with through this prefix, as noted in the comments of the template.  That's a terrible TX to put onto theme developers, and we shouldn't do it.

templates/apigee-entity-default.html.twig
```diff
@@ -0,0 +1,80 @@
...
+ *
+ *   function template_preprocess_developer_app(&$variables) {};
+ *
+ *   function MY_MODULE_theme_suggestions_developer_app(array $variables) {}
+ * @endcode
+ *
+ * If you would like to re-use the implementation from
+ * template_preprocess_apigee_entity_default() you can call this function in
+ * your own entity type specific template preprocess function or add it to the
+ * entity type specific preprocess functions in hook_theme_registry_alter().
```
jacine: Everything from line 6-31 should not be here. We should not have to educate on how to create theme hooks, base preprocess hooks, or have class references, especially not in a Twig template, you know?  This is not worth the perceived gain of having one vs multiple templates IMO.

apigee_edge.module
cnovak: If multiple reviewers would expect the name to be `apigee-entity.html.twig`, maybe it would be worth using it since it would reduce confusion.

apigee_edge.module
cnovak: I do agree with @jacine on the fact that our goal is to make things simple for developer that need to make customizations even if we have some basic template markup repetition.  Let’s error on the side of ease of use than worrying about a bit of replication in templates

templates/apigee-entity-default.html.twig
cnovak: I always appreciate more documentation than less, but that does seem like a problem if we need to explain how our patterns work versus already established conventions.

apigee_edge.module
kittiradovics: I vote for `apigee-entity.html.twig` too. I think it's obvious that it's the default template without adding the `-default` suffix.

templates/apigee-entity-default.html.twig
kittiradovics: We constantly modify the node template for teaser mode because `<h2>` is almost never good. That's why I suggested to make it configurable from preprocess, which is less code than adding a new template. But I'm also ok with the original solution if you want something similar to `node.html.twig`.

css/apigee_edge.app_view.css
kittiradovics: I didn't say that you should create dedicated templates, just add a common class to both entities like `apigee-entity--app` to make frontenders' life easier :)

templates/apigee-entity-default.html.twig
mxr576: @jacine I felt necessary to educate developers who are going to customize the look and feel of the Apigee entities because if the fix for the https://www.drupal.org/project/drupal/issues/2808481 issue lands in Drupal core it will create a confusing situation.

There will be multiple themes and templates that could be used to theme an entity, for example, a developer app and you must know the order how Drupal will try to apply them:
* If you define the `developer_app` theme in `hook_theme()` on your module, this is going to be the primary theme for the developer_app entity and it always gets chosen by the EntityViewBuilder. The related template that you use is `developer-app.html.twig`. (Just like now)
See:
https://github.com/drupal/core/blob/17557c97a816bc78e70989a86064c562bc27027f/lib/Drupal/Core/Entity/EntityViewBuilder.php#L174-L177
* If that theme does not exist, Drupal will try to apply the default "entity" theme and because of that a developer app's view could be customized`entity--developer-app.twig.html` template.

This is two theme and multiple templates already that you can possibly use to customize one entity's view.

I did not want to create another with the "apigee_entity_default" theme. It is just a fallback solution to fix a Drupal issue that causing problems for us and it seems it won't be fixed till Drupal 8.8. This is the reason why I am trying to discourage everybody to try to implement something like this: `apigee-entity-developer-app.html.twig`. (Originally, I implemented it a way like this.)

So with this, I am trying to avoid confusion and educate developers who may not know what EntityViewBuilder does and will do in the background if the fix lands in Drupal core. This is the reason why I left this comment. Everybody who would like to customize the look and feel of an entity should define the entity type specific theme and template because that is the one that always gets applied by EntityViewBuilder if exists.

apigee_edge.module
mxr576: @jacine Please read my explanation in the next comment which I hope also explains what do I mean when this way of doing this is KISS.
https://github.com/apigee/apigee-edge-drupal/pull/181#discussion_r278981143

Drupal core already provides a way to KISS without defining (copy-pasting) theme for all entities that you have. This is implementing a theme hook with the entity_type.

If this [patch](https://www.drupal.org/project/drupal/issues/2808481#comment-11775158) lands in core it will also try to suggest not to implement themes for all entities you have because Drupal will provide a default `entity` theme. So if will you need small customization you can do it by using the new `entity--ENTITY_TYPE.html.twig` template. If you need something more custom, something that always override the default entity theme, you should implement an entity type-specific theme hook and do your stuff there. If there is an entity type specific defined it will always win. This is why I am trying to suggest not to implement entity type specific themes in this module, leave it to the themers to do that in their theme/custom module and they will have full control over everything. Which I guess even better than fighting against defaults that a module provides/enforces.
This is how I understand what is happening in 2808481.

css/apigee_edge.app_view.css
mxr576: Ohh, I got it now. :) If there is an only reason to define an "app" theme and force it to be applied on developer- and team apps I would say it adds too much complexity to the code compared with the value it gives. I mean, you can theme the app page by using the default selectors provided by the apigee_edge_default theme with those selectors that I used in the app_view.css. Yes, `apigee-entity--app` would look nicer but we would need to maintain a new theme hook, preprocess function and template just to add this "nice" class to apps.
We are not trying to provide a complex front-end in this module, just something that works. Front-enders will customize it anyway according to the needs of a customer. This is the reason why we are trying to use Drupal core render elements only. I think this rule can be also applied here, we should not introduce a template for apps unless we have a really good reason.

apigee_edge.module
mxr576: Ok, I'll update the PR.

mxr576:
@jacine @cnovak @arshad Sorry, I have just realized that there was a missing reference to the related Drupal.org issue that we are trying to address by introducing the "apigee_entity". I added that reference to the first comment now, please check the issue and the latest patch and reevaulate this solution after that.

apigee_edge.module
mxr576: > The first thing someone looking to customize one of these templates will do is look for the a template named after the entity.
>  this is about enabling the ability to customize these entities just like anyone doing theme development would for any other entity: node, media, paragraph, or otherwise

Probably these entities only have default themes because https://www.drupal.org/project/drupal/issues/2808481  has not been fixed yet.

@Jaesin Please also FIXME if I am making the wrong assumption, but as a back-ender I think this is how a front-ender's mind works, when someone says you need to customize the developer_app/team/team_app page:
* As you wrote: look for the a template named after the entity.
* If there is no template, enable the theme debug and check what template is being applied on the page
* It is the apigee-entity.html.twig, no template suggestions other than that.
* Ok, let's check what it is in that template:
https://github.com/apigee/apigee-edge-drupal/blob/02a29d6f5082306bfa319eca57c5c157962ead1a/templates/apigee-entity-default.html.twig#L9-L31
* Cool, so I'll define my own template, I can also re-use as much I want, but my hands are not tied by default theme that the entity provides.

css/apigee_edge.app_view.css
mxr576: I see two ways to do that without creating a dedicated developer_app/team_app theme:

1. Create a common `app` theme and apply it on developer_app and team_apps in its view builder, as the apigee_kickstarter_enhancements does. I would not add additional complexity like this to the code just to add one "pretty" CSS class to a render element. If the "apigee_entity" theme would not define any selectors that a front-ender developer could use by default that would be a different situation. But this is not the case.
2. Attaching this custom CSS class tp app entities in the `template_preprocess_apigee_entity()` function based on the type of the entity. I would not do this, because I want to keep the "apigee_entity" theme generic, an actual fallback/workaround for the Drupal core issue that it addresses. But you will be able to add this nice class to apps in your module's `MY_MODULE_preprocess_apigee_entity()` function unless you want to define the "app" theme and override the entity view builder on apps.

src/Annotation/AppEntityType.php
```diff
@@ -0,0 +1,42 @@
...
+ * @Annotation
+ */
+class AppEntityType extends EdgeEntityType {
+
+  /**
+   * {@inheritdoc}
+   *
+   * Use the generic app theme, if exists for all app entity type.
+   */
+  public $default_theme = 'app';
```
mxr576: https://github.com/apigee/apigee-devportal-kickstart-drupal/blob/8.x-1.x/modules/custom/apigee_kickstart_enhancement/src/Entity/AppViewBuilder.php#L37-L41 is deprecated, thanks for this.

src/Entity/EdgeEntityViewBuilder.php
```diff
@@ -0,0 +1,64 @@
...
+    // not exist.
+    // When 2808481 gets in core, the default theme becomes the "entity". For
+    // BC, we only use that if it configured on the entity type to be used.
+    // @see apigee-entity.html.twig
+    // @see https://www.drupal.org/project/drupal/issues/2808481
+    if (!isset($defaults['#theme']) || $defaults['#theme'] === 'entity') {
+      /** @var \Drupal\apigee_edge\Entity\EdgeEntityTypeInterface $entity_type */
+      $entity_type = $entity->getEntityType();
+      // Add the default #theme key if a template exists for it.
+      if ($entity_type->getDefaultTheme() !== 'apigee_entity' && $this->themeRegistry->getRuntime()->has($entity_type->getDefaultTheme())) {
```
mxr576: This is a more robust solution that does not just allow to override the default theme used by the entity view builder to render this entity, but else helps to ensure BC after 2808481 gets solved. (See more details in the comment above.)

src/Entity/EdgeEntityTypeInterface.php
```diff
@@ -34,4 +34,14 @@ interface EdgeEntityTypeInterface extends EntityTypeInterface {
...
 
+  /**
+   * The default theme used by the entity view builder to render this entity.
+   *
+   * @return string
+   *   Name of the theme.
+   *
+   * @see \Drupal\apigee_edge\Entity\EdgeEntityViewBuilder::getBuildDefaults()
+   */
+  public function getDefaultTheme(): string;
```
mxr576: This is a BC break because it introduces a new public method, but I believe it worth it. See in EdgeEntityViewBuilder why.

mxr576:
@jacine @kittiradovics @arshad  I updated the PR, please review it again. I closed the old conversation that I believe this update addressed, feel free to re-open them if you disagree. 
Changes:
* I renamed the theme from "apigee_entity" to "apigee_entity_default".
* Added a more details comment to the twig template file about what is the "apigee_entity" theme and why it exists.
* Removed the concrete example about how an entity type specific theme can be defined and how the `template_preprocess_apigee_entity()` can be re-used in that.
* The default theme used by the EdgeEntityViewBuilder to render an entity if the entity type specific theme does not exist can be configured in the entity annotation. The default theme (after the entity type specific one) is "app" for developer- and team apps from now - if it exists.

css/apigee_edge.app_view.css
mxr576: The updated PR still does not add this "pretty" class to apps because of the reasons that I described in my previous comment ^.

mxr576: (Another option that came in my mind is allowing to define additional classes in the entity type annotation, but it seems an overkill just to tag all kinds of apps with a pretty "app" class.)

kittiradovics: Thanks for the reasoning. We can add the class from the preprocess function you mentioned.

apigee_edge.module
```diff
@@ -108,6 +114,41 @@ function apigee_edge_user_format_name_alter(&$name, AccountInterface $account) {
...
+  // Helpful $content variable for templates.
+  $variables += ['content' => []];
+  foreach (Element::children($variables['elements']) as $key) {
+    $variables['content'][$key] = $variables['elements'][$key];
+  }
+
+  // Helpful classes for front-end developers.
+  $variables['attributes']['class'][] = 'apigee-entity-default';
+  $variables['attributes']['class'][] = Html::getClass($entity->getEntityTypeId());
+  $variables['attributes']['class'][] = Html::getClass($entity->getEntityTypeId() . '--view-mode-' . $variables['elements']['#view_mode']);
```
jacine: The class attributes should be in the Twig template, not preprocess, which is more of a Drupal 7 thing.  In Drupal 8, this is [not encouraged anymore](https://www.drupal.org/node/2325067), as it makes it harder for themers to manipulate/remove the classes.

jacine: Also `-default` needs to be removed.  That has a loaded meaning in Drupal, as you know, and we don't want to confuse people into thinking this has to do with view modes. 

So, in the Twig template this should look like:

```twig
{% set classes = [
  'apigee-entity',
  view_mode ? 'apigee-entity--view-mode-' ~ view_mode|clean_class,
  entity.getEntityTypeId|clean_class,
  view_mode ? entity.getEntityTypeId|clean_class ~ '--view-mode-' ~ view_mode|clean_class,
] %}

{# Provide common classes for Apps. #}
{% if entity.getEntityTypeId in ['developer_app', 'team_app'] %}
  {% set classes = classes|merge([
    'apigee-app',
    'apigee-app--view-mode-' ~ view_mode|clean_class,
  ]) %}
{% endif %}

<article{{ attributes.addClass(classes)>
   ...
```

This would give when looking at a developer app, rendered with full view mode:

- `.apigee-entity`
- `.apigee-entity--view-mode-full`
- `.developer-app`
- `.developer-app--view-mode-full`
- `.apigee-app`
- `.apigee-app--view-mode-full`

apigee_edge.module
```diff
@@ -108,6 +114,41 @@ function apigee_edge_user_format_name_alter(&$name, AccountInterface $account) {
...
+
+  // Add article ARIA role.
+  $variables['attributes']['role'] = 'article';
```
jacine: This can also move to the template.

```twig
<div role="article" {{ attributes|without('role').addClass(classes) }}>
```

css/apigee_edge.app_view.css
jacine: We do not want or need to use an attribute selector here. Just because it's *possible* for a theme developer to handle this themselves doesn't mean it's right to make them have to. See one of my previous comments (above), which proposes the classes to add to the Twig template.  Once implemented, this can be changed to:

```css
.apigee-app--view-mode-full x {
  ...
}
```

src/Annotation/AppEntityType.php
jacine: If we are going to do this here, then why not just add the `app.html.twig` template?  That would help keep the main template a bit cleaner.  I still maintain that having explicit templates would be the best and simplest to understand solution, despite the additional code. 

We are also missing theme hook suggestions here, which should be provided in this PR.

src/Entity/EdgeEntityViewBuilder.php
jacine: While that issue would help here (with theme hook suggestions, which we can still create on our own), we are still left to deal with what we have in core, and as far as I understand #2808481:

- Is not fully fleshed out.
- Likely not going to make it into 8.x, for BC reasons.
- Is meant to add *additional* sane defaults for entities.

Right now, EntityViewBuilder always uses `$entity_type_id` as `#theme`. If we just use those, and implement theme hook suggestions, instead of trying to make one template do everything, we wouldn't need to do any of this.

templates/apigee-entity.html.twig
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * @see \Drupal\apigee_edge\Entity\EdgeEntityViewBuilder::getBuildDefaults()
+ * @see \Drupal\Core\Entity\EntityViewBuilder::getBuildDefaults()
+ * @see template_preprocess_apigee_entity()
+ * @see https://www.drupal.org/project/drupal/issues/2808481
+ *
+ * @ingroup themeable
+ */
+#}
+<article{{ attributes }}>
```
jacine: See one of my previous comments, for adding classes and ARIA role here.

apigee_edge.module
```diff
@@ -108,6 +114,41 @@ function apigee_edge_user_format_name_alter(&$name, AccountInterface $account) {
...
+ * Default template: apigee-entity.html.twig.
+ *
+ * @param array $variables
+ *   An associative array containing:
+ *   - elements: An array of elements to display in view mode.
+ */
+function template_preprocess_apigee_entity(array &$variables) {
+  $variables['view_mode'] = $variables['elements']['#view_mode'];
+  /** @var \Drupal\apigee_edge\Entity\EdgeEntityInterface $entity */
+  $entity = $variables['elements']['#entity'];
```
jacine: This should be exposed to the template at top level, as is done by core for node, and other contrib modules, for theming convenience.

```php
/** @var \Drupal\apigee_edge\Entity\EdgeEntityInterface $entity */
$variables['entity'] = $variables['elements']['#entity'];
```

jacine (changes requested):

mxr576:
@jacine thanks for the review, I'll go through your comments on Thursday. It is interesting that not just the patch in the referenced issue but also the Drupal core contains outdated template preprocess tricks.

src/Annotation/AppEntityType.php
mxr576: With this, I intended to do something that Drupal core does too in the EntityViewBuilder: dynamically apply a theme on a render array if it exists, without explicitly requiring that theme to be defined. Template suggestions for this theme should be defined when the "app" theme gets defined.
If you call "apigee-entity.html.twig" as the main template, yes, I intend to keep it simple and does not add entity type specific additional classes to it as you suggested belove.

apigee_edge.module
mxr576: Although I understand the problem and motivation behind reference CR, but Drupal core and contrib modules can and do attach custom CSS classes to render arrays in preprocess functions so we can not just ignore them
screeshot-2019-05-02-09-45-36

https://github.com/drupal/core/blob/7d1adf98928d69d509b7b743f4abb01205813ba0/modules/contextual/contextual.module#L129

Until we have to support them we can also leverage them as well: https://github.com/apigee/apigee-edge-drupal/pull/181/files#diff-228def7411670fa59e3a8fc1a707afceR49

mxr576:
@jacine Hopefully my latest update could be a common ground between how we imagine individually the optimal solution for #169 .

apigee_edge.module
jacine: > Until we have to support them we can also leverage them as well: https://github.com/apigee/apigee-edge-drupal/pull/181/files#diff-228def7411670fa59e3a8fc1a707afceR49

I think you misunderstand the features and how this works.

Sure, as long as we have preprocess (which may not actually be long at all TBH, as there are active issues trying to replace it), modules will be *able* to work with this variable.  However, this is *not* recommended as I mentioned, because it makes it more difficult to change things in the template. 

Preprocess is obviously still needed in some cases. The screenshot you posted is a prime example of where this is a good thing. Features like Contextual Links and Quick Edit do not make sense to handle at template level, so using preprocess is perfectly sensible and acceptable because:

- They are used/needed across too many theme hooks / templates.
- You actually _want_ to make it hard for theme developers to remove those classes, because it will literally break functionality when if they do.

Our case here, working with an entity is very different.

```
{% set classes = attributes.class|merge([
  ...
]) %}
```

This is wrong.

1. `classes` in the Twig template are explicitly separate and printed via `attributes`.  This is so they can be easily tossed by the theme developer, if they're not wanted. 

2.  You don't want or need to do this `attributes.class|merge([`.  `attributes` has any classes and whatever other attributes exist in the object from preprocess or render elements.  Simply printing `attributes` and using `.addClass`, handles any merging that needs to occur.  You don't need to do anything to "support" this.  This is handled for ALL theme hooks by Drupal core uniformly in Drupal 8.   See:

    -  [\Drupal\Core\Template\Attribute](https://git.drupalcode.org/project/drupal/blob/8.7.x/core/lib/Drupal/Core/Template/Attribute.php)
    - [template_preprocess()](https://api.drupal.org/api/drupal/core%21includes%21theme.inc/function/template_preprocess/8.7.x)
    - [_template_preprocess_default_variables()](https://api.drupal.org/api/drupal/core%21includes%21theme.inc/function/_template_preprocess_default_variables/8.7.x)

HTH.

apigee_edge.module
```diff
@@ -114,6 +115,14 @@ function apigee_edge_form_key_form_alter(array &$form, FormStateInterface $form_
  */
 function apigee_edge_theme() {
   return [
+    'apigee_entity' => [
+      'render element' => 'elements',
+    ],
+    'apigee_entity__app' => [
+      'render element' => 'elements',
+      // DRY.
```
jacine: Remove this comment?

apigee_edge.module
```diff
@@ -141,6 +150,53 @@ function apigee_edge_user_format_name_alter(&$name, AccountInterface $account) {
...
+
+  if ($entity instanceof AppInterface) {
+    $suggestions[] = 'apigee_entity__app';
+    $suggestions[] = 'apigee_entity__app__view_mode__' . $sanitized_view_mode;
+  }
+
+  $suggestions[] = 'apigee_entity__' . $entity->getEntityTypeId();
+  $suggestions[] = 'apigee_entity__' . $entity->getEntityTypeId() . '__view_mode__' . $sanitized_view_mode;
+
+  return $suggestions;
```
jacine, line 196: Great! Thanks 👍

css/apigee_edge.app_view.css
```diff
@@ -15,13 +15,13 @@
  * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
-.field {
+.apigee-entity--app--view-mode-full .field {
```
jacine, line 18: 👏

templates/apigee-entity--app.html.twig
```diff
@@ -0,0 +1,54 @@
...
+ * @ingroup themeable
+ */
+#}
+
+{% extends "apigee-entity.html.twig" %}
+{% block apigee_entity %}
+  {% if entity.getEntityTypeId in ['developer_app', 'team_app'] %}
+    {% set attributes = attributes.addClass('apigee-entity--app', 'apigee-entity--app--view-mode-' ~ view_mode|clean_class) %}
+  {% endif %}
+  {{ parent() }}
```
jacine: I think using Twig template inheritance here is ok, and the general idea of adding the app-specific here is exactly what I was getting at in a previous comment 👏, but this can/should be done differently.  `extends` is the strictest of all methods of doing this, and `parent()` suggests a really tightly coupled relation between the parent and child template, which is not necessarily going to be the case here.  If the structure of the base template is changed, and `{% block apigee_entity %}` is removed, this will throw a fatal error, which the person modifying the base template might not become aware of until after the change is pushed to production and someone happens to look at a page where an app renders and gets a WSOD. 

All we really need here is to use the main template and pass in additional classes, so structuring the base template to allow for that more easily would be ideal.  I can help you with this.

templates/apigee-entity.html.twig
```diff
@@ -0,0 +1,70 @@
...
+ * @see \Drupal\Core\Entity\EntityViewBuilder::getBuildDefaults()
+ * @see template_preprocess_apigee_entity()
+ * @see https://www.drupal.org/project/drupal/issues/2808481
+ *
+ * @ingroup themeable
+ */
+#}
+
+{% block apigee_entity %}
+  {% set classes = attributes.class|merge([
```
jacine: See my previous comment about this.  It's not necessary.

templates/apigee-entity--app.html.twig
jacine: RE: 
> I can help you with this.

There are 2 parts to this, as we have to make classes extendable from the base template.  I will post that in context in the apigee-entity.html.twig source, but here is this part:

```twig
{% if entity.getEntityTypeId in ['developer_app', 'team_app'] %}
  {% set classes = [
    'apigee-entity--app',
    'apigee-entity--app--view-mode-' ~ view_mode|clean_class
  ] %}
{% endif %}

{% include 'apigee-entity.html.twig' %}
```

Here we:

- Don't make any assumptions about the base template structure, we just use it directly.
- Add additional classes, which get passed through to the base template, and merged there.  More explicitly, this is equivalent to: `{% include 'apigee-entity.html.twig' with: { classes: classes }  %}`

templates/apigee-entity.html.twig
```diff
@@ -0,0 +1,70 @@
...
+      </h2>
+    {% endif %}
+    {{ title_suffix }}
+
+    <div{{ content_attributes }}>
+      {{ content }}
+    </div>
+
+  </article>
+{% endblock %}
```
jacine, line 75: The second part of this [comment](https://github.com/apigee/apigee-edge-drupal/pull/181/files#r280457148):

```twig
{%
  set classes = [
    'apigee-entity',
    view_mode ? 'apigee-entity--view-mode-' ~ view_mode|clean_class,
    entity.getEntityTypeId|clean_class,
    view_mode ? entity.getEntityTypeId|clean_class ~ '--view-mode-' ~ view_mode|clean_class,
  ]|merge(classes|default([]))
%}

{% block apigee_entity %}
  <article role="article"{{ attributes|without('role').addClass(classes) }}>

    {% block content %}
      {{ title_prefix }}
      {% if label and view_mode != 'full' %}
        <h2{{ title_attributes }}>
          {{ label }}
        </h2>
      {% endif %}
      {{ title_suffix }}

    <div{{ content_attributes }}>
      {{ content }}
    </div>
    {% endblock %}

  </article>
{% endblock %}
```

Here we:

-  Make `classes` extendable, via `  ]|merge(classes|default([]))` allowing to merge with any classes passed though from child/extending templates.
- Improve upon the Twig block structure: `{% block entityname %}` as you had, and adds `{% block content %}`. This is a better pattern, and core is not up to date here, so I think if we use inheritance here, this would be a more useful template.  It allows extending templates to use the wrapper, and affect or add to the content which is a common use case for entities.  Paragraphs module is a good [example](https://git.drupalcode.org/project/paragraphs/blob/8.x-1.x/templates/paragraph.html.twig) of this.

Note:

- You don't need a space here: `<article role="article" {{ attributes` . That space is handled by core, so I have removed it in my example.

jacine (commented):
@mxr576 I added some code, as promised.  Lmk if you have questions/concerns.  Hope it helps.

apigee_edge.module
mxr576: It is weird, I do remember that without 
```
{% set classes = attributes.class|merge([
  ...
]) %}
```
The classes added by preprocessed did not appear before.

templates/apigee-entity.html.twig
mxr576: Thanks @jacine for the detailed example! :)

apigee_edge.module
```diff
@@ -141,6 +149,53 @@ function apigee_edge_user_format_name_alter(&$name, AccountInterface $account) {
...
+  $entity = $variables['elements']['#entity'];
+  $sanitized_view_mode = str_replace('.', '_', $variables['elements']['#view_mode']);
+
+  if ($entity instanceof AppInterface) {
+    $suggestions[] = 'apigee_entity__app';
+    $suggestions[] = 'apigee_entity__app__view_mode__' . $sanitized_view_mode;
+  }
+
+  $suggestions[] = 'apigee_entity__' . $entity->getEntityTypeId();
+  $suggestions[] = 'apigee_entity__' . $entity->getEntityTypeId() . '__view_mode__' . $sanitized_view_mode;
```
jacine: @mxr576 I'm so sorry I missed this the last time around, but the `view_mode__` part should not be here.  This is done with CSS classes, but not with theme hook suggestions (I've never seen a template with this despite the potential confusion and clashes - there is an issue for this somewhere). Could we just change these real quick?

```diff
- $suggestions[] = 'apigee_entity__app__view_mode__' . $sanitized_view_mode;
+ $suggestions[] = 'apigee_entity__app__' . $sanitized_view_mode;
```

```diff
- $suggestions[] = 'apigee_entity__' . $entity->getEntityTypeId() . '__view_mode__' . $sanitized_view_mode;
+ $suggestions[] = 'apigee_entity__' . $entity->getEntityTypeId() . '__' . $sanitized_view_mode;
```

jacine (approved):
@mxr576 I missed one thing last time around. Sorry about that!  Maybe you can fix it real quick before merging?  It's also okay with me to do it as a follow-up,  if you prefer.  It's up to you.

Otherwise, this PR looks great. 👏 Thank you!!! Approved! 👍 

And thank you for being receptive to my feedback here.  😃

apigee_edge.module
mxr576: I implemented it like this because I agree with the answer in comment 44 to comment 43 here:
https://www.drupal.org/project/drupal/issues/2808481#comment-11784959
Although Apigee entities do not support bundles at this moment and they probably won't support them for a long time or never, but for consistency, it could make sense to use this pattern. Having "view-mode" in the template name may seem odd for the first sight, but it can not do any harm, although and unexpected, the unwaned clash could create problems. @jacine What do you think? 

This could be also related: https://www.drupal.org/project/drupal/issues/2831144

mxr576:
> And thank you for being receptive to my feedback here. 😃

@jacine I am always open for suggestions and ideas and brainstorming even if it sometimes does not seem like that from my comments :) I am open to learning new tricks and this is what happened here. First I thought you are suggestings something that could add more complexity to the module than value but after some iterations, I could understand how we both can win. Probably we could speed this up if we would have done a quick sync call in the middle of the week :)

Usually, when I am pushing back on things I am just trying to protect the scope that this module covers because I do not want it to become a monolith in a year or so. For me, this module is a bootstrap that provides integration between Apigee Edge and Drupal 8 in the _simplest_ but best possible way. This is the reason why we are/were prefer to rely on what Drupal core provides when we can and only implement something custom when literally there is nothing that we could re-use. (This approach should reduce the maintenance cost for us and in every project that depends on this module.)
Because side-projects, like Kickstarter, exists. Those side projects could add those enhancements, I guess mostly on the front-end side, that provides something better than OOTB Drupal components can do.

(F.e.: I do not think that table view is always the best possible way to render a list of entities - for examples developer/team apps - but this is the best generic way to do that. I expect that it can be easily customized (replaced) for customers who do not like it.)

So just let me know if the Apigee Edge module somehow makes it impossible or just makes it really complicated to do something in a side-project and we'll figure something out.

apigee_edge.module
jacine: I think it's definitely an anti-pattern. I haven't come across any templates using this pattern in over 3 years of doing D8 exclusively. Clashes are possible (we deal with this everyday in Drupal 8, think `field_image` on a node and also on a media entity), but they're not really a hinderance / a big deal. I've implemented custom theme hook suggestions hooks many times, for contextual reasons, but this was never one of them. It kinda gets into optimizing for edge case territory IMO. 

Those issues are going to take a LONG time to work out, IMO.  It's a tough issue, in general.  There are also lots of proposed changes ahead: https://www.drupal.org/project/issues/search?issue_tags=Theme%20System%20Modernization%20Initiative

So, yeah, I would take it out, but it also doesn't really do harm, so if you disagree I'm fine with you making that call.

mxr576: Ok, let's get rid of the "view mode" because that makes template suggestion names shorter and deal with the clashes if we ever start supporting bundles.

PR #182, opened 2019-04-24
[DRUP-722] Remove colons from submodule names.

Jaesin:


mxr576:
@Jaesin I accept this PR without a green Travis build because it only contains text changes.
Next time please provide a Travis CI build link that confirms all test pass after this change.
(`PHP_VERSION=7.1 DEPENDENCIES="--prefer-lowest" TEST_ROOT="modules/contrib/apigee_edge/tests"` builds usually fails because of a Drupal 8.6.0 core issue in the memory cache implementation that has been fixed since that release.)

mxr576 (approved):

PR #185, opened 2019-04-29
DRUP-748 The render array can contain control words

mxr576:
https://github.com/apigee/apigee-devportal-kickstart-drupal/blob/8.x-1.x/modules/custom/apigee_kickstart_enhancement/src/Entity/ListBuilder/DeveloperAppListBuilder.php#L37-L40

Travis: https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/525984241


shadcn (approved):
This looks good. Tested and confirmed it fixes the issue with ViewBuilder::viewMultiple. Thanks.

PR #188, opened 2019-05-03
DRUP-639 Drupal 8.7 support

Jaesin:
This is ready for testing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "php": ">=7.1",
         "apigee/apigee-client-php": "^2.0.1",
         "cweagans/composer-patches": "^1.6.5",
-        "drupal/core": "^8.6.3",
+        "drupal/core": "^8.7.0",
         "drupal/entity": "^1.0",
         "drupal/key": "^1.8",
         "php-http/guzzle6-adapter": "^1.1.1"
@@ -15,9 +15,8 @@
     "require-dev": {
         "behat/mink" : "dev-master#9ea1cebe",
         "drush/drush": "^9.0",
-        "egulias/email-validator": "1.2.14",
         "phpunit/phpunit": "^6.5",
-        "webflo/drupal-core-require-dev": "^8.6"
+        "webflo/drupal-core-require-dev": "^8.7.0"
     },
     "config": {
         "sort-packages": true
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,31 +27,31 @@ env:
     - PHP_VERSION=7.1 DEPENDENCIES="" TEST_ROOT="modules/contrib/apigee_edge/tests"
     - PHP_VERSION=7.1 DEPENDENCIES="" TEST_ROOT="modules/contrib/apigee_edge/tests" DB_DRIVER=pgsql DB_IMAGE=wodby/postgres:9.6-1.3.1
     - PHP_VERSION=7.2 DEPENDENCIES="" TEST_ROOT="modules/contrib/apigee_edge/tests" PHP_IMAGE=wodby/drupal-php:7.2-dev-4.5.0
-    - PHP_VERSION=7.1 DRUPAL_CORE=8.6.x-dev DEPENDENCIES="--prefer-lowest" TEST_ROOT="modules/contrib/apigee_edge/tests"
+    - PHP_VERSION=7.1 DRUPAL_CORE=8.7.x-dev DEPENDENCIES="--prefer-lowest" TEST_ROOT="modules/contrib/apigee_edge/tests"
     # Apigee Edge API product RBAC module tests.
     - PHP_VERSION=7.1 DEPENDENCIES="--prefer-lowest" TEST_ROOT="modules/contrib/apigee_edge/modules/apigee_edge_apiproduct_rbac/tests"
     - PHP_VERSION=7.1 DEPENDENCIES="" TEST_ROOT="modules/contrib/apigee_edge/modules/apigee_edge_apiproduct_rbac/tests"
     - PHP_VERSION=7.1 DEPENDENCIES="" TEST_ROOT="modules/contrib/apigee_edge/modules/apigee_edge_apiproduct_rbac/tests" DB_DRIVER=pgsql DB_IMAGE=wodby/postgres:9.6-1.3.1
     - PHP_VERSION=7.2 DEPENDENCIES="" TEST_ROOT="modules/contrib/apigee_edge/modules/apigee_edge_apiproduct_rbac/tests" PHP_IMAGE=wodby/drupal-php:7.2-dev-4.5.0
-    - PHP_VERSION=7.1 DRUPAL_CORE=8.6.x-dev DEPENDENCIES="--prefer-lowest" TEST_ROOT="modules/contrib/apigee_edge/modules/apigee_edge_apiproduct_rbac/tests"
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
         "drush/drush": "^9.0",
         "egulias/email-validator": "1.2.14",
```
mxr576: This should be removed as well because Drupal 8.7 requires egulias/email-validator >= 2.0.

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
```diff
@@ -270,8 +270,8 @@ protected function validateForm(callable $visitFormAsAdmin): void {
     // Test invalid password.
     $random_pass = $this->randomString();
     $page->fillField('Password', $random_pass);
-    $this->assertSendRequestMessage('.messages--error', "Failed to connect to Apigee Edge. The given username ({$this->username}) or password is incorrect. Error message: Unauthorized");
-    $web_assert->elementContains('css', 'textarea[data-drupal-selector="edit-debug-text"]', 'HTTP/1.1 401 Unauthorized');
+    $this->assertSendRequestMessage('.messages--error', "Failed to connect to Apigee Edge. Error message: Invalid host or public key. Please provide the valid host and key value.");
```
mxr576: This new error message by the MGMT API does not make any sense :) Is it a bug or should we handle this in the module, so we could still provide valuable information to an end-user about what went wrong?

@cnovak

mxr576:
https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/527659295

Jaesin:
This is passing on my local.

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
Jaesin: It looks like a bug in the management API to me but I think we should open a ticket and handle this in a new issue.

mxr576: Indeed, this is unrelated, but because of this change in the MGMT API, all test fail in all builds.
~This problem rather made me thinking, now I think we should not expose the original/raw error message to end-users just like this. We implemented in this way because we thought that it is the best for end-users and these particular messages will not change and the MGMT API should be able to tell the best what went wrong.
Although, after this problem occurred, maybe we should not display the MGMT API error messages we should display a translation of them. We could still display the original MGMT API error message in the Debug window.~ Suggestions that we displayed are already constructed in Drupal. So we only need to remove the assertion from tests on the expected error messages returned by the MGMT API.

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
```diff
@@ -287,7 +287,7 @@ protected function validateForm(callable $visitFormAsAdmin): void {
     // Test invalid endpoint.
     $invalid_domain = "{$this->randomGenerator->word(16)}.example.com";
     $page->fillField('Apigee Edge endpoint', "http://{$invalid_domain}/");
-    $this->assertSendRequestMessage('.messages--error', "Failed to connect to Apigee Edge. The given endpoint (http://{$invalid_domain}/) is incorrect or something is wrong with the connection. Error message: cURL error 6: Could not resolve host: {$invalid_domain} (see http://curl.haxx.se/libcurl/c/libcurl-errors.html)");
```
mxr576: This is also weird, I am getting the original error on my local that you have just fixed. Which PHP version or CURL lib version you have installed? Maybe this issue also highlights a different problem, we should not test the exceptions returned by an external API or a library.

image

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
mxr576: This change is only required because there is a regression bug in the new Apigee Edge MGMT API server release. If you use a valid org/username combination with an invalid password, instead of getting back HTTP 401 you get back HTTP 500 with the following error:

image

mxr576: @Jaesin @cnovak Please review and accept #192 if you agree that the solution I implemented can replace these changes you made.

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
Jaesin: OK, It looks like cURL may have changed there message. It's probably not worth digging into since the PR https://github.com/apigee/apigee-edge-drupal/pull/192 truncates the message fixing the issue.

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
Jaesin: I merged https://github.com/apigee/apigee-edge-drupal/pull/192 and pulled `8.x-1.x` into this branch. 

image

mxr576:
@Jaesin Do no use merge commits in feature branches, rebase your feature branch onto upstream 8.x-1.x.

According to my Travis CI build from #192, the Status report test is still failing. I'll check this today.
https://travis-ci.org/mxr576/apigee-devportal-drupal/jobs/528831290#L2642

tests/src/Functional/StatusReportTest.php
```diff
@@ -96,7 +96,7 @@ public function testInvalidCredentials() {
     $key->save();
 
     $this->drupalGet($status_report_path);
-    $this->assertSession()->pageTextContains('Unauthorized');
+    $this->assertSession()->pageTextContains('Invalid host or public key. Please provide the valid host and key value.');
```
mxr576: This change is deprecated by #193.

mxr576: If we would commit this change as-is then after the next MGMT API server release - that hopefully fixes this invalid/incorrect message - test builds would be broken again.

mxr576:
@Jaesin If you rebase this PR, 7718c90 and 1f821835 should be merged, because even if you update the required Drupal version to 8.7 it could not be installed until the email validator version constraint has not been removed.

tests/src/Functional/StatusReportTest.php
Jaesin: At that point we will know it's been fixed/changed. I suggest we leave this as we don't know if or when this change will be reverted.

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
```diff
@@ -321,7 +317,7 @@ protected function validateForm(callable $visitFormAsAdmin): void {
     // Test invalid client id.
     $client_id = $this->randomGenerator->word(8);
     $page->fillField('Client ID', $client_id);
-    $this->assertSendRequestMessage('.messages--error', "Failed to connect to the OAuth authorization server. The given username ({$this->username}) or password or client ID ({$client_id}) or client secret is incorrect. Error message: ");
+    $this->assertSendRequestMessage('.messages--error', "Failed to connect to the OAuth authorization server. The given username ({$this->username}) or password or client ID ({$client_id}) or client secret is incorrect. Error message: {\"error\":\"unauthorized\",\"error_description\":\"Bad credentials\"}");
```
mxr576: @Jaesin I do not think that you would like to re-add these :) Issue caused by the rebase?

mxr576 (changes requested):

mxr576 (approved):

Jaesin:
I fixed the rebase issue and tests are passing again:

image

Issue #189, opened 2019-05-04
API product access tab missing from API Product Admin settings


cnovak:
| Q                           | A
| --------------------------- | -----------------------------------------------
| Bug?                        | yes
| New Feature?                | no
| Module Version              |  2473dfac8287566034fccc54d6d559a6da731d90 dev-8.x-1.x-smartdocs
| PHP Client version          |  2.0.1
| Requires PHP client fix     | no
#### Actual Behavior
The API product settings admin page is missing the "API product access" tab, it is being shown on the "Developers" tab
<img width="956" alt="Screen Shot 2019-05-03 at 5 01 14 PM" src="https://user-images.githubusercontent.com/613093/57171035-8cd0f480-6dc5-11e9-9f6f-18c445b75880.png">

#### Expected Behavior

Should display on "API product" settings page

#### Steps to Reproduce the behavior

Go to `admin/config/apigee-edge/product-settings/alias`, you will see "API product access" tab is missing.  

#### Proposed Solution

Change tab to display on "API product" settings page

mxr576:
It is not "missing". It has been moved to a new place. See explanation in https://github.com/apigee/apigee-edge-drupal/pull/190#issuecomment-489514000.

mxr576:
PR has been merged, documentation has to be updated. @llynch00

PR #190, opened 2019-05-04
DRUP-761 Move API product access tab

cnovak:
#189 Move API Product access tab from developer settings to API product settings

### DIFF
--- a/apigee_edge.links.task.yml
+++ b/apigee_edge.links.task.yml
@@ -72,10 +72,10 @@ apigee_edge.settings.developer.cache:
   base_route: apigee_edge.settings.developer
   weight: 10
 
-apigee_edge.settings.developer.api_product_access:
+apigee_edge.settings.product.api_product_access:
   route_name: apigee_edge.settings.developer.api_product_access
   title: 'API product access'
-  base_route: apigee_edge.settings.developer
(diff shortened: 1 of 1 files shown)


mxr576:
@cnovak API product access tab moved from Products to Developers to avoid confusion and ensure consistency when Team API product had been introduced. #121 
I think if it would be displayed under "API products" and Team module is enabled, end-users may think that the configuration they provide on this UI also controls the API product access on the team level, but it does not. We could figure out a better way to display developer- and team API product access settings on the admin UI, but I do not think that moving _only_ the developer API product access under API products would be better than the current solution.

Also, the change you made probably require some adjustments in tests as well because changing routes should not be this easy. (This is what Travis CI build links could confirm.)

mxr576 (dismissed):
See my comment above.

mxr576:
(Missing JIRA issue number reference)

cnovak:
I will have to validate tests, will add that.  I referenced the GitHub bug number in this PR instead of JIRA.

mxr576:
> I referenced the GitHub bug number in this PR instead of JIRA.
But aren't all GH issues/PRs should be a reference in JIRA? So they could be scheduled for a sprint.

Jaesin (approved):
LGTM:

image

PR #191, opened 2019-05-04
DRUP-610 Add url as source for apidocs spec file and (DRUP-459) allow re-fetching it afterwards

arlina-espinoza:
This PR is for the API Docs submodule. It adds the option to specify the OpenAPI specification file as a URL, and upon saving the entity, it will fetch the file locally from that location. Features:

- Widget for choosing if providing the file as an upload or a URL. 
Screenshot_2019-05-04 Edit Identity Apigee Developer Portal Kickstart
Screenshot_2019-05-04 Edit Identity Apigee Developer Portal Kickstart(1)

- Validation of the URL (should resolve to a valid file - checks file extension, and that no HTTP error codes are returned).
Screenshot_2019-05-04 Edit Identity Apigee Developer Portal Kickstart(2)

- Fetching the file and saving it locally.
- Added a local task operation to "Update OpenAPI spec" file. This operation will re-fetch the associated file from the URL, and create a new revision if revisions are enabled. If the remote file hasn't changed (checking by MD5), the entity isn't changed/saved.
API Docs   Apigee Developer Portal Kickstart
Screenshot_2019-05-04 Are you sure you want to update the OpenAPI specification file from URL on API Doc Identity Apigee De


mxr576:
> Added a local task operation to "Update OpenAPI spec" file. This operation will re-fetch the associated file from the URL, and create a new revision if revisions are enabled. If the remote file hasn't changed (checking by MD5), the entity isn't changed/saved.

@arlina-espinoza
1. Is this local task and operation only available if the API documentation is imported from an URL and not from a file? If not, what it does when the source is a file? 
2. Would not be a bulk operation better replacement or an addition to this local tasks? Updating (let's call it re-importing or re-fetching) multiple API specifications with a local task only could be really time-consuming.

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
```diff
@@ -74,4 +79,40 @@ protected function getSettingsFormRoute(EntityTypeInterface $entity_type) {
...
+      $entity_type_id = $entity_type->id();
+      $route = new Route($entity_type->getLinkTemplate('update-spec-form'));
+      // Use the update_spec form handler.
+      if ($entity_type->getFormClass('update_spec')) {
+        $operation = 'update_spec';
+      }
+      $route
+        ->setDefaults([
+          '_entity_form' => "{$entity_type_id}.{$operation}",
+          '_title' => 'Update API Doc OpenAPI specification',
```
mxr576: Re-import? Re-fetch? Or something more explanatory than "update".

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -65,6 +69,7 @@
  *     "add-form" = "/admin/structure/apidoc/add",
  *     "edit-form" = "/admin/structure/apidoc/{apidoc}/edit",
  *     "delete-form" = "/admin/structure/apidoc/{apidoc}/delete",
+ *     "update-spec-form" = "/admin/structure/apidoc/{apidoc}/update",
```
mxr576: `import` or `fetch` instead of `update`? It could be confusing, especially because `edit` exists.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -134,6 +140,26 @@ public function setPublished(bool $published) : ApiDocInterface {
     return $this;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getSpecAsFile() : bool {
```
mxr576: I getter that always returns a boolean value should be an isser, `isSpecAsFile()`.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -134,6 +140,26 @@ public function setPublished(bool $published) : ApiDocInterface {
...
+      return TRUE;
+    }
+
+    return (bool) $this->get('spec_as_file')->getValue()[0]['value'];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function isRevisionable() : bool {
```
mxr576: This seems to me an unnecessary method. Whether an entity supports revisioning or not is revisionable is stored and decided on the entity type level, not per entity.

`\Drupal\Core\Entity\EntityTypeInterface::isRevisionable`

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -176,8 +202,23 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
       ])
       ->setDisplayConfigurable('form', TRUE);
 
+    $fields['spec_as_file'] = BaseFieldDefinition::create('boolean')
+      ->setLabel(t('Provide OpenAPI specification as a file'))
+      ->setDescription(t('Indicate if the OpenAPI spec will be provided as a
```
mxr576, line 187: Fix the indention of this line to comply with Drupal CS.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -176,8 +202,23 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
+                          file (or a URL otherwise).'))
+      ->setDefaultValue(TRUE)
+      ->setDisplayOptions('form', [
+        'type' => 'boolean_checkbox',
```
mxr576: Should not this be rather a select list, where the options are "URL" or "File"? You could even provide value specific descriptions on the API doc entity form. Also, "spec_file_source" could be a better and more easily understandable name for the field than "spec_as_file" which returns a boolean value.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -134,6 +140,26 @@ public function setPublished(bool $published) : ApiDocInterface {
     return $this;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getSpecAsFile() : bool {
+    // Default is to use spec as file.
+    if ($this->get('spec_as_file')->isEmpty()) {
+      return TRUE;
```
mxr576: I would be better to retrieve the default value programmatically from the field instead of being hardcoded.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -234,4 +299,83 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
+  public function preSave(EntityStorageInterface $storage) {
+    parent::preSave($storage);
+
+    $this->updateOpenApiSpecFile(FALSE, FALSE);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function updateOpenApiSpecFile($save = TRUE, $new_revision = TRUE) {
```
mxr576: Bool typehint is missing from parameters.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -234,4 +299,83 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
+    $needs_save = FALSE;
+
+    // If "spec_as_file", grab file from "file_link" and save it into the
+    // "spec" file field. The file_link field should already have validated that
+    // a valid file exists at that URL.
+
```
mxr576, line 295: Unnecessary empty line.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -234,4 +299,83 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
     return $fields;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function preSave(EntityStorageInterface $storage) {
+    parent::preSave($storage);
+
+    $this->updateOpenApiSpecFile(FALSE, FALSE);
```
mxr576: The entity storage handles the actual CRUD operations for an entity. It is just a Drupal specific "design decisiton" that the entity storage also calls operations on the entity _value_ object itself when it performs a CRUD operation. Because of this, the `updateOpenApiSpecFile`  would be in a better place in API doc entity storage than in the entity. (Where DI also possible.)

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -234,4 +299,83 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
+    // "spec" file field. The file_link field should already have validated that
+    // a valid file exists at that URL.
+
+    $spec_value = $this->get('spec')->isEmpty() ? [] : $this->get('spec')->getValue()[0];
+
+    if (!$this->getSpecAsFile()) {
+      $file_uri = $this->get('file_link')->getValue()[0]['uri'];
+      $data = file_get_contents($file_uri);
+      if (empty($data)) {
+        \Drupal::messenger()
```
mxr576: This service should be DI-d.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -234,4 +299,83 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
+            '%url' => $file_uri,
+          ]), 'error');
+        return FALSE;
+      }
+
+      // Only save file if it hasn't been fetched previously.
+      $data_md5 = md5($data);
+      $prev_md5 = $this->get('spec_md5')->isEmpty() ? NULL : $this->get('spec_md5')->getValue()[0]['value'];
+      if ($prev_md5 != $data_md5) {
+        $filename = \Drupal::service('file_system')->basename($file_uri);
```
mxr576: This service should be DI-d.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
mxr576: In addition, it would be probably a better design if the Open API spec fetcher would be decoupled from the entity/entity storage, it would be a separate service that could be replaced or mocked if needed. (Like the fetcher in the aggregator module, but we do not a robust pluggable system for fetching.)

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -234,4 +299,83 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
+
+    // If "spec_as_file", grab file from "file_link" and save it into the
+    // "spec" file field. The file_link field should already have validated that
+    // a valid file exists at that URL.
+
+    $spec_value = $this->get('spec')->isEmpty() ? [] : $this->get('spec')->getValue()[0];
+
+    if (!$this->getSpecAsFile()) {
+      $file_uri = $this->get('file_link')->getValue()[0]['uri'];
+      $data = file_get_contents($file_uri);
```
mxr576: If you would use the HTTP client to fetch the file rather than:
* The service that takes care of the import could be replaced, the file fetching process could be mocked by mocking the HTTP client
* The HTTP client could send the [If-Modified-Since](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since) header to the server which could return HTTP 304 without a response body if it has not changed. That could speed up the fetching process. See: `\Drupal\aggregator\Plugin\aggregator\fetcher\DefaultFetcher::fetch()`

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -234,4 +299,83 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
+
+      // Only save file if it hasn't been fetched previously.
+      $data_md5 = md5($data);
+      $prev_md5 = $this->get('spec_md5')->isEmpty() ? NULL : $this->get('spec_md5')->getValue()[0]['value'];
+      if ($prev_md5 != $data_md5) {
+        $filename = \Drupal::service('file_system')->basename($file_uri);
+        $specs_definition = $this->getFieldDefinition('spec')->getItemDefinition();
+        $target_dir = $specs_definition->getSetting('file_directory');
+        $uri_scheme = $specs_definition->getSetting('uri_scheme');
+        $file = file_save_data($data, "$uri_scheme://$target_dir/$filename", FILE_EXISTS_RENAME);
```
mxr576: What if the file could not be saved, for example, because the URI scheme is private and private FS has no been configured?
`@see \Drupal\apigee_edge\OauthTokenFileStorage::checkRequirements()`

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -234,4 +299,83 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
+        $this->set('spec', $spec_value);
+        $this->set('spec_md5', $data_md5);
+
+        $needs_save = TRUE;
+      }
+    }
+
+    elseif (!empty($spec_value['target_id'])) {
+      /* @var \Drupal\file\Entity\File $file */
+      $file = \Drupal::entityTypeManager()
```
mxr576: This service should be DI-d.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -234,4 +299,83 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
+        ->getStorage('file')
+        ->load($spec_value['target_id']);
+
+      if ($file) {
+        $prev_md5 = $this->get('spec_md5')->isEmpty() ? NULL : $this->get('spec_md5')->getValue()[0]['value'];
+        $file_md5 = md5_file($file->getFileUri());
+        if ($prev_md5 != $file_md5) {
+          $this->set('spec_md5', $file_md5);
+
```
mxr576: Unnecessary empty line.

modules/apigee_edge_apidocs/src/Form/ApiDocUpdateSpecForm.php
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  protected $operation = 'update_spec';
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getQuestion() {
+    return $this->t('Are you sure you want to update the OpenAPI specification
```
mxr576: Fix the indention of this line.

modules/apigee_edge_apidocs/src/Form/ApiDocUpdateSpecForm.php
```diff
@@ -0,0 +1,76 @@
...
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Messenger\MessengerTrait;
+use Drupal\Core\Url;
+
+/**
+ * Class ApiDocUpdateSpecForm.
+ */
+class ApiDocUpdateSpecForm extends ContentEntityConfirmFormBase {
+
+  use MessengerTrait;
```
mxr576: Do not just use the Messenger trait also inject the messenger service in the constructor as a parameter, otherwise `$this->messenger()` calls the global static container which is a bad-smell.

modules/apigee_edge_apidocs/src/Form/ApiDocUpdateSpecForm.php
```diff
@@ -0,0 +1,76 @@
...
+    $status = $entity->updateOpenApiSpecFile(TRUE, TRUE);
+
+    if ($status) {
+      $this->messenger()->addStatus($this->t('API Doc %label: updated the OpenAPI
+      specification file from URL.', [
+        '%label' => $this->entity->label(),
+      ]));
+    }
+    else {
+      $this->messenger()->addError($this->t('API Doc %label: could not update
```
mxr576: Fix the indention of this line.

modules/apigee_edge_apidocs/src/Form/ApiDocUpdateSpecForm.php
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    /* @var \Drupal\apigee_edge_apidocs\Entity\ApiDocInterface $entity */
+    $entity = $this->getEntity();
+    $status = $entity->updateOpenApiSpecFile(TRUE, TRUE);
+
+    if ($status) {
+      $this->messenger()->addStatus($this->t('API Doc %label: updated the OpenAPI
```
mxr576: Fix the indention of this line.

modules/apigee_edge_apidocs/src/Plugin/Validation/Constraint/ApiDocFileLinkConstraintValidator.php
```diff
@@ -0,0 +1,78 @@
...
+        $options = [
+          'exceptions' => TRUE,
+          'allow_redirects' => [
+            'strict' => TRUE,
+          ],
+        ];
+
+        // Perform only a HEAD method to save bandwidth.
+        /* @var $response ResponseInterface */
+        $response = \Drupal::httpClient()->head($url, $options);
```
mxr576: The client factory service should be DI-d.

modules/apigee_edge_apidocs/src/Plugin/Validation/Constraint/ApiDocFileLinkConstraintValidator.php
```diff
@@ -0,0 +1,78 @@
...
+      try {
+        $options = [
+          'exceptions' => TRUE,
+          'allow_redirects' => [
+            'strict' => TRUE,
+          ],
+        ];
+
+        // Perform only a HEAD method to save bandwidth.
+        /* @var $response ResponseInterface */
```
mxr576: Always use fully qualified class names.

modules/apigee_edge_apidocs/src/Plugin/Validation/Constraint/ApiDocFileLinkConstraintValidator.php
mxr576: Implement `\Drupal\Core\DependencyInjection\ContainerInjectionInterface` interface.

modules/apigee_edge_apidocs/src/Form/ApiDocUpdateSpecForm.php
```diff
@@ -0,0 +1,76 @@
...
+    parent::validateForm($form, $form_state);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    /* @var \Drupal\apigee_edge_apidocs\Entity\ApiDocInterface $entity */
+    $entity = $this->getEntity();
+    $status = $entity->updateOpenApiSpecFile(TRUE, TRUE);
```
mxr576: If the Open API spec fetcher becomes a service, the design here also changes. For example, the new service could provide two public methods:
a) for fetching Open API specification (and saving it to a new API doc entity in the API doc entity storage)
b) for updating Open API specification on a given API document entity *if* it has been created (its latest revision has been created) from an URL import and *if* the Open API file has changed on the remote server.

Although, having an Open API fetcher and an API documentation updater service would be a more robust design. The fetcher would only handle the fetching and the updater would update speciation on all or a give API documentation entity.

mxr576 (changes requested):
See my comments above.

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
arlina-espinoza: Renamed the operation to "reimport". :+1:

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: Renamed to "reimport".

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: Thanks! I've removed the method and code now checks on the entity type instead.

modules/apigee_edge_apidocs/src/Plugin/Validation/Constraint/ApiDocFileLinkConstraintValidator.php
arlina-espinoza: Done.

modules/apigee_edge_apidocs/src/Plugin/Validation/Constraint/ApiDocFileLinkConstraintValidator.php
arlina-espinoza: :+1:

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: Agreed, I've updated the field name, type and widget:
Screenshot_2019-05-08 Edit Inventory Apigee Developer Portal Kickstart

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: Refactored so this method was not needed anymore and was removed.

modules/apigee_edge_apidocs/src/Form/ApiDocUpdateSpecForm.php
arlina-espinoza: Done.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: Refactored - unused and removed.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: Moved to a service class that is using DI.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: Moved to a service class that is using DI.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: Moved to a service class that is using DI.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: Moved to a service "apigee_edge_apidocs.spec_fetcher".

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: Added validations to check if the scheme is private and has been configured, as well as if the destination directory is available and writable.

modules/apigee_edge_apidocs/apigee_edge_apidocs.links.task.yml
```diff
@@ -19,3 +19,9 @@ entity.apidoc.delete_form:
   base_route:  entity.apidoc.canonical
   title: 'Delete'
   weight: 10
+
+entity.apidoc.reimport_spec_form:
+  route_name:  entity.apidoc.reimport_spec_form
```
mxr576: Please remove double spaces from key value paires. `:  `

modules/apigee_edge_apidocs/src/ApiDocHtmlRouteProvider.php
```diff
@@ -47,6 +48,10 @@ public function getRoutes(EntityTypeInterface $entity_type) {
       $apidoc_collection_route->setDefault('_title', 'API Docs');
     }
 
+    if ($reimport_spec_route = $this->getUpdateSpecFormRoute($entity_type)) {
```
mxr576: s/Update/Reimport/g (Replace "Update" with "Reimport" in this file).

modules/apigee_edge_apidocs/src/ApiDocSpecFetcherInterface.php
```diff
@@ -0,0 +1,51 @@
...
+   * @param bool $new_revision
+   *   Boolean indicating if method should create a new revision when saving
+   *   the entity.
+   * @param bool $show_messages
+   *   Boolean indicating if method should display status messages.
+   *
+   * @return bool
+   *   Returns TRUE if the operation completed without errors.
+   */
+  public function fetchSpec(ApiDocInterface $apidoc, bool $save = TRUE, bool $new_revision = TRUE, bool $show_messages = TRUE) : bool;
```
mxr576: We do not have a PHPCS checker for this, but there should be no space between `)` and `:`.  `$show_messages = TRUE): bool;`

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
```diff
@@ -0,0 +1,238 @@
...
+    // validated that a valid file exists at that URL.
+    if ($apidoc->get('spec_file_source')->value == ApiDocInterface::SPEC_AS_URL) {
+
+      // If the file_link field is empty, return without error.
+      if ($apidoc->get('file_link')->isEmpty()) {
+        return TRUE;
+      }
+
+      $file_uri = $apidoc->get('file_link')->getValue()[0]['uri'];
+      $data = file_get_contents($file_uri);
```
mxr576: Still, use the HTTP client for downloading the file from the server. This is the only way how you can send the If-Modified-Since header.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
```diff
@@ -0,0 +1,238 @@
...
+      if (empty($data)) {
+        $message = 'Could not retrieve OpenAPI specification file located at %url';
+        $params = [
+          '%url' => $file_uri,
+        ];
+        $this->logger->error($message, $params);
+        if ($show_messages) {
```
mxr576: Instead of this boolean variable, it would be better if a service specific exception would be thrown if something goes wrong meanwhile the process (file not available, private FS is not configured, etc.) that callers could handle as they want.
For example, if the fetching process is running on the UI it makes sense to display error an error message to the user. If the fetching is running in an update hook, or Drush command, calling the messenger service does not make any sense.
This approach also allows displaying different messages for an error based on the current context.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
```diff
@@ -0,0 +1,238 @@
...
+          $this->messenger->addMessage($this->t($message, $params), 'error');
```
mxr576: `$this->t($message, $params)` This is an antipattern, you should not translate values of variables unless the variable's value has been already wrapped to `t()` before.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcherInterface.php
mxr576: I'd still recommend decoupling the file fetching logic from the remote resource and saving the remote file to an API documentation entity. Probably this should be two services where the "saver service" depends on the "fetcher service".

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -234,4 +281,13 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
     return $fields;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function preSave(EntityStorageInterface $storage) {
+    parent::preSave($storage);
+
+    \Drupal::service('apigee_edge_apidocs.spec_fetcher')->fetchSpec($this, FALSE, FALSE);
```
mxr576: Move this logic to the API doc entity's entity storage because there you can use dependency inject this service.

modules/apigee_edge_apidocs/src/Form/ApiDocUpdateSpecForm.php
```diff
@@ -0,0 +1,124 @@
...
+use Drupal\Core\Entity\EntityTypeBundleInfoInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Messenger\MessengerInterface;
+use Drupal\Core\Url;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Class ApiDocUpdateSpecForm.
+ */
+class ApiDocUpdateSpecForm extends ContentEntityConfirmFormBase {
```
mxr576: s/Update/Reimport/g

mxr576 (changes requested):

mxr576:
A kernel test for the API spec fetcher and a functional test for the spec importing/re-importing processes on the UI would be necessary.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
arlina-espinoza: I've pushed changes to use the _http_client_ service and check the _If-Modified-Since_ header.

modules/apigee_edge_apidocs/src/Form/ApiDocUpdateSpecForm.php
arlina-espinoza: The Open API spec fetcher becomes is now a service.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
arlina-espinoza: Done.

arlina-espinoza:
>     1. Is this local task and operation only available if the API documentation is imported from an URL and not from a file? If not, what it does when the source is a file?
Then nothing is updated on the entity, and a message gets displayed to the user saying _API Doc %label is using a file upload as source. Nothing to update._
> 
>     2. Would not be a bulk operation better replacement or an addition to this local tasks? Updating (let's call it re-importing or re-fetching) multiple API specifications with a local task only could be really time-consuming.
It would be a nice to have, perhaps a follow up task?

mxr576:
> Then nothing is updated on the entity, and a message gets displayed to the user saying API Doc %label is using a file upload as source. Nothing to update.

If it does nothing it should not be visible because end-users get confused if they have access to something that does nothing. With a simple access checker, this operation could be removed if the source provider is not the file upload on an API documentation entity.

> It would be a nice to have, perhaps a follow up task?

It could be a follow-up tasks, the question is whether we need both the operation, local task, and the bulk operation.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
```diff
@@ -0,0 +1,262 @@
...
+   *   The messenger service.
+   * @param \Psr\Log\LoggerInterface $logger
+   *   The logger service.
+   */
+  public function __construct(FileSystemInterface $file_system, ClientInterface $http_client, EntityTypeManagerInterface $entityTypeManager, MessengerInterface $messenger, LoggerInterface $logger) {
+    $this->fileSystem = $file_system;
+    $this->httpClient = $http_client;
+    $this->entityTypeManager = $entityTypeManager;
+    $this->messenger = $messenger;
+    $this->logger = $logger;
```
mxr576, line 96: Inject the string translation service and set `$this->stringTranslation` in the construction otherwise `$this->t()` calls the static container.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
```diff
@@ -0,0 +1,262 @@
...
+    switch ($type) {
+      case static::TYPE_ERROR:
+        $this->logger->error($message, $params);
+        break;
+      case static::TYPE_STATUS:
+      default:
+        $this->logger->info($message, $params);
+    }
+    if ($show_messages) {
+      $this->messenger->addMessage($this->t($message, $params), $type);
```
mxr576, line 229: I do not think this solves the problem with internationalization.
You can easily verify whether this work or not enable internationalization and try to translate `API Doc %label is using a file upload as source. Nothing to update.`.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcherInterface.php
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Interface ApiDocSpecFetcherInterface.
+ */
+interface ApiDocSpecFetcherInterface {
+
+  /**
+   * A status message.
+   */
+  const TYPE_STATUS = 'status';
```
mxr576, line 33: Define visibility.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
```diff
@@ -0,0 +1,262 @@
...
+      $options = [
+        'exceptions' => TRUE,
+        'allow_redirects' => [
+          'strict' => TRUE,
+        ],
+      ];
+
+      // Generate conditional GET header.
+      if (!$apidoc->get('fetched_timestamp')->isEmpty()) {
+        $request = $request->withAddedHeader('If-Modified-Since', gmdate(DateTimePlus::RFC7231, $apidoc->get('fetched_timestamp')->value));
```
mxr576, line 128: 🎉

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -231,7 +278,35 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
     return $fields;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function preSave(EntityStorageInterface $storage) {
+    parent::preSave($storage);
+
+    \Drupal::service('apigee_edge_apidocs.spec_fetcher')->fetchSpec($this, TRUE);
```
mxr576, line 294: Still, do this in the entity stroage and not in the entity object, you can use DI there. Override `\Drupal\Core\Entity\EntityStorageBase::doPreSave()` or \Drupal\Core\Entity\EntityStorageBase::save()`

mxr576 (changes requested):
Haven't finished with the review yet, TBC

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
Jaesin: The default `$show_messages` isn't ever overridden and the logger should be OK to use so we can just drop the parameter.

modules/apigee_edge_apidocs/src/ApiDocListBuilder.php
```diff
@@ -82,4 +82,22 @@ public function buildRow(EntityInterface $entity) {
...
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getOperations(EntityInterface $entity) {
+    $operations = parent::getOperations($entity);
+
+    // Add "Re-import OpenAPI spec" link.
+    if ($entity->access('update') && $entity->hasLinkTemplate('reimport-spec-form')) {
```
mxr576, line 92: Because of [this](https://github.com/apigee/apigee-edge-drupal/blob/b4d0ba7d54ee404255dece1fa04afa3829858d97/modules/apigee_edge_apidocs/src/ApiDocAccessControlHandler.php#L50-L52) access check needs to be changed here.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
```diff
@@ -0,0 +1,264 @@
...
```
mxr576, line 128: "GMT" is missing from the end of the value.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since#Syntax

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
```diff
@@ -0,0 +1,264 @@
...
+      }
+
+      try {
+        $response = $this->httpClient->send($request, $options);
+
+        // In case of a 304 Not Modified there are no changes, but update
+        // last fetched timestamp.
+        if ($response->getStatusCode() == 304) {
+          $apidoc->set('fetched_timestamp', time());
```
mxr576, line 137: Use `\Drupal\Component\Datetime\Time::getCurrentTime()` (from the injeced service: `datetime.time`) instead as per: https://www.drupal.org/node/2785211

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
```diff
@@ -0,0 +1,264 @@
...
+      if (!$apidoc->get('fetched_timestamp')->isEmpty()) {
+        $request = $request->withAddedHeader('If-Modified-Since', gmdate(DateTimePlus::RFC7231, $apidoc->get('fetched_timestamp')->value));
+      }
+
+      try {
+        $response = $this->httpClient->send($request, $options);
+
+        // In case of a 304 Not Modified there are no changes, but update
+        // last fetched timestamp.
+        if ($response->getStatusCode() == 304) {
```
mxr576, line 136: Use strict comparison.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
```diff
@@ -0,0 +1,264 @@
...
+        $spec_value = ['target_id' => $file->id()] + $spec_value;
+        $apidoc->set('spec', $spec_value);
+        $apidoc->set('spec_md5', $data_md5);
+        $apidoc->set('fetched_timestamp', time());
+
+        $needs_save = TRUE;
+      }
+    }
+
+    elseif ($apidoc->get('spec_file_source')->value == ApiDocInterface::SPEC_AS_FILE) {
```
mxr576, line 193: After you added a proper access check to this should not happen. I would consider throwing an exception here rather silently failing with a message.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
```diff
@@ -0,0 +1,264 @@
...
+   *   TRUE if message should be displayed to the UI as well.
+   */
+  private function log(ApiDocInterface $apidoc, string $type, string $message, array $params = [], bool $show_messages = TRUE) {
+    switch ($type) {
+      case static::TYPE_ERROR:
+        $this->logger->error($message, $params);
+        break;
+
+      case static::TYPE_STATUS:
+
```
mxr576, line 224: Unnecessary new line?

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
```diff
@@ -0,0 +1,264 @@
...
+   *
+   * @param string $destination
+   *   The specification file destination directory, including scheme.
+   *
+   * @throws \Exception
+   */
+  private function checkRequirements(string $destination): void {
+    // If using private filesystem, check that it's been configured.
+    if (strpos($destination, 'private://') === 0 && !$this->isPrivateFileSystemConfigured()) {
+      throw new \Exception('Private filesystem has not been configured.');
```
mxr576, line 246: Throw a module-specific exception or something more specific than a generic one so callers could differentiate what went wrong and what was the source of the exception.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
```diff
@@ -231,7 +278,35 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
...
+
+    // Update spec_md5 value if using "file" as source.
+    if ($this->get('spec_file_source')->value == static::SPEC_AS_FILE) {
```
mxr576, line 297: Strict comparison

modules/apigee_edge_apidocs/src/Form/ApiDocReimportSpecForm.php
```diff
@@ -0,0 +1,122 @@
...
+  public function getDescription() {
+    return $this->t('This will replace the current OpenAPI specification file.
+     This action cannot be undone.');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validateForm(array &$form, FormStateInterface $form_state) {
+    parent::validateForm($form, $form_state);
```
mxr576, line 100: Unnecessary override.

modules/apigee_edge_apidocs/src/Form/ApiDocReimportSpecForm.php
```diff
@@ -0,0 +1,122 @@
...
+    /* @var \Drupal\apigee_edge_apidocs\Entity\ApiDocInterface $entity */
+    $entity = $this->getEntity();
+
+    $needs_save = $this->specFetcher->fetchSpec($entity, TRUE);
+    if ($needs_save) {
+      if ($entity->getEntityType()->isRevisionable()) {
+        $entity->setNewRevision();
+      }
+      $entity->save();
+      $this->messenger->addStatus($this->t('API Doc %label: imported the OpenAPI specification file from URL.', [
```
mxr576, line 116: Call the method on `$this->messenger()` and not the property.
https://pronovix.com/blog/drupal-9-automated-deprecated-code-removal-proof-concept
"Autocorrect framework specific mistakes and ensure best practices"

modules/apigee_edge_apidocs/src/Plugin/Validation/Constraint/ApiDocFileLinkConstraint.php
```diff
@@ -0,0 +1,43 @@
...
+ * )
+ */
+class ApiDocFileLinkConstraint extends Constraint {
+
+  /**
+   * Message to be shown when it is not a valid link.
+   *
+   * @var string
+   */
+  public $notValid = '%value is not a valid link';
```
mxr576, line 41: It could happen that the URI is syntactically valid but it points to nowhere. IOW, this error message does not cover all possible issues that the `ApiDocFileLinkConstraintValidator` actually handles.

modules/apigee_edge_apidocs/src/Plugin/Validation/Constraint/ApiDocFileLinkConstraintValidator.php
```diff
@@ -0,0 +1,106 @@
...
+      }
+
+      $uri = $item->getValue()['uri'];
+
+      // Try to resolve the given URI to a URL. It may fail if it's schemeless.
+      try {
+        $url = Url::fromUri($uri, ['absolute' => TRUE])->toString();
+      }
+      catch (\InvalidArgumentException $e) {
+        $this->context->addViolation("The following error occurred while getting the link URL: @error", ['@error' => $e->getMessage()]);
```
mxr576, line 82: Always retrieve the validation error message from the constraint.
See `\Drupal\apigee_edge\Plugin\Validation\Constraint\DeveloperEmailUnique`

mxr576 (changes requested):

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
Jaesin: I don't think it's creating a new storage handler just for that.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
Jaesin: `const RFC7231 = 'D, d M Y H:i:s \G\M\T';`

modules/apigee_edge_apidocs/src/ApiDocListBuilder.php
Jaesin: Changing to `$entity->access('reimport')`.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
mxr576: Cool, I missed that.

modules/apigee_edge_apidocs/src/Entity/ApiDoc.php
mxr576: What is the problem with that? The storage handlers suppose to handle entity CRUD operations this is the reason why DI is available there and not in the entity classes that should be value objects only.

modules/apigee_edge_apidocs/src/ApiDocSpecFetcher.php
Jaesin: `time() !== REQUEST_TIME`. Why would it be necessary to go through `\Drupal\Component\Datetime\Time::getCurrentTime()` since it just calls the native PHP function.

https://api.drupal.org/api/drupal/core!lib!Drupal!Component!Datetime!Time.php/class/Time/8.7.x

mxr576: Yes, I checked this one (not like the RFC7231 ;P) , injecting time as a service (as a mockable PHP object) allows to mock time while you are testing the API doc fetcher service :)

Jaesin:
Closing in favor of: https://github.com/cnovak/apigee-api-catalog-drupal/pull/3

PR #195, opened 2019-05-14
[DRUP-787] Lock drupal core version to ~8.7.0

shadcn:


mxr576:
Commented DRUP-787.

Jaesin (approved):
We talked about this with @cnovak a bit and decided that the test matrix is already complex and it might not be an efficient use of time to deal with changes in 8.8.x during the development cycle. We can create a 8.8.x compatible branch to test against 8.8.0-rc1 when it is released.

PR #196, opened 2019-05-16
[DRUP-738] Rename the default RBAC attribute value from `APIGEE_EDGE_…

Jaesin:
…APIPRODUCT_RBAC` to `DRUPAL_RBAC`.


mxr576 (commented):
This won't affect the existing installations, they will remain using the old attribute name, so I am fine by this. Otherwise, we would need an update hook and an explanation in the changelog why we did this change.

mxr576 (approved):

PR #198, opened 2019-05-22
[DRUP-803] Deprecate apidocs module

Jaesin:


modules/apigee_edge_apidocs/apigee_edge_apidocs.info.yml
```diff
@@ -8,3 +8,4 @@ dependencies:
 - drupal:text
 - drupal:file
 - apigee_edge:apigee_edge
+- apigee_edge:deprecated
```
Jaesin: This makes the module uninstallable in the install screen before the requirements is ever reached.

image

modules/apigee_edge_apidocs/README.md
```diff
@@ -2,6 +2,13 @@
 
 A module to render OpenAPI specs as documentation to your API developers.
 
+## End of life
+
+This sub-project has been deprecated in favor of the __[Apigee API Catalog](https://www.drupal.org/project/apigee_api_catalog)__
+module. Uninstall `apigee_edge_apidocs` before enabling `apigee_api_catalog`. The Apigee API Catalog module has the same
+the same planned features as were planned for Apigee Edge API Docs and is currently more complete. This module will be
```
mxr576: _the same the same_

modules/apigee_edge_apidocs/apigee_edge_apidocs.info.yml
mxr576: Well, this is rather seems to be some kind of hack then a proper solution ;S Have you seen this doing by other modules?
What I originally suggested on the sprint planning call is an "install" phase check in `apigee_edge_apidocs_requirements()` that does not allow to enable the module anymore and displays a deprecation message. Isn't that possible?

modules/apigee_edge_teams/src/EventSubscriber/TeamStatusWarningSubscriber.php
```diff
@@ -100,9 +86,7 @@ class TeamStatusWarningSubscriber implements EventSubscriberInterface {
   public function __construct(AccountInterface $current_user, RouteMatchInterface $route_match, EntityTypeManagerInterface $entity_type_manager, TeamMembershipManagerInterface $team_membership_manager, MessengerInterface $messenger, TranslationInterface $string_translation) {
     $this->routeMatch = $route_match;
     $this->currentUser = $current_user;
-    $this->teamStorage = $entity_type_manager->getStorage('team');
```
mxr576, line 103: These are unrelated changes to the PR, please remove them.

modules/apigee_edge_teams/src/TeamPermissionHandler.php
```diff
@@ -167,10 +158,10 @@ public function getDeveloperPermissionsByTeam(TeamInterface $team, AccountInterf
     // is still member of the team in Apigee Edge.
     if (in_array($team->id(), $developer_team_ids)) {
       /** @var \Drupal\apigee_edge_teams\Entity\TeamRoleInterface $member_role */
-      $member_role = $this->teamRoleStorage->load(TeamRoleInterface::TEAM_MEMBER_ROLE);
+      $member_role = $this->entityTypeManager->getStorage('team_role')->load(TeamRoleInterface::TEAM_MEMBER_ROLE);
       $permissions += $member_role->getPermissions();
       /** @var \Drupal\apigee_edge_teams\Entity\TeamMemberRoleInterface|null $dev_team_role */
-      $dev_team_role = $this->teamMemberRoleStorage->loadByDeveloperAndTeam($account, $team);
+      $dev_team_role = $this->entityTypeManager->getStorage('team_member_role')->loadByDeveloperAndTeam($account, $team);
```
mxr576, line 164: Also unrelated changes, please remove them.

src/EventSubscriber/DeveloperStatusWarningSubscriber.php
```diff
@@ -84,7 +84,7 @@ final class DeveloperStatusWarningSubscriber implements EventSubscriberInterface
   public function __construct(AccountInterface $current_user, RouteMatchInterface $route_match, EntityTypeManagerInterface $entity_type_manager, MessengerInterface $messenger, TranslationInterface $string_translation) {
     $this->routeMatch = $route_match;
     $this->currentUser = $current_user;
-    $this->developerStorage = $entity_type_manager->getStorage('developer');
+    $this->entityTypeManager = $entity_type_manager;
```
mxr576, line 87: Also unrelated changes, please remove them.

src/KeyEntityFormEnhancer.php
```diff
@@ -67,11 +67,11 @@ final class KeyEntityFormEnhancer {
...
 
   /**
-   * The key entity storage.
+   * The entity type manager.
    *
-   * @var \Drupal\Core\Config\Entity\ConfigEntityStorageInterface
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
    */
-  private $keyStorage;
+  private $entityTypeManager;
```
mxr576, line 74: Also unrelated changes, please remove them.

mxr576 (dismissed):

mxr576:
@Jaesin please remove all unrelated changes from this PR.

mxr576:
@Jaesin Not sure that https://github.com/apigee/apigee-edge-drupal/pull/198/commits/b49093a592c1077dbe283367ad889dc230721f13 is the right resolution of the problem. I do not see any issue with the changed lines. Other contrib modules do the same and you can not alter them in a way like this. It is odd that we haven't seen this issue before, couldn't it be an API docs module related problem? How can I reproduce this issue? I tried, but I could not:

Peek 2019-05-22 09-15

Did you have any MINT module enabled?

modules/apigee_edge_apidocs/apigee_edge_apidocs.info.yml
```diff
@@ -1,4 +1,4 @@
-name: 'Apigee Edge API Docs'
+name: 'Apigee Edge API Docs (deprecated)'
```
mxr576, line 1: I would rather add it to a new `Apigee (Deprecated)` group.

modules/apigee_edge_apidocs/apigee_edge_apidocs.info.yml
mxr576: Actually what you have added to `apigee_edge_apidocs_requirements()` is perfectly enough.

Peek 2019-05-22 08-10

modules/apigee_edge_apidocs/apigee_edge_apidocs.install
```diff
@@ -0,0 +1,42 @@
...
+
+use Drupal\Core\Link;
+use Drupal\Core\Url;
+
+/**
+ * Implements hook_requirements().
+ */
+function apigee_edge_apidocs_requirements($phase) {
+  return [
+    'apigee_edge_apidocs_all' => [
```
mxr576, line 34: Probably this check should not run in the "update" phase because this could break deployments. Limit this to "install" and "runtime".

modules/apigee_edge_apidocs/apigee_edge_apidocs.info.yml
mxr576: Meh... so there is an issue with Drush... It does not run `hook_requirements()`.
https://github.com/drush-ops/drush/issues/3669
@Jaesin If we do not find a better workaround for this problem, please document in the info.yml with a comment why this "hack" was needed. Preferably with a deep link to the related Drush issue.

mxr576: The drawback of the info.yml trick is this:
![image](https://user-images.githubusercontent.com/1755573/58154887-f03a9d80-7c72-11e9-8db8-41a3b3ed7b0b.png)

(and the problem is the same with Drush).


Another potential solution that I could figure out is this. The possible drawback of this solution that module gets installed partially with Drush so we have to uninstall it.

```
<?php
/**
 * Implements hook_module_preinstall().
 */
function apigee_edge_apidocs_module_preinstall($module) {
  if (PHP_SAPI === 'cli' && $module === 'apigee_edge_apidocs') {
    // Workaround for a bug in Drush: https://github.com/drush-ops/drush/issues/3669
    \Drupal::service('module_installer')->uninstall(['apigee_edge_apidocs']);
    throw new RuntimeException('The Apigee Edge Apidocs module is deprecated and will be removed before Apigee Edge version 8.x-1.0 is released. Please use the Apigee API Catalog module as an alternative: https://www.drupal.org/project/apigee_api_catalog.');
  }
}
```

```sh
cli-drupal:/app$ drush pmu apigee_edge_apidocs -y
 [success] Successfully uninstalled: apigee_edge_apidocs
cli-drupal:/app$ drush en apigee_edge_apidocs -y

In apigee_edge_apidocs.module line 53:
                                                                                                                                                                                                                       
  The Apigee Edge Apidocs module is deprecated and will be removed before Apigee Edge version 8.x-1.0 is released. Please use the Apigee API Catalog module as an alternative: https://www.drupal.org/project/apigee_  
  api_catalog.                                                                                                                                                                                                         
                                                                                                                                                                                                                       

pm:enable [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-d|--debug] [-y|--yes] [--no] [--remote-host REMOTE-HOST] [--remote-user REMOTE-USER] [-r|--root ROOT] [-l|--uri URI] [--simulate] [--pipe] [-D|--define DEFINE] [--druplicon] [--xh-link XH-LINK] [--notify [NOTIFY]] [--] <command> [<modules>]...

cli-drupal:/app$ drush pm:list | grep apigee_edge_apidocs
  Apigee (Deprecated)                  Apigee Edge API Docs (apigee_edge_apidocs)                              Disabled   8.x-1.0-rc4   
```

At least Drupal Console calls `hook_requirements()`.

modules/apigee_edge_teams/src/EventSubscriber/TeamStatusWarningSubscriber.php
Jaesin: This is related as it prevents a clean uninstall of the `apigee_edge_apidocs` module. Accessing methods that access entity type discovery in a service constructor prevents clearing the entity type cache. See the following commit message: https://github.com/apigee/apigee-edge-drupal/pull/198/commits/b49093a592c1077dbe283367ad889dc230721f13

Jaesin:
To reproduce the `PluginNotFoundException` error that happens on uninstall, just install edge and apidocs and then uninstall the apidocs module through the UI using the current dev version (https://github.com/apigee/apigee-edge-drupal/commit/8a3938bdfa023891269b74562eb9d24f0411d004) of the `apigee_edge` module.

modules/apigee_edge_apidocs/apigee_edge_apidocs.install
Jaesin: There shouldn't be any update hooks added to this module.

modules/apigee_edge_apidocs/apigee_edge_apidocs.info.yml
Jaesin: Does drush 9 still try to install the module if you keep `- apigee_edge:deprecated`?

modules/apigee_edge_apidocs/README.md
Jaesin: Thanks for spotting, I pushed a fix.

modules/apigee_edge_apidocs/apigee_edge_apidocs.info.yml
Jaesin: @cnovak What do you think about this suggestion?

Jaesin:
Using `apigee_edge:deprecated` is definitely a hack but it prevents anyone from trying to install the module in the UI. I added a comment that the project doesn't actually exist. Keeping in mind that this sub-module will be completely removed before `apigee_edge` 8.x-1.0.0 is released I'm focusing on functionality.

With drush 8, I have the following result when I remove `apigee_edge:deprecated` form the dependencies:

```
>>> drush @apigee.l en apigee_edge_apidocs
Module apigee_edge_apidocs doesn't meet the requirements to be enabled.                                                                                                                                  [error]
The Apigee Edge Apidocs module is deprecated and will be removed before Apigee Edge version 8.x-1.0 is released. Please use the Apigee API Catalog module as an alternative.                             [error]
```

A note about services that access the entity type manager in their constructors. Of course it's OK to inject the entity type manager in a service constructor but accessing any methods that result in `getDefinitions` being called should not be used in a service constructor. When a module is uninstalled, the last thing that happens is the service container is rebuilt. This happens when the module that is being uninstalled is still discoverable. Any entities the module defines will be cached when `getDefinitions` is called on the entity type manager. 

This results in a `PluginNotFoundException` exception for subsequent page requests. The only way to get back to a working site is to clear the cache which would have to happen through drush. That is fine for developers but it is a problem for anyone that doesn't have drush or the correct alias set up.

mxr576:
> To reproduce the `PluginNotFoundException` error that happens on uninstall, just install edge and apidocs and then uninstall the apidocs module through the UI using the current dev version ([8a3938b](https://github.com/apigee/apigee-edge-drupal/commit/8a3938bdfa023891269b74562eb9d24f0411d004)) of the `apigee_edge` module.

@Jaesin This is what I did [here] exactly(https://github.com/apigee/apigee-edge-drupal/pull/198#issuecomment-494662902) 🙄Could you attach a test to this PR that confirms this bug exists and this is the right fix for that? IIRC we had a test that uninstall the Apigee Edge module to confirm that there is no issue like https://github.com/apigee/apigee-m10n-drupal/issues/75. I could not find that at this moment.

modules/apigee_edge_apidocs/apigee_edge_apidocs.install
mxr576: I did not suggest to add an update hook, I suggested to add a condition that checks `$phase! = 'update' `.
https://github.com/drupal/core/blob/8.6.x/lib/Drupal/Core/Extension/module.api.php#L888-L891

Jaesin:
I am using Drupal 8.7.1. with no other modules installed and the Apigee Edge credentials are not set. This is just after `drush si` and is also reproducible but installing and then uninstalling the `apigee_edge_teams` module.

Related: https://github.com/apigee/apigee-edge-drupal/issues/140

Jaesin:
plugin-not-found

mxr576:
Which is resolved by #141. Sorry it is too late here, I'll try to reproduce it again tomorrow morning. Did you used the standard installation profile? If you have some time, could you check whether this occurs with rc4 also?

Jaesin:
I changed the update phase to use `REQUIREMENT_WARNING` which allows your to still run DB updates but shows the warning.

image

Drush also shows: 

```
>>> drush @apigee.l updb
The Apigee Edge Apidocs module is deprecated and will be removed before Apigee Edge version 8.x-1.0 is released. Please use the <a href="https://www.drupal.org/project/apigee_api_catalog">Apigee API   [warning]
Catalog</a> module as an alternative.
No database updates required                                                                                                                                                                             [success]

```

modules/apigee_edge_apidocs/apigee_edge_apidocs.info.yml
Jaesin: I left the module name as is but changed the package to `Apigee (Deprecated)` for good measure.

modules/apigee_edge_apidocs/apigee_edge_apidocs.install
Jaesin: I changed the update phase to `REQUIREMENT_WARNING`.

modules/apigee_edge_apidocs/apigee_edge_apidocs.info.yml
cnovak: Either way is fine w/me as long as it is clear this module is going away.  Should we update the module description to point out that they should use the other module after uninstalling this one?

modules/apigee_edge_apidocs/apigee_edge_apidocs.install
cnovak: This brings up a warning page, hopefully this is not confusing to user to click "try again" to get it to go away.

Screen Shot 2019-05-22 at 1 31 50 PM

modules/apigee_edge_teams/src/EventSubscriber/TeamStatusWarningSubscriber.php
cnovak: If I remove @Jaesin 's changes I get the following error on uninstall, so these changes need to be in here:

```
The website encountered an unexpected error. Please try again later.</br></br><em class="placeholder">Drupal\Component\Plugin\Exception\PluginNotFoundException</em>: The &quot;apidoc&quot; entity type does not exist. in <em class="placeholder">Drupal\Core\Entity\EntityTypeManager-&gt;getDefinition()</em> (line <em class="placeholder">150</em> of <em class="placeholder">core/lib/Drupal/Core/Entity/EntityTypeManager.php</em>). <pre class="backtrace">Drupal\Core\Entity\EntityTypeManager-&gt;getHandler(&#039;apidoc&#039;, &#039;views_data&#039;) (Line: 157)
Drupal\Core\Entity\EntityManager-&gt;getHandler(&#039;apidoc&#039;, &#039;views_data&#039;) (Line: 177)
views_views_data()
call_user_func_array(&#039;views_views_data&#039;, Array) (Line: 392)
Drupal\Core\Extension\ModuleHandler-&gt;invoke(&#039;views&#039;, &#039;views_data&#039;) (Line: 243)
Drupal\views\ViewsData-&gt;getData() (Line: 160)
Drupal\views\ViewsData-&gt;get(&#039;apidoc&#039;) (Line: 91)
Drupal\views\Plugin\Derivative\ViewsEntityRow-&gt;getDerivativeDefinitions(Array) (Line: 101)
Drupal\Component\Plugin\Discovery\DerivativeDiscoveryDecorator-&gt;getDerivatives(Array) (Line: 87)
Drupal\Component\Plugin\Discovery\DerivativeDiscoveryDecorator-&gt;getDefinitions() (Line: 284)
Drupal\Core\Plugin\DefaultPluginManager-&gt;findDefinitions() (Line: 175)
Drupal\Core\Plugin\DefaultPluginManager-&gt;getDefinitions() (Line: 147)
views_theme(Array, &#039;module&#039;, &#039;views&#039;, &#039;core/modules/views&#039;) (Line: 447)
Drupal\Core\Theme\Registry-&gt;processExtension(Array, &#039;views&#039;, &#039;module&#039;, &#039;views&#039;, &#039;core/modules/views&#039;) (Line: 334)
Drupal\Core\Theme\Registry-&gt;build() (Line: 233)
Drupal\Core\Theme\Registry-&gt;get() (Line: 86)
Drupal\Core\Utility\ThemeRegistry-&gt;initializeRegistry() (Line: 67)
Drupal\Core\Utility\ThemeRegistry-&gt;__construct(&#039;theme_registry:runtime:seven&#039;, Object, Object, Array, 1) (Line: 253)
Drupal\Core\Theme\Registry-&gt;getRuntime() (Line: 142)
Drupal\Core\Theme\ThemeManager-&gt;render(Array, Array) (Line: 437)
Drupal\Core\Render\Renderer-&gt;doRender(Array, ) (Line: 195)
Drupal\Core\Render\Renderer-&gt;render(Array, ) (Line: 226)
Drupal\Core\Render\MainContent\HtmlRenderer-&gt;Drupal\Core\Render\MainContent\{closure}() (Line: 582)
Drupal\Core\Render\Renderer-&gt;executeInRenderContext(Object, Object) (Line: 227)
Drupal\Core\Render\MainContent\HtmlRenderer-&gt;prepare(Array, Object, Object) (Line: 117)
Drupal\Core\Render\MainContent\HtmlRenderer-&gt;renderResponse(Array, Object, Object) (Line: 90)
Drupal\Core\EventSubscriber\MainContentViewSubscriber-&gt;onViewRenderArray(Object, &#039;kernel.view&#039;, Object)
call_user_func(Array, Object, &#039;kernel.view&#039;, Object) (Line: 111)
Drupal\Component\EventDispatcher\ContainerAwareEventDispatcher-&gt;dispatch(&#039;kernel.view&#039;, Object) (Line: 156)
Symfony\Component\HttpKernel\HttpKernel-&gt;handleRaw(Object, 1) (Line: 68)
Symfony\Component\HttpKernel\HttpKernel-&gt;handle(Object, 1, 1) (Line: 57)
Drupal\Core\StackMiddleware\Session-&gt;handle(Object, 1, 1) (Line: 47)
Drupal\Core\StackMiddleware\KernelPreHandle-&gt;handle(Object, 1, 1) (Line: 106)
Drupal\page_cache\StackMiddleware\PageCache-&gt;pass(Object, 1, 1) (Line: 85)
Drupal\page_cache\StackMiddleware\PageCache-&gt;handle(Object, 1, 1) (Line: 47)
Drupal\Core\StackMiddleware\ReverseProxyMiddleware-&gt;handle(Object, 1, 1) (Line: 52)
Drupal\Core\StackMiddleware\NegotiationMiddleware-&gt;handle(Object, 1, 1) (Line: 23)
Stack\StackedHttpKernel-&gt;handle(Object, 1, 1) (Line: 693)
Drupal\Core\DrupalKernel-&gt;handle(Object) (Line: 19)
</pre>
```

cnovak (approved):

modules/apigee_edge_teams/src/TeamPermissionHandler.php
```diff
@@ -135,8 +128,6 @@ public function __construct(ModuleHandlerInterface $module_handler, ClassResolve
     $this->moduleHandler = $module_handler;
     $this->classResolver = $class_resolver;
     $this->teamMembershipManager = $team_membership_manager;
-    $this->teamRoleStorage = $entity_type_manager->getStorage('team_role');
-    $this->teamMemberRoleStorage = $entity_type_manager->getStorage('team_member_role');
```
Jaesin, line 139: This is missing `$this->entityTypeManager = $entity_type_manager;`

modules/apigee_edge_teams/src/EventSubscriber/TeamStatusWarningSubscriber.php
mxr576: It would have been great if someone had validated #141 before this got merged.

mxr576: > A note about services that access the entity type manager in their constructors. Of course it's OK to inject the entity type manager in a service constructor but accessing any methods that result in `getDefinitions` being called should not be used in a service constructor. When a module is uninstalled, the last thing that happens is the service container is rebuilt. This happens when the module that is being uninstalled is still discoverable. Any entities the module defines will be cached when `getDefinitions` is called on the entity type manager.
> 
> This results in a `PluginNotFoundException` exception for subsequent page requests. The only way to get back to a working site is to clear the cache which would have to happen through drush. That is fine for developers but it is a problem for anyone that doesn't have drush or the correct alias set up.

@Jaesin Thanks for the investigation and the detailed answer but you haven't captured the _real_ root of the problem. Generally speaking, there should not be any problem with retrieving an instance of a class from an object repository (EntityTypeManager) in a constructor - unless there is a bug somewhere in the system. This is what I tried to convince you in #140 and #141 and this is the reason why I continued my investigation regarding this problem today. Based on my findings this Drupal core issue causes the problem: https://www.drupal.org/project/drupal/issues/3056604

When you uninstall the Apigee Edge Teams module then it gets uninstalled, you can see in the logs: `apigee_edge_teams module uninstalled.`. After that HTTP Kernel triggers the `KernelEvents::RESPONSE ` event and the :hankey: event dispatcher calls the uninstalled Apigee Edge Teams module's `TeamStatusWarningSubscriber` event subscriber, but it should not.

screeshot-2019-05-23-11-08-39

(Stack trace captured after the Teams module uninstalled and the request has not been resolved. This confirms that the `TeamStatusWarningSubscriber` gets called because it is still registered in that instance of the event dispatcher and it fails because the "team" and "team_app" entity types have been unregistered already.)

Update: POC 2, what you changed only works because the `TeamStatusWarningSubscriber` only listening on team app routes at this moment. If you remove that condition from event subscriber you still get a WSOD.

https://drive.google.com/open?id=1jQKtoMerpQXvuRQtQ_2KReJVl7u62lu1

Jaesin: I found the same thing. Besides the `TeamStatusWarningSubscriber`, the `DeveloperStatusWarningSubscriber` causes the same issue. The teams module wasn't installed in the case of the error described in this issue.

This is not likely to get fixed in Drupal core any time soon (see: https://www.drupal.org/project/drupal/issues/3056604#comment-13118160). It is true that this results from issues in Drupal core but to be practical, we should work around it by not accessing methods that would cause the entity type cache to be rebuilt when the container is built. There aren't any **services** in Drupal core that use this pattern even through the entity type manager is used extensively in services so we should avoid it as well. Although the problem is showing itself while uninstalling modules, there could be edge cases where it shows up again.

mxr576: But entity type cache is not being rebuilt when you call `$entity_type_manager->getStorage()` in the event's constructor... it had been rebuilt when the module uninstalled (this is what [this method does](https://github.com/drupal/core/blob/8.7.2/lib/Drupal/Core/Extension/ModuleInstaller.php#L555)) and after that the only problem that Drupal core calls an uninstalled event dispatcher which tries to retrieve an uninstalled entity type. It does not matter where it tries to retrieve it, in its constructor or in the method that handles the event, it fails. (It does not even matter that this is an event subscriber, any code that runs in the same page request after a module has been uninstalled and it would try to retrieve an uninstalled entity type it would fail.) This is what [the video](https://drive.google.com/open?id=1jQKtoMerpQXvuRQtQ_2KReJVl7u62lu1) above confirms. 

> This is not likely to get fixed in Drupal core any time soon

I'll work on this, just as I worked on the JSON API related bugs to ensure this gets fixed sooner as you would expect.

Let's continue this conversation in #203.

PR #186, opened 2019-04-30
DRUP-747 SDKConnector refactoring, mock API responses in tests and extra test for Authentication form

mxr576:
For DRUP-747 we have to be able to replace the API client returned by the SDKConnector in a test. This could be possible without refactoring it but it is better for everyone if we do a last refactoring on the SDKConnector service before the stable release.

I wanted to do this refactoring for a long time ago this seems the best time to do that. These changes simplify the logic, apply OOP- and service decorator best practices on the SDKConnector. Because 99% of code relies on the SdkConnector::getClient() method this is a smaller BC breaking change.

Check the commit message for more details.

Latest Travis build: https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/535298059

PR #200, opened 2019-05-23
DRUP-789 Fix PHP notice caused by missing array key

Jaesin:
> "obscured" array key should not be removed because \Drupal\key\Form\KeyFormBase::submitForm() expects that it exist.

Logic fix from https://github.com/apigee/apigee-edge-drupal/pull/186.


mxr576 (approved):

mxr576:
@cnovak Please post-verify the CLAs, but they should be ok. What is wrong with the Googlebot? :)

cnovak:
I thought Googlebot would disallow merge.  It cannot verify when there is more than one author in a PR.  Verified

mxr576:
Meh, that is an odd limitation of the service ;S

PR #203, opened 2019-05-23
[DRUP-803] Real fix for the module uninstall problem + test coverage

mxr576:
As it revealed the real problem with the module uninstalls in #198 were caused by a Drupal core bug:
https://github.com/apigee/apigee-edge-drupal/pull/198#discussion_r286904065

This PR reverts the unnecessary changes and adds workarounds for the Drupal core. Also contains tests for module uninstalls.

Latest test on Travis CI: https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/538575563

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,11 @@
             "services": {
                 "drush.services.yml": "^9"
             }
+        },
+        "patches": {
+            "drupal/core": {
+                "Add/remove event listeners provided by a module on module install/uninstall": "https://www.drupal.org/files/issues/2019-05-28/uninstalled-module-kernel-response-listener-fix-3056604-29.patch"
+            }
         }
     }
 }
--- a/modules/apigee_edge_apidocs/tests/src/Functional/UninstallTest.php
+++ b/modules/apigee_edge_apidocs/tests/src/Functional/UninstallTest.php
@@ -1,0 +1,39 @@
+<?php
+
+/**
+ * Copyright 2018 Google Inc.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * version 2 as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+namespace Drupal\Tests\apigee_edge_apidocs\Functional;
+
+use Drupal\Tests\apigee_edge\Functional\UninstallModuleTestBase;
+
+/**
+ * Ensures the module can be uninstalled on the UI.
+ *
+ * @group apigee_edge
+ */
+class UninstallTest extends UninstallModuleTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function uninstalledModule(): string {
+    return 'apigee_edge_apidocs';
+  }
+
+}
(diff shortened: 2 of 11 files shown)


cnovak (changes requested):
We don't want to add try/catch blocks just to follow a different pattern than what core is doing.  Adding tests are a good idea, but we need to focus on getting features out the door instead of fixing solved problems in different ways.

mxr576:
> We don't want to add try/catch blocks just to follow a different pattern than what core is doing.

@cnovak I think you are misunderstanding what is going on here :) In Drupal core, you may not see this pattern* (which is bad, because hiding object dependencies is anti-pattern) but this problem is not about where do you retrieve entity storages from entity type manager in a class and why.
I know [this was a lengthy comment](https://github.com/apigee/apigee-edge-drupal/pull/198#discussion_r286904065) but if you [check the video](https://drive.google.com/file/d/1jQKtoMerpQXvuRQtQ_2KReJVl7u62lu1/view) in the end and read [the Drupal core issue's description](https://www.drupal.org/project/drupal/issues/3056604) you should understand the change in b49093a592

```php
<?php
class ResponseSubscriber implements EventSubscriberInterface {

  /**
   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
   */
  private $entity_type_manager;

  /**
   * ResponseSubscriber constructor.
   *
   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entityTypeManager
   */
  public function __construct(EntityTypeManagerInterface $entityTypeManager) {
    $this->entity_type_manager = $entityTypeManager;
  }

  /**
   * Does something with foo entities.
   *
   * @param \Symfony\Component\HttpKernel\Event\FilterResponseEvent $event
   *   The event.
   */
  public function handle(FilterResponseEvent $event) {
    $foo_entities = $this->entity_type_manager->getStorage('foo')->loadMultiple();
    // TODO Do something with foo entities.
  }

  /**
   * {@inheritdoc}
   */
  public static function getSubscribedEvents() {
    $events[KernelEvents::RESPONSE][] = ['handle'];
    return $events;
  }

}
```
and this

```php
<?php
  /**
   * ResponseSubscriber constructor.
   *
   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entityTypeManager
   */
  public function __construct(EntityTypeManagerInterface $entityTypeManager) {
    $this->fooStorage = $entityTypeManager->getStorage('foo');
  }

  /**
   * Does something with foo entities.
   *
   * @param \Symfony\Component\HttpKernel\Event\FilterResponseEvent $event
   *   The event.
   */
  public function handle(FilterResponseEvent $event) {
    $foo_entities = $this->fooStorage->loadMultiple();
    // TODO Do something with foo entities.
  }
```
fails exactly the same way with Drupal 8.7. So it does not matter where do you retrieve the entity storage  instance, until 3056604 has not been fixed in Drupal core. 

Consequently, [this change](https://github.com/apigee/apigee-edge-drupal/commit/b49093a592c1077dbe283367ad889dc230721f13) does not fix anything, it just works _at this moment_ because of the `TeamStatusWarningSubscriber` does not run on the module listing page. Otherwise, it would fail the same way as before.

Compared this with my fix which actually provides a proper temporary workaround for the Drupal core bug which could be easily followed in the new event subscribers and it could be removed when the Drupal core bug gets fixed.

mxr576:
Fixed failing Apigee Edge uninstall test because of dependencies. New test build: https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/536737890

mxr576:
@Jaesin Let's continue the discussion about this if you still disagree, shall we?
image

https://github.com/apigee/apigee-edge-drupal/pull/198#discussion_r287224672

mxr576:
I have a working patch for the Drupal core bug which is in review at this moment. Added that instead of the workaround that I implemented.

https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/537746931

mxr576:
The previous test also passed, but I updated to patch so updated this PR as well.
https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/537781480

mxr576:
Well, interesting. Apigee Edge tests passed with the [18](https://www.drupal.org/project/drupal/issues/3056604#comment-13122371) patch but failed with the [23](https://www.drupal.org/project/drupal/issues/3056604#comment-13122487) from the Drupal issue. Exactly the opposite happened on Drupal.org :) So the perfect location of the event subscriber unregistration should be in-between.

mxr576:
https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/538575563

cnovak:
The API Docs module was removed some time back, closing issue

PR #207, opened 2019-05-29
Bump minimum Drupal core requirement

mxr576:
in info.yml as well.

The missing piece from on #188.

https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/538649187


mxr576:
Found this when I scanned the module for other issues.

jacine (approved):
Nice catch!

PR #225, opened 2019-06-17
[BUG] Use sync must not depend on the debug module

mxr576:
Incorrect logger channel injected to the user sync service therefore the
Apigee Edge module started to depend on the Debug submodule.

https://travis-ci.org/mxr576/apigee-devportal-drupal/builds/546613257


Jaesin (approved):
LGTM

PR #227, opened 2019-06-18
[#115] Add a config to enable additional debug messages on the error page

### LINKED ISSUE #115: "Cannot reach management server" error and DeveloperDoesNotExist exception in logs
| Module Version              | dev branch
| PHP Client version          | 7.1.25
| Requires PHP client fix     | no|link to an Apigee Client PHP library issue or pull request

#### Actual Behavior

What is the actual behavior?

I was able successfully connect to Apigee Edge after completing a fresh install of the Apigee Developer Portal Kickstart module.  however, when I click on "My Apps", the connection fails and  I get the following error in  the log files:

```Drupal\apigee_edge\Exception\DeveloperDoesNotExistException: Developer with [email] email address not found. in Drupal\apigee_edge\Entity\ListBuilder\DeveloperAppListBuilderForDeveloper->checkDeveloperStatus() (line 48 of /app/web/modules/contrib/apigee_edge/src/Entity/DeveloperStatusCheckTrait.php).```

#### Expected Behavior

What is the behavior you expect?

I expect to see a list of apps associated with the product.

#### Proposed Solution

Although I was listed as an org administrator for the org I was connecting to, I had not been added as a Developer.  Manually adding myself as a developer to the org fixed the problem.  A developer sync through the Drupal admin might fix it as well.

### PR DESCRIPTION
**Pull request by shadcn**
Fixes #115 
> This message is confusing since the management server is configured properly and is able to reach the Apigee Edge management server, the actual issue is that the developer is not found in Apigee Edge.

As per #115 this PR adds a config to turn on additional debug messages on the error page. Instead of showing only the generic error message this also displays the message from the exception.

This setting can be turned off on production sites.

Error_page_settings___Drupal_8_m10n

Connection_error___Drupal_8_m10n

@cnovak @mxr576 Let me know what you think.

### DIFF
--- a/config/install/apigee_edge.error_page.yml
+++ b/config/install/apigee_edge.error_page.yml
@@ -3,3 +3,4 @@ error_page_title: 'Connection error'
 error_page_content:
  value: 'Cannot reach management server. An administrator should check the configuration of this site.'
  format: 'plain_text'
+error_page_debug_messages: true
--- a/src/EventSubscriber/EdgeExceptionSubscriber.php
+++ b/src/EventSubscriber/EdgeExceptionSubscriber.php
@@ -20,10 +20,16 @@
 namespace Drupal\apigee_edge\EventSubscriber;
 
 use Apigee\Edge\Exception\ApiException;
+use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\EventSubscriber\DefaultExceptionHtmlSubscriber;
+use Drupal\Core\Messenger\MessengerInterface;
+use Drupal\Core\Routing\RedirectDestinationInterface;
 use Drupal\Core\Utility\Error;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+use Symfony\Component\Routing\Matcher\UrlMatcherInterface;
 
 /**
  * Handles uncaught ApiExceptions.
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by shadcn** [comment]
Another example:

Connection_error___Drupal_8_m10n

**Review by cnovak — APPROVED** [review]
This is a great start for removing confusion when setting up module.  Added #231 so we can document this feature, and added #232 for adding tests, which I have done.

File: config/install/apigee_edge.error_page.yml [hunk 1]
```diff
@@ -3,3 +3,4 @@ error_page_title: 'Connection error'
 error_page_content:
  value: 'Cannot reach management server. An administrator should check the configuration of this site.'
  format: 'plain_text'
+error_page_debug_messages: true
```
mxr576, line 6: An update hook is missing from this PR that initiates this configuration on existing sites.

src/EventSubscriber/EdgeExceptionSubscriber.php
```diff
@@ -51,6 +93,11 @@ public function onException(GetResponseForExceptionEvent $event) {
       $context = Error::decodeException($event->getException());
       $this->logger->critical('@message %function (line %line of %file). <pre>@backtrace_string</pre>', $context);
       $this->makeSubrequest($event, '/api-communication-error', Response::HTTP_SERVICE_UNAVAILABLE);
+
+      // Display additional debug messages.
+      if ($this->configFactory->get('apigee_edge.error_page')->get('error_page_debug_messages')) {
+        $this->messenger->addError($event->getException()->getMessage());
```
mxr576, line 99: These error messages are never going to be translatable this way.

src/Exception/DeveloperDoesNotExistException.php
```diff
@@ -45,7 +45,7 @@ class DeveloperDoesNotExistException extends ApiException implements ApigeeEdgeE
    * @param \Throwable|null $previous
    *   Previous exception.
    */
-  public function __construct(string $email, string $message = 'Developer with @email email address not found.', int $code = 0, \Throwable $previous = NULL) {
+  public function __construct(string $email, string $message = 'Developer with @email email address not found. Running the developer sync could fix this.', int $code = 0, \Throwable $previous = NULL) {
```
mxr576, line 48: Okay, but only admins can run the developer sync... why we would suggest this to a simple authenticated user?

PR #230, opened 2019-06-19
Drush command to create Apigee Edge role

### LINKED ISSUE #210: Add Edge role creation script to module
Currently the [module's documentation installation steps](https://www.drupal.org/docs/8/modules/apigee-edge/configure-the-connection-to-apigee-edge#assigning-the-developer-administrator-role) say to download a gist from my GitHub repo, which is weird.  This should be instead a script in the module.

cnovak:
Fixes #210

apigee-edge:create-edge-role Drush command will create role will correct permissions for Apigee Edge connection.

### DIFF
--- a/apigee_edge.services.yml
+++ b/apigee_edge.services.yml
@@ -24,7 +24,7 @@ services:
 
   apigee_edge.key_entity_form_enhancer:
     class: Drupal\apigee_edge\KeyEntityFormEnhancer
-    arguments: ['@apigee_edge.sdk_connector', '@apigee_edge.authentication.oauth_token_storage', '@entity_type.manager', '@config.factory']
+    arguments: ['@apigee_edge.sdk_connector', '@apigee_edge.authentication.oauth_token_storage', '@entity_type.manager', '@config.factory', '@email.validator']
     calls:
       - [setMessenger, ['@messenger']]
       - [setStringTranslation, ['@string_translation']]
@@ -120,6 +120,8 @@ services:
       - '@logger.channel.apigee_edge'
       - '@redirect.destination'
       - '@router.no_access_checks'
+      - '@config.factory'
+      - '@messenger'
     tags:
       - { name: event_subscriber }
 
--- a/apigee_edge.services.yml
+++ b/apigee_edge.services.yml
@@ -13,6 +13,10 @@ services:
   apigee_edge.cli:
     class: Drupal\apigee_edge\CliService
 
+  apigee_edge.edge_connection_util:
+    class: Drupal\apigee_edge\util\EdgeConnectionUtilService
+    arguments: ['@http_client']
+
   apigee_edge.sdk_connector:
     class: Drupal\apigee_edge\SDKConnector
     arguments: ['@http_client_factory', '@key.repository', '@entity_type.manager', '@config.factory', '@module_handler', '@info_parser']
--- a/apigee_edge.services.yml
+++ b/apigee_edge.services.yml
@@ -12,9 +12,10 @@ services:
 
   apigee_edge.cli:
     class: Drupal\apigee_edge\CliService
+    arguments: ['@apigee_edge.apigee_edge_mgmt_cli_service']
 
-  apigee_edge.edge_connection_util:
-    class: Drupal\apigee_edge\util\EdgeConnectionUtilService
+  apigee_edge.apigee_edge_mgmt_cli_service:
+    class: Drupal\apigee_edge\Command\Util\ApigeeEdgeManagementCliService
     arguments: ['@http_client']
 
   apigee_edge.sdk_connector:
--- a/console/translations/en/apigee_edge.role.create.yml
+++ b/console/translations/en/apigee_edge.role.create.yml
@@ -1,0 +1,12 @@
+description: 'Create Apigee role for Drupal'
+help: 'Create a custom Apigee role that limits permissions for Drupal connections to the Apigee API.'
+arguments:
+  org: 'The Apigee Edge org to create the role in'
+  email: 'An Apigee user email address with orgadmin role for this org'
+options:
+  password: 'Password for the Apigee orgadmin user (if not passed in you will be prompted)'
+  base-url: 'Base URL to use, defaults to public cloud URL https://api.enterprise.apigee.com/v1'
+  role-name: 'The role to create in the Apigee Edge org, defaults to "drupalportal"'
+questions:
+  password: 'Enter password for Apigee orgadmin user'
+messages:
--- a/console/translations/en/apigee_edge.role.create.yml
+++ b/console/translations/en/apigee_edge.role.create.yml
@@ -7,6 +7,7 @@ options:
   password: 'Password for the Apigee orgadmin user (if not passed in you will be prompted)'
   base-url: 'Base URL to use, defaults to public cloud URL https://api.enterprise.apigee.com/v1'
   role-name: 'The role to create in the Apigee Edge org, defaults to "drupalportal"'
+  force: 'Force running of permissions on a role that already exists.'
 questions:
   password: 'Enter password for Apigee orgadmin user'
 messages:
--- a/console/translations/en/apigee_edge.role.create.yml
+++ b/console/translations/en/apigee_edge.role.create.yml
@@ -1,5 +1,5 @@
 description: 'Create Apigee role for Drupal'
-help: 'Create a custom Apigee role that limits permissions for Drupal connections to the Apigee API.'
+help: 'Create a custom Apigee role that limits permissions for Drupal connections to the Apigee API. '
 arguments:
   org: 'The Apigee Edge org to create the role in'
   email: 'An Apigee user email address with orgadmin role for this org'
@@ -7,7 +7,7 @@ options:
   password: 'Password for the Apigee orgadmin user (if not passed in you will be prompted)'
   base-url: 'Base URL to use, defaults to public cloud URL https://api.enterprise.apigee.com/v1'
   role-name: 'The role to create in the Apigee Edge org, defaults to "drupalportal"'
-  force: 'Force running of permissions on a role that already exists.'
+  force: 'Force running of permissions on a role that already exists. Note that permissions are only added, any current permissions not not removed.'
 questions:
   password: 'Enter password for Apigee orgadmin user'
 messages:
--- a/src/Command/CreateEdgeRoleCommand.php
+++ b/src/Command/CreateEdgeRoleCommand.php
@@ -21,6 +21,7 @@ namespace Drupal\apigee_edge\Command;
 
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
@@ -32,12 +33,6 @@ use Symfony\Component\Console\Output\OutputInterface;
  * )
  */
 class CreateEdgeRoleCommand extends CommandBase {
-
-  // Default base url.
-  const DEFAULT_BASE_URL = 'https://api.enterprise.apigee.com/v1';
-
-  // Default role name to create in Apigee Edge.
-  const DEFAULT_ROLE_NAME = 'drupalportal';
 
   /**
    * {@inheritdoc}
@@ -64,16 +59,23 @@ class CreateEdgeRoleCommand extends CommandBase {
         $this->trans('commands.apigee_edge.role.create.options.password')
       )
       ->addOption(
-      'base-url',
-      'b',
-      InputArgument::OPTIONAL,
-      $this->trans('commands.apigee_edge.role.create.options.base-url')
-    )->addOption(
+        'base-url',
+        'b',
+        InputArgument::OPTIONAL,
+        $this->trans('commands.apigee_edge.role.create.options.base-url')
+      )
+      ->addOption(
         'role-name',
         'r',
         InputArgument::OPTIONAL,
         $this->trans('commands.apigee_edge.role.create.options.role-name')
-      );;
+      )->addOption(
+        'force',
+        'f',
+        InputOption::VALUE_NONE,
+        $this->trans('commands.apigee_edge.role.create.options.force')
+      );
+
   }
 
   /**
@@ -100,8 +102,9 @@ class CreateEdgeRoleCommand extends CommandBase {
     $password = $input->getOption('password');
     $base_url = $input->getOption('base-url');
     $role_name = $input->getOption('role-name');
+    $force = $input->getOption('force');
 
-    $this->cliService->createEdgeRoleForDrupal($this->getIo(), 't', $org, $email, $password, $base_url, $role_name);
+    $this->cliService->createEdgeRoleForDrupal($this->getIo(), 't', $org, $email, $password, $base_url, $role_name, $force);
   }
 
 }
--- a/src/Command/CreateEdgeRoleCommand.php
+++ b/src/Command/CreateEdgeRoleCommand.php
@@ -1,0 +1,107 @@
+<?php
+
+/**
+ * Copyright 2019 Google Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License version 2 as published by the
+ * Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+namespace Drupal\apigee_edge\Command;
+
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Developer synchronization command class for Drupal Console.
+ *
+ * @Drupal\Console\Annotations\DrupalCommand (
+ *     extension="apigee_edge",
+ *     extensionType="module"
+ * )
+ */
+class CreateEdgeRoleCommand extends CommandBase {
+
+  // Default base url.
+  const DEFAULT_BASE_URL = 'https://api.enterprise.apigee.com/v1';
+
+  // Default role name to create in Apigee Edge.
+  const DEFAULT_ROLE_NAME = 'drupalportal';
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function configure() {
+    $this
+      ->setName('apigee_edge:role:create')
+      ->setDescription($this->trans('commands.apigee_edge.role.create.description'))
+      ->setHelp('commands.apigee_edge.role.create.help')
+      ->addArgument(
+        'org',
+        InputArgument::REQUIRED,
+        $this->trans('commands.apigee_edge.role.create.arguments.org')
+      )
+      ->addArgument(
+        'email',
+        InputArgument::REQUIRED,
+        $this->trans('commands.apigee_edge.role.create.arguments.email')
+      )
+      ->addOption(
+        'password',
+        'p',
+        InputArgument::OPTIONAL,
+        $this->trans('commands.apigee_edge.role.create.options.password')
+      )
+      ->addOption(
+      'base-url',
+      'b',
+      InputArgument::OPTIONAL,
+      $this->trans('commands.apigee_edge.role.create.options.base-url')
+    )->addOption(
+        'role-name',
+        'r',
+        InputArgument::OPTIONAL,
+        $this->trans('commands.apigee_edge.role.create.options.role-name')
+      );;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  function interact(InputInterface $input, OutputInterface $output) {
+    $this->setupIo($input, $output);
+    $password = $input->getOption('password');
+    if (!$password) {
+      $password = $this->getIo()->askHidden(
+        $this->trans('commands.apigee_edge.role.create.questions.password')
+      );
+      $input->setOption('password', $password);
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  function execute(InputInterface $input, OutputInterface $output) {
+    $this->setupIo($input, $output);
+    $org = $input->getArgument('org');
+    $email = $input->getArgument('email');
+    $password = $input->getOption('password');
+    $base_url = $input->getOption('base-url');
+    $role_name = $input->getOption('role-name');
+
+    $this->cliService->createEdgeRoleForDrupal($this->getIo(), 't', $org, $email, $password, $base_url, $role_name);
+  }
+
+}
--- a/src/Command/CreateEdgeRoleCommand.php
+++ b/src/Command/CreateEdgeRoleCommand.php
@@ -79,7 +79,7 @@ class CreateEdgeRoleCommand extends CommandBase {
   /**
    * {@inheritdoc}
    */
-  function interact(InputInterface $input, OutputInterface $output) {
+  public function interact(InputInterface $input, OutputInterface $output) {
     $this->setupIo($input, $output);
     $password = $input->getOption('password');
     if (!$password) {
@@ -93,7 +93,7 @@ class CreateEdgeRoleCommand extends CommandBase {
   /**
    * {@inheritdoc}
    */
-  function execute(InputInterface $input, OutputInterface $output) {
+  public function execute(InputInterface $input, OutputInterface $output) {
     $this->setupIo($input, $output);
     $org = $input->getArgument('org');
     $email = $input->getArgument('email');
--- a/src/Command/Util/ApigeeEdgeManagementCliService.php
+++ b/src/Command/Util/ApigeeEdgeManagementCliService.php
@@ -21,6 +21,7 @@ namespace Drupal\apigee_edge\Command\Util;
 
 use Drupal\Component\Utility\UrlHelper;
 use GuzzleHttp\ClientInterface;
+use Apigee\Edge\ClientInterface as ApigeeClientInterface;
 use GuzzleHttp\Exception\ClientException;
 use GuzzleHttp\Exception\TransferException;
 use Symfony\Component\Console\Style\StyleInterface;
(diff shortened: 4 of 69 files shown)


cnovak:
Tests for this module are failing: https://travis-ci.org/cnovak/apigee-edge-drupal/builds/547549135

Here is screenshot of all tests added running properly locally:

Screen Shot 2019-06-19 at 4 41 27 PM

src/Commands/ApigeeEdgeCommands.php
```diff
@@ -19,29 +19,43 @@
...
+use Drupal\apigee_edge\Util\EdgeConnectionUtilServiceInterface;
 use Drush\Commands\DrushCommands;
 
 /**
  * Drush 9 command file.
  */
 class ApigeeEdgeCommands extends DrushCommands {
 
+  // Default base url.
+  const DEFAULT_BASE_URL = 'https://api.enterprise.apigee.com/v1';
```
mxr576: This information belongs to the API client that takes care of the role creation, not to this base class. (See my comment below about re-using the API client from the PHP API client.)

src/Util/EdgeConnectionUtilService.php
```diff
@@ -0,0 +1,279 @@
...
+  protected $httpClient;
+
+  /**
+   * Constructs a EdgeRoleCreator object.
+   *
+   * @param \GuzzleHttp\ClientInterface $http_client
+   *   The HTTP client.
+   */
+  public function __construct(ClientInterface $http_client) {
+    $this->httpClient = $http_client;
```
mxr576: Even if the PHP API client does not support this API, you can still use it to communicate with this MGMT API. You do not need to start from scratch. The PHP API Client's API client can also provide you better HTTP error handling out of the box.

src/Util/EdgeConnectionUtilService.php
```diff
@@ -0,0 +1,279 @@
...
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+namespace Drupal\apigee_edge\Util;
```
mxr576: Move this to the `Drupal\apigee_edge\Command\Util\` namespace or something similar otherwise external developers may get tempted to use this in Drupal modules instead of the SDKConnector service.

src/Util/EdgeConnectionUtilService.php
```diff
@@ -0,0 +1,279 @@
...
+
+  /**
+   * The HTTP client.
+   *
+   * @var \GuzzleHttp\Client
+   */
+  protected $httpClient;
+
+  /**
+   * Constructs a EdgeRoleCreator object.
```
mxr576: Outdated docblock.

src/Util/EdgeConnectionUtilServiceInterface.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+namespace Drupal\apigee_edge\Util;
+
+use Symfony\Component\Console\Style\StyleInterface;
+
+/**
+ * Defines an interface for Edge connection classes.
+ */
+interface EdgeConnectionUtilServiceInterface {
```
mxr576: We probably do not want anybody to replace the implementation shipped by this module so we do not need an interface for this.

src/Util/EdgeConnectionUtilService.php
```diff
@@ -0,0 +1,279 @@
...
+        return;
+      }
+    }
+    else {
+      $io->text('Role ' . $role_name . ' already exists.');
+    }
+
+    $this->setDefaultPermissions($io, $t, $org, $email, $password, $base_url, $role_name);
+
+    $io->success($t('Role !role is configured.  Log into apigee.com to assign a user to this role.', ['!role' => $role_name]));
```
mxr576: Unnecessary extra space: `ed.  Log i`

src/Commands/ApigeeEdgeCommands.php
```diff
@@ -19,29 +19,43 @@
...
+
+  // Default role name to create in Apigee Edge.
+  const DEFAULT_ROLE_NAME = 'drupalportal';
```
mxr576: This information belongs to the service that takes care of the role creation, not to this base class.

src/Commands/ApigeeEdgeCommands.php
```diff
@@ -57,4 +71,61 @@ public function sync() {
...
+   */
+  public function createEdgeRole(
+    string $org,
+    string $email,
+    array $options = [
+      'password' => NULL,
+      'base-url' => self::DEFAULT_BASE_URL,
+      'role-name' => self::DEFAULT_ROLE_NAME,
+    ]) {
+    $this->edgeConnectionUtilService->createEdgeRoleForDrupal($this->io(), 'dt', $org, $email, $options['password'], $options['base-url'], $options['role-name']);
```
mxr576: User input should be validated, ex.: base-url should be a valid URI, role name should be string, etc. Symfony's options resolver could handle this easily.

src/Util/EdgeConnectionUtilService.php
```diff
@@ -0,0 +1,279 @@
...
+
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Exception\ClientException;
+use GuzzleHttp\Exception\TransferException;
+use Symfony\Component\Console\Style\StyleInterface;
+
+/**
+ * Defines an interface for Edge connection classes.
+ */
+class EdgeConnectionUtilService implements EdgeConnectionUtilServiceInterface {
```
mxr576: How about calling this as `ApigeeEdgeManagementCliService` or something else that provides a more descriptive and also generic name? I add `Cli` because this service (class) should be used only by CLIs tools.

src/Util/EdgeConnectionUtilService.php
```diff
@@ -0,0 +1,279 @@
...
+      '/apiproducts' => ['get', 'put'],
+      // GET, PUT, & DELETE access.
+      '/developers' => ['get', 'put', 'delete'],
+      '/developers/*/apps' => ['get', 'put', 'delete'],
+      '/developers/*/apps/*' => ['get', 'put', 'delete'],
+      '/companies/*' => ['get', 'put', 'delete'],
+      '/companies/*/apps/*' => ['get', 'put', 'delete'],
+    ];
+
+    $url = $base_url . '/o/' . $org . '/userroles/' . $role_name . '/permissions';
```
mxr576: This form is better and easier readable `"/o/{$org}/userroles/{$role_name}/permissions";`

src/Util/EdgeConnectionUtilService.php
```diff
@@ -0,0 +1,279 @@
...
+   * @param string $base_url
+   *   The base url of the Edge API.
+   * @param string $role_name
+   *   The role name to add the permissions to.
+   */
+  protected function setDefaultPermissions(StyleInterface $io, callable $t, string $org, string $email, string $password, string $base_url, string $role_name) {
+    $io->text('Setting permissions on role ' . $role_name . '.');
+
+    $permissions = [
+      // GET access.
```
mxr576: Here you should rather describe in comments why we grant only `GET` or only `GET & PUT` or `GET, PUT & DELETE` access to certain API endpoints. These comments could provide more value than the current ones for later when someone needs to change something on this permission matrix.

mxr576 (changes requested):
Did not do a complete review, I've just I shared some initial suggestions.
Just like the sync supports both Drupal Console and Drush, this new command should also support Drupal Console as well. See `console.services.yml`

apigee_edge.services.yml
```diff
@@ -13,6 +13,10 @@ services:
   apigee_edge.cli:
     class: Drupal\apigee_edge\CliService
 
+  apigee_edge.edge_connection_util:
```
mxr576: IIRC we planned to use the `apigee_edge.cli` service as the source for all CLI commands that this module provides. Maybe you should add the create role command to this service and that way we do not need to worry about how we should call this new service.

src/Util/EdgeConnectionUtilService.php
cnovak: The Drush commands will need to also use this directory, but that works for me

src/Util/EdgeConnectionUtilService.php
cnovak: done

src/Util/EdgeConnectionUtilServiceInterface.php
cnovak: Always good for testing purposes to have interfaces

src/Util/EdgeConnectionUtilService.php
cnovak: done

src/Commands/ApigeeEdgeCommands.php
cnovak: Moved to `EdgeConnectionUtilServiceInterface`

src/Commands/ApigeeEdgeCommands.php
cnovak: Moved to `EdgeConnectionUtilServiceInterface`

src/Commands/ApigeeEdgeCommands.php
cnovak: Turns out that role and other params can be a number and not sure what the rules are, so we can let the Apigee API respond back with any errors. Added validation for Base URL however.

src/Util/EdgeConnectionUtilService.php
cnovak: Done

src/Util/EdgeConnectionUtilService.php
cnovak: Done

apigee_edge.services.yml
cnovak: Done

src/Util/EdgeConnectionUtilService.php
cnovak: good idea, done

src/Command/CreateEdgeRoleCommand.php
```diff
@@ -0,0 +1,107 @@
...
+        'p',
+        InputArgument::OPTIONAL,
+        $this->trans('commands.apigee_edge.role.create.options.password')
+      )
+      ->addOption(
+      'base-url',
+      'b',
+      InputArgument::OPTIONAL,
+      $this->trans('commands.apigee_edge.role.create.options.base-url')
+    )->addOption(
```
kscheirer: needs more indent

src/Command/Util/ApigeeEdgeManagementCliService.php
```diff
@@ -0,0 +1,330 @@
...
+   */
+  public function createEdgeRoleForDrupal(StyleInterface $io,
+                                          callable $t,
+                                          string $org,
+                                          string $email,
+                                          string $password,
+                                          ?string $base_url,
+                                          ?string $role_name) {
+
+    if ($base_url === NULL || $base_url === '') {
```
kscheirer: `empty($base_url)` is a little shorter, unless you really only want to test for those 2 conditions.

src/Command/Util/ApigeeEdgeManagementCliService.php
```diff
@@ -0,0 +1,330 @@
...
+      // Set default base URL if var is null or empty string.
+      $base_url = self::DEFAULT_BASE_URL;
+    }
+    else {
+      // Validate it is a valid URL.
+      if (!UrlHelper::isValid($base_url, TRUE)) {
+        $io->error($t('Base URL is not valid.'));
+        return;
+      }
+
```
kscheirer: extra blank line

src/Command/Util/ApigeeEdgeManagementCliService.php
```diff
@@ -0,0 +1,330 @@
...
+    if (!$does_role_exist) {
+      $io->text($t('Role :role does not exist. Creating role.', [':role' => $role_name]));
+
+      $url = "{$base_url}/o/{$org}/userroles";
+      try {
+        $this->httpClient->post($url, [
+          'body' => json_encode([
+            'role' => [$role_name],
+          ]),
+          'auth' => [$email, $password],
```
kscheirer, line 101: This won't support oauth-enabled orgs I think. Is there a way to use the php-sdk instead here, which should handle that for you?

src/Command/Util/ApigeeEdgeManagementCliService.php
```diff
@@ -0,0 +1,330 @@
...
+      catch (TransferException $exception) {
+        $this->handleHttpClientExceptions($exception, $io, $t, $url, $org, $email);
+        return;
+      }
+    }
+    else {
+      $io->text('Role ' . $role_name . ' already exists.');
+    }
+
+    $this->setDefaultPermissions($io, $t, $org, $email, $password, $base_url, $role_name);
```
kscheirer, line 114: This could potentially reset some permissions if the role already exists and had its permissions modified from default. Maybe only set defaults if the role was just created? ie move into `try {}` block

src/Command/Util/ApigeeEdgeManagementCliService.php
```diff
@@ -0,0 +1,330 @@
...
+        'auth' => [$email, $password],
+        'headers' => ['Accept' => 'application/json'],
+      ]);
+    }
+    catch (TransferException $exception) {
+      $this->handleHttpClientExceptions($exception, $io, $t, $url, $org, $email);
+      return FALSE;
+    }
+
+    $body = json_decode($response->getBody());
```
kscheirer: `JSON_THROW_ON_ERROR` would be nice here, but requires php 7.3 :(

src/Command/Util/ApigeeEdgeManagementCliService.php
```diff
@@ -0,0 +1,330 @@
...
+   * @param string $url
+   *   The url being connected to.
+   * @param string $org
+   *   The organization to connect to.
+   * @param string $email
+   *   The email of an Edge user with org admin role to make Edge API calls.
+   */
+  public function handleHttpClientExceptions(TransferException $exception, StyleInterface $io, callable $t, string $url, string $org, string $email): void {
+    $io->error($t('Error connecting to Apigee Edge. :exception_message', [':exception_message' => $exception->getMessage()]));
+    switch ($exception->getCode()) {
```
kscheirer, line 316: add a default case to handle when the error is not 0,401,403,302?

src/Command/Util/ApigeeEdgeManagementCliService.php
```diff
@@ -0,0 +1,330 @@
...
+        ]));
+        return;
+
+      case 302:
+        $io->note($t('Edge endpoint gives a redirect response, is the url :url does not seem to be a valid Apigee Edge endpoint.', [':url' => $url]));
+        return;
+
+    }
+  }
+
```
kscheirer, line 339: extra blank line

kscheirer (commented):
Looks pretty good, only supporting basic auth seems like an issue though. Better than supporting no auths though :)

src/Command/CreateEdgeRoleCommand.php
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * @Drupal\Console\Annotations\DrupalCommand (
+ *     extension="apigee_edge",
+ *     extensionType="module"
+ * )
+ */
+class CreateEdgeRoleCommand extends CommandBase {
+
+  // Default base url.
+  const DEFAULT_BASE_URL = 'https://api.enterprise.apigee.com/v1';
```
arlina-espinoza: It would be better to use the default endpoint from the PHP client library (`Apigee\Edge\ClientInterface::DEFAULT_ENDPOINT`) instead of duplicating the value here.

src/Command/Util/ApigeeEdgeManagementCliService.php
```diff
@@ -0,0 +1,330 @@
...
+        $io->error($t('Base URL is not valid.'));
+        return;
+      }
+
+    }
+
+    if ($role_name === NULL || $role_name === '') {
+      // Set default if null or empty string.
+      $role_name = self::DEFAULT_ROLE_NAME;
+    }
```
arlina-espinoza: Just a suggestion: `$role_name = $role_name ?: self::DEFAULT_ROLE_NAME;`

src/Command/Util/ApigeeEdgeManagementCliService.php
arlina-espinoza: Maybe adding another command line option to "force" resetting permissions if the role has already been created?

src/Command/Util/ApigeeEdgeManagementCliService.php
```diff
@@ -0,0 +1,330 @@
...
+      '/' => ['get'],
+      // Read only access to environments.
+      '/environments/' => ['get'],
+      '/environments/*/stats/*' => ['get'],
+      // We do not need to update/edit roles, just read them.
+      '/userroles' => ['get'],
+      // No need to create API products, only read and edit.
+      '/apiproducts' => ['get', 'put'],
+      // Full CRUD for developers.
+      '/developers' => ['get', 'put', 'delete'],
```
arlina-espinoza, line 151: I think this is missing the `/developers/*` path, it would be:
```
'/developers' => ['get', 'put'],
'/developers/*' => ['get', 'put', 'delete'],
```

arlina-espinoza: Sorry, further reading of the docs (https://docs.apigee.com/api-platform/system-administration/managing-roles-api#aboutpermissionsetting-syntaxandprecedence) made me see your code is correct - disregard previous comment.

src/Command/Util/ApigeeEdgeManagementCliServiceInterface.php
```diff
@@ -0,0 +1,55 @@
...
+
+use Symfony\Component\Console\Style\StyleInterface;
+
+/**
+ * Defines an interface for Edge connection classes.
+ */
+interface ApigeeEdgeManagementCliServiceInterface {
+
+  // Default base url.
+  const DEFAULT_BASE_URL = 'https://api.enterprise.apigee.com/v1';
```
arlina-espinoza: It would be better to use the constant from `Apigee\Edge\ClientInterface::DEFAULT_ENDPOINT`.

tests/src/Functional/ApiProductAccessTest.php
```diff
@@ -306,8 +306,8 @@ protected function developerAppEditFormTest() {
 
     // >> Bypass user.
     $this->drupalLogin($this->users[self::USER_WITH_BYPASS_PERM]);
-    // Even if a user has bypass permission it should see only those API
-    // Products on an other user's add/edit form that the other user has
+    // Even if a user has bypass permission they should see only those API
```
arlina-espinoza, line 309: Changes in this file (tests/src/Functional/ApiProductAccessTest.php) seem unrelated to this PR - although I'm ok with adding them as they are only comment typo fixes.

tests/src/Unit/Command/CreateEdgeRoleCommandTest.php
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function setUp() {
+      parent::setUp();
+      $this->cliService = $this->prophesize(CliServiceInterface::class);
+      $this->logMessageParser = $this->prophesize(LogMessageParserInterface::class);
+      $this->loggerChannelFactory = $this->prophesize(LoggerChannelFactoryInterface::class);
+      $this->createEdgeRoleCommand = new CreateEdgeRoleCommand($this->cliService->reveal(),
```
arlina-espinoza, line 104: Running this test locally gives me an error because Edge does not require Drupal Console. Trying it out in travis right now, but I think we should at least add Drupal Console to `.travis/composer.json`.

arlina-espinoza (changes requested):
I've added some comments. I think besides @kscheirer suggestions, we should add Drupal Console to travis' composer build, or to the module's require-dev. We had talked before about adding it to `apigee/devportal-kickstart-project-composer`, but I think it is needed here as well.

src/Command/Util/ApigeeEdgeManagementCliService.php
arlina-espinoza: Also per the docs, this call "does not remove any existing permissions in the role", but I agree with @kscheirer that calling it for an existing role (which might have been altered outside of this call) might create the  impression that this command validated the role to only have the required permissions. I think that a) either exit with error if the role already exists, or b) force resetting the role, which would involve deleting all permissions and adding the right ones again.

src/Util/EdgeConnectionUtilService.php
arlina-espinoza: This would be an interesting addition. I think using the PHP API client would be more consistent. The `apigee_edge.sdk_connector` service has a `getClient()` method that you can pass in a custom auth and endpoint, so maybe that can work.
Also, I think adding these API endpoints to the management API would be better in the long run - although not a blocker for adding this here.

cnovak: I added a new issue for this #274 so that we can handle it in a new PR.

cnovak (commented):
> Looks pretty good, only supporting basic auth seems like an issue though. Better than supporting no auths though :)

Agreed, new issue #274 will handle this.

src/Command/CreateEdgeRoleCommand.php
cnovak: done

src/Command/Util/ApigeeEdgeManagementCliService.php
cnovak: done

src/Command/Util/ApigeeEdgeManagementCliService.php
cnovak: Are you not suppose to have blank lines in if/else? I changed it however.

src/Command/Util/ApigeeEdgeManagementCliService.php
cnovak: new issue #274 will handle this.

src/Command/Util/ApigeeEdgeManagementCliService.php
cnovak: Was thinking this was OK, but you both bring up points that someone may make changes to the role so we do not want to stomp those changes.

src/Command/Util/ApigeeEdgeManagementCliService.php
cnovak: Agree, but we would need to up the PHP version allowed for our module, I don't think we should restrict it that much yet

src/Command/Util/ApigeeEdgeManagementCliService.php
cnovak: Line 304 `$io->error()` will display out info about any error, the switch statement is only to add extra info for the end user for common use cases. Here is example of output from response status code of 506:

Screen Shot 2019-10-04 at 3 53 57 PM

Adding comment to section to clarify.

tests/src/Functional/ApiProductAccessTest.php
cnovak: This is unrelated, did a comment typo fix.

tests/src/Unit/Command/CreateEdgeRoleCommandTest.php
cnovak: I will also add drush which is missing

src/Command/Util/ApigeeEdgeManagementCliService.php
cnovak: Added a `--force` flag

cnovak:
Added changes, this PR needs to be looked over again to see if it should be approved

cnovak:
@arlina-espinoza the merge conflict is now fixed, ready for review again.

console/translations/en/apigee_edge.role.create.yml
```diff
@@ -0,0 +1,13 @@
+description: 'Create Apigee role for Drupal'
+help: 'Create a custom Apigee role that limits permissions for Drupal connections to the Apigee API.'
+arguments:
+  org: 'The Apigee Edge org to create the role in'
+  email: 'An Apigee user email address with orgadmin role for this org'
+options:
+  password: 'Password for the Apigee orgadmin user (if not passed in you will be prompted)'
+  base-url: 'Base URL to use, defaults to public cloud URL https://api.enterprise.apigee.com/v1'
+  role-name: 'The role to create in the Apigee Edge org, defaults to "drupalportal"'
+  force: 'Force running of permissions on a role that already exists.'
```
arlina-espinoza: Does this mean that it will add the permissions, or reset the permissions of the role if it already exists? The wording does not make that clear.

tests/src/Functional/ApiProductAccessTest.php
```diff
@@ -306,8 +306,8 @@ protected function developerAppEditFormTest() {
...
+    // Products on on an other user's add/edit form that the other user has
```
arlina-espinoza:
```suggestion
    // Products on another user's add/edit form that the other user has
```

arlina-espinoza (changes requested):
Just a couple of typos/wording changes requested.
Otherwise it looks good! I tested the command using drush and drupal console, and both worked correctly, creating the role with the right permissions:
Screenshot_2019-12-12 arlina-9da89b94-eval - Apigee

console/translations/en/apigee_edge.role.create.yml
cnovak: Check out my new wording in last commit to see if it helps explain

console/translations/en/apigee_edge.role.create.yml
```diff
@@ -0,0 +1,13 @@
+description: 'Create Apigee role for Drupal'
+help: 'Create a custom Apigee role that limits permissions for Drupal connections to the Apigee API. '
+arguments:
+  org: 'The Apigee Edge org to create the role in'
+  email: 'An Apigee user email address with orgadmin role for this org'
+options:
+  password: 'Password for the Apigee orgadmin user (if not passed in you will be prompted)'
+  base-url: 'Base URL to use, defaults to public cloud URL https://api.enterprise.apigee.com/v1'
+  role-name: 'The role to create in the Apigee Edge org, defaults to "drupalportal"'
+  force: 'Force running of permissions on a role that already exists. Note that permissions are only added, any current permissions not not removed.'
```
arlina-espinoza, line 10:
```suggestion
  force: 'Force running of permissions on a role that already exists. Note that permissions are only added, any current permissions are not removed.'
```

arlina-espinoza (approved):
@cnovak All LGTM! Just typo suggestion I made.

PR #242, opened 2019-07-17
[#241] Handle exceptions for $entity->toUrl()

### LINKED ISSUE #241: Exception not handled for $entity->toUrl() in template_preprocess_apigee_entity
**Describe the bug**
In `template_preprocess_apigee_entity`, exceptions are not caught for `$entity->toUrl()`. This results in a fatal error if `toUrl()` fails.

**To Reproduce**
See https://github.com/apigee/apigee-m10n-drupal/issues/166

**Expected behavior**
Wrap into a try catch so that we do not see a fatal error.

shadcn:
Fixes #241

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -166,7 +166,12 @@ function template_preprocess_apigee_entity(array &$variables) {
   $variables['label'] = $entity->label();
 
   if (!$entity->isNew() && $entity->hasLinkTemplate('canonical')) {
-    $variables['url'] = $entity->toUrl('canonical', ['language' => $entity->language()])->toString();
+    try {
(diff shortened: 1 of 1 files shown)


mxr576 (changes requested):
Based on this explanation I think this PR should not be merged.

https://github.com/apigee/apigee-m10n-drupal/issues/166#issuecomment-512287020

mxr576:
If you check usages of `toUrl()` in Drupal core you do not see that it would be try-catched either.

mxr576:
@arshad @cnovak Can we close this PR as a proper fix for the real issue has been merged with https://github.com/apigee/apigee-m10n-drupal/pull/176?

shadcn:
Sure

PR #251, opened 2019-08-08
[#250] Validate test connection by checking for a valid org.

### LINKED ISSUE #250: Edge URL connection not properly validated during connection test
**Describe the bug**
Related to https://github.com/apigee/apigee-devportal-kickstart-drupal/issues/245.  The Edge connection admin settings screen at `admin/config/apigee-edge/settings` validates the connection when the form is saved or the "Test connection" button is clicked. 

However, you can put in an invalid Edge endpoint URL and the system will not throw any errors, causing user to think the Edge endpoint URL is properly configured.

**To Reproduce**
1. Modify the `Apigee Edge endpoint` on the `Apigee Edge general settings` page to have the value `https://enterprise.apigee.com/platform/orgname/`
2. Click on 'Test connection'

**Actual behavior**
Message `Connection successful.` is displayed.

**Expected behavior**
An error should be displayed saying the connection is not valid.

**Additional context**
`\Drupal\apigee_edge\SDKConnector::testConnection` loads an organization entity but does not validate that is a valid org.  If you put in an invalid URL you will get back an organization entity w/no errors but all attributes will be `NULL`

**Version Info**
1.0

arlina-espinoza:
Fixes #250. It will check that the test connection returns a valid organization. This will need an additional test that mocks an invalid endpoint response, so a follow up issue will be needed after finalizing on a mocking test strategy.
Build: https://travis-ci.org/arlina-espinoza/apigee-edge-drupal/builds/569521122

### DIFF
--- a/src/KeyEntityFormEnhancer.php
+++ b/src/KeyEntityFormEnhancer.php
@@ -20,6 +20,7 @@
 
 namespace Drupal\apigee_edge;
 
+use Apigee\Edge\Exception\ApiException;
 use Apigee\Edge\Exception\ApiRequestException;
 use Apigee\Edge\Exception\OauthAuthenticationException;
 use Apigee\Edge\HttpClient\Plugin\Authentication\Oauth;
@@ -529,6 +530,12 @@ final class KeyEntityFormEnhancer {
               '%endpoint' => $key_type->getEndpoint($key),
             ]);
           }
+        }
+        elseif ($exception instanceof ApiException) {
+          $suggestion = $this->t('@fail_text The given endpoint (%endpoint) is incorrect or something is wrong with the connection.', [
+            '@fail_text' => $fail_text,
+            '%endpoint' => $key_type->getEndpoint($key),
+          ]);
         }
       }
     }
--- a/src/KeyEntityFormEnhancer.php
+++ b/src/KeyEntityFormEnhancer.php
@@ -20,10 +20,10 @@
 
 namespace Drupal\apigee_edge;
 
-use Apigee\Edge\Exception\ApiException;
 use Apigee\Edge\Exception\ApiRequestException;
 use Apigee\Edge\Exception\OauthAuthenticationException;
 use Apigee\Edge\HttpClient\Plugin\Authentication\Oauth;
+use Drupal\apigee_edge\Exception\InvalidArgumentException;
 use Drupal\apigee_edge\Exception\KeyProviderRequirementsException;
 use Drupal\apigee_edge\Plugin\EdgeKeyTypeInterface;
 use Drupal\apigee_edge\Plugin\KeyProviderRequirementsInterface;
@@ -531,7 +531,10 @@ final class KeyEntityFormEnhancer {
             ]);
           }
         }
-        elseif ($exception instanceof ApiException) {
+
+        // If SDKConnector::testConnection() fails to retrieve a valid org,
+        // then this exception is thrown.
+        elseif ($exception instanceof InvalidArgumentException) {
           $suggestion = $this->t('@fail_text The given endpoint (%endpoint) is incorrect or something is wrong with the connection.', [
             '@fail_text' => $fail_text,
             '%endpoint' => $key_type->getEndpoint($key),
--- a/src/SDKConnector.php
+++ b/src/SDKConnector.php
@@ -276,6 +276,10 @@ class SDKConnector implements SDKConnectorInterface {
       $oc = new OrganizationController($client);
       /* @var \Apigee\Edge\Api\Management\Entity\Organization $org */
       $org = $oc->load($credentials->getKeyType()->getOrganization($credentials->getKey()));
+
+      // Calling an invalid endpoint under some circumstances might return an
+      // empty organization object, so we check if it indeed loaded an org.
+      // @see https://github.com/apigee/apigee-edge-drupal/issues/250
       if (empty($org->id())) {
         throw new InvalidArgumentException('Failed to load a valid organization.');
       }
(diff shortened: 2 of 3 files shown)


src/SDKConnector.php
```diff
@@ -273,7 +274,11 @@ public function testConnection(KeyInterface $key = NULL) {
     try {
       // We use the original, non-decorated organization controller here.
       $oc = new OrganizationController($client);
-      $oc->load($credentials->getKeyType()->getOrganization($credentials->getKey()));
+      /* @var \Apigee\Edge\Api\Management\Entity\Organization $org */
+      $org = $oc->load($credentials->getKeyType()->getOrganization($credentials->getKey()));
+      if (empty($org->id())) {
+        throw new ApiException('Failed to load a valid organization.');
```
mxr576: ApiExceptions are thrown by the API client and not by the module. This is an invalid use of that exception here. Maybe let's use the module-specific InvalidArgumentException for this instead https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/src/Exception/InvalidArgumentException.php

mxr576 (changes requested):

src/SDKConnector.php
```diff
@@ -273,7 +274,11 @@ public function testConnection(KeyInterface $key = NULL) {
     try {
       // We use the original, non-decorated organization controller here.
       $oc = new OrganizationController($client);
-      $oc->load($credentials->getKeyType()->getOrganization($credentials->getKey()));
+      /* @var \Apigee\Edge\Api\Management\Entity\Organization $org */
+      $org = $oc->load($credentials->getKeyType()->getOrganization($credentials->getKey()));
+      if (empty($org->id())) {
```
mxr576, line 283: Please add a short comment about how this could occur because this is not expected behavior. Please mention the problematic API endpoint from #250

mxr576:
@arlina-espinoza Please also add test coverage to this bugfix to https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php

src/KeyEntityFormEnhancer.php
```diff
@@ -530,6 +531,12 @@ private function createSuggestion(\Exception $exception, KeyInterface $key): Mar
             ]);
           }
         }
+        elseif ($exception instanceof ApiException) {
```
cnovak: Can you add a comment on when this statement happens?

src/SDKConnector.php
cnovak: @mxr576 I did expect loading an org on an invalid endpoint would throw an exception, so if you agree I can put an issue in for [apigee/apigee-client-php](https://github.com/apigee/apigee-client-php) project.

cnovak (dismissed):

src/SDKConnector.php
mxr576: Well the endpoint is not "invalid" if it answers with 200 to request. Basically, you could validate everywhere if the provided API endpoint is actually and Apigee Edge API endpoint and not httpbin or something like that. Precisely, you could only validate if the endpoint returned the expected API reponse (check its structure? well we should not hardcode that) because there is no pattern for valid endpoints thanks for on-prem installs. I hope you see this would be an overkill. This is a user error and it is not the PHP API client's task to handle those.
Actually, we should not take care of this here either, because the form element has a description and it is very well documented what should be the value of the API endpoint. But if you would like to add a sanity check, this is the right place to do it.

mxr576: I wrote the previous comment on Friday afternoon so let me explain what did I mean by: 

> Basically, you could validate everywhere if the provided API endpoint is actually and Apigee Edge API endpoint 

In Drupal, you only need to check _in one place_ whether the API endpoint that accepted the request actually returned an organization id (is it enough to check?) to confirm this is a valid API endpoint.
In the API client, you would need to add a sanity check to all controller classes (developer, api product, stats, MINT reports, API packeges, etc) to validate if the API endpoint is correct. That would be an overkill. In my opinion, [the examples](https://github.com/apigee/apigee-client-php#usage-examples) that the PHP API client contains should be already enough for a start.

src/KeyEntityFormEnhancer.php
arlina-espinoza: Added comment to code:

>  If SDKConnector::testConnection() fails to retrieve a valid org, then this exception is thrown.

src/SDKConnector.php
arlina-espinoza: Sounds good, updated code to use `InvalidArgumentException`.

src/SDKConnector.php
arlina-espinoza: > Please add a short comment about how this could occur because this is not expected behavior. Please mention the problematic API endpoint from #250

Added a comment to the code:

> // Calling an invalid endpoint under some circumstances might return an
> // empty organization object, so we check if it indeed loaded an org.
> // @see https://github.com/apigee/apigee-edge-drupal/issues/250

cnovak:
Can you add a link to your tests passing @arlina-espinoza ?

src/SDKConnector.php
arlina-espinoza: @mxr576 The _invalid_ endpoint pointed out in #250 is invalid in that it returns HTTP 200 and a JSON response, just not the expected JSON response (not an organization, in this case). So the SDK is returning an empty organization object, and that is why we are checking that the loaded org is valid by having a non-empty ID.
I agree that it would be an overkill to add this kind of check everywhere, which is why I've only added it to `SDKConnector::testConnection()`.

@cnovak I would also have expected `apigee/apigee-client-php` to throw an exception for this scenario.

arlina-espinoza:
@mxr576 Here's the link to this branch build: https://travis-ci.org/arlina-espinoza/apigee-edge-drupal/builds/571497387
I keep getting failed tests in unrelated stuff from this PR, but then again, the main project is also failing on some of the same tests (https://travis-ci.org/apigee/apigee-edge-drupal).

shadcn (approved):
Tested. This looks good to me. LGTM.

PR #255, opened 2019-08-12
#243 Default team member role help text and context

### LINKED ISSUE #243: "Developer" default user role does show as an option while adding team members
**Describe the bug**
By default, the teams feature has 2 default roles- Admin and Developer. However while adding new users to the team, only admin and any new role you add shows up as a option but not developer.

**To Reproduce**
1. Go to teams and select an existing team.
2. Click on Add members
3. Scroll down and you will notice "developer" missing under roles.

**Expected behavior**
"developer" role should show as an option.

**Screenshots**
Screen Shot 2019-07-24 at 3 31 15 PM
Screen Shot 2019-07-24 at 3 31 33 PM

**Additional context**
None

**Version Info**
Apigee teams version: 8.x-1.0
Apigee edge version: 8.x-1.0

arlina-espinoza:
Fixes #243 . It adds help text to the team roles configuration page _(/admin/config/apigee-edge/team-settings/team_role)_:
image

It also adds a checkbox to the add/edit member form that will show the member role as checked and disabled:
image

### DIFF
--- a/modules/apigee_edge_teams/src/Form/AddTeamMembersForm.php
+++ b/modules/apigee_edge_teams/src/Form/AddTeamMembersForm.php
@@ -58,10 +58,10 @@ class AddTeamMembersForm extends TeamMembersFormBase {
    *   The entity type manager.
    */
   public function __construct(TeamMembershipManagerInterface $team_membership_manager, EntityTypeManagerInterface $entity_type_manager) {
+    parent::__construct($entity_type_manager);
+
     $this->teamMembershipManager = $team_membership_manager;
     $this->userStorage = $entity_type_manager->getStorage('user');
-    $this->teamRoleStorage = $entity_type_manager->getStorage('team_role');
-    $this->teamMemberRoleStorage = $entity_type_manager->getStorage('team_member_role');
   }
 
   /**
--- a/modules/apigee_edge_teams/src/Form/AddTeamMembersForm.php
+++ b/modules/apigee_edge_teams/src/Form/AddTeamMembersForm.php
@@ -25,7 +25,6 @@ use Drupal\apigee_edge_teams\Entity\TeamRole;
 use Drupal\apigee_edge_teams\Entity\TeamRoleInterface;
 use Drupal\apigee_edge_teams\TeamMembershipManagerInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
-use Drupal\Core\Form\FormBase;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Utility\Error;
 use Drupal\user\UserInterface;
@@ -34,14 +33,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
 /**
  * Add team members form.
  */
-class AddTeamMembersForm extends FormBase {
-
-  /**
-   * The team from the route.
-   *
-   * @var \Drupal\apigee_edge_teams\Entity\TeamInterface
-   */
-  protected $team;
+class AddTeamMembersForm extends TeamMembersFormBase {
 
   /**
    * The team membership manager service.
@@ -56,20 +48,6 @@ class AddTeamMembersForm extends FormBase {
    * @var \Drupal\user\UserStorageInterface
    */
   protected $userStorage;
-
-  /**
-   * Team role storage.
-   *
-   * @var \Drupal\apigee_edge_teams\Entity\Storage\TeamRoleStorageInterface
-   */
-  protected $teamRoleStorage;
-
-  /**
-   * Team member role storage.
-   *
-   * @var \Drupal\apigee_edge_teams\Entity\Storage\TeamMemberRoleStorage
-   */
-  protected $teamMemberRoleStorage;
 
   /**
    * AddTeamMemberForms constructor.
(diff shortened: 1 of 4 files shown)


modules/apigee_edge_teams/src/Form/EditTeamMemberForm.php
```diff
@@ -151,7 +154,10 @@ public function buildForm(array $form, FormStateInterface $form_state, TeamInter
   public function submitForm(array &$form, FormStateInterface $form_state) {
     $logger = $this->logger('apigee_edge_teams');
 
-    $selected_roles = array_filter($form_state->getValue('team_roles', []));
+    $selected_roles = array_filter($form_state->getValue('team_roles', []), function ($role) {
```
mxr576: Do not repeat yourself.

modules/apigee_edge_teams/src/Form/AddTeamMembersForm.php
```diff
@@ -136,6 +134,12 @@ public function buildForm(array $form, FormStateInterface $form_state, TeamInter
       '#options' => $role_options,
       '#multiple' => TRUE,
       '#required' => FALSE,
+      TeamRoleInterface::TEAM_MEMBER_ROLE => [
```
mxr576: This is not a valid render element array key.

mxr576 (changes requested):
The help text for the page looks fine, but this "display the member role as a disabled checkbox" thing, well I do not think it worth the effort and the additional overhead that it causes. @cnovak what do you think? Would not it be easier just adding a description to the element that also describes if no team role is selected the added user(s) still gets an implicit "member" role in the team?

modules/apigee_edge_teams/src/Form/AddTeamMembersForm.php
mxr576: This is what I would expect here if we would like to replace what Drupal core does.
```
    // Special handling for the inevitable "Authenticated user" role.
    $form['account']['roles'][RoleInterface::AUTHENTICATED_ID] = [
      '#default_value' => TRUE,
      '#disabled' => TRUE,
    ];
```

modules/apigee_edge_teams/src/Form/EditTeamMemberForm.php
```diff
@@ -118,6 +118,9 @@ public function buildForm(array $form, FormStateInterface $form_state, TeamInter
       '#default_value' => $current_role_options,
       '#multiple' => TRUE,
       '#required' => FALSE,
+      TeamRoleInterface::TEAM_MEMBER_ROLE => [
```
mxr576: DRY

modules/apigee_edge_teams/src/Form/AddTeamMembersForm.php
arlina-espinoza: :+1:

arlina-espinoza:
> The help text for the page looks fine, but this "display the member role as a disabled checkbox" thing, well I do not think it worth the effort and the additional overhead that it causes. @cnovak what do you think? Would not it be easier just adding a description to the element that also describes if no team role is selected the added user(s) still gets an implicit "member" role in the team?

@cnovak has added a comment [here](https://github.com/apigee/apigee-edge-drupal/issues/243#issuecomment-520525591) about how it would be good to follow the Drupal standard of showing the role as checked and disabled, as core does for the authenticated user role in the user add/edit forms.

modules/apigee_edge_teams/src/Form/EditTeamMemberForm.php
arlina-espinoza: A disabled checkbox would be filtered out just by `array_filter($form_state->getValue('team_roles', []));` . However,  a more throughout filter is needed in case there was some manual tampering of the form, i.e., if someone enabled the checkbox in the browser and submits it, then we need to filter it out. It's not the usual, but we can't trust user input.

arlina-espinoza:
To avoid some code duplication I've added a base team member role form (`TeamMembersFormBase`) with utility functions and `AddTeamMembersForm` and `EditTeamMemberForm` forms extend it.

modules/apigee_edge_teams/src/Form/TeamMembersFormBase.php
```diff
@@ -0,0 +1,87 @@
...
+
+namespace Drupal\apigee_edge_teams\Form;
+
+use Drupal\apigee_edge_teams\Entity\TeamRoleInterface;
+use Drupal\Core\Form\FormBase;
+
+/**
+ * Base team members form.
+ */
+abstract class TeamMembersFormBase extends FormBase {
```
shadcn, line 31: 👍 @arlina-espinoza

modules/apigee_edge_teams/src/Form/TeamMembersFormBase.php
```diff
@@ -0,0 +1,87 @@
...
+  protected $teamMemberRoleStorage;
+
+  /**
+   * Returns an array of team role options keyed by team role id.
+   *
+   * @return array
+   *   Role options.
+   */
+  protected function getRoleOptions() {
+    $role_options = array_reduce($this->teamRoleStorage->loadMultiple(), function (array $carry, TeamRoleInterface $role) {
```
shadcn, line 81: We should probably add a constructor here to avoid cases where `$this->teamRoleStorage` might be null. Or add an `abstract function getTeamRoleStorage(): TeamRoleStorageInterface;`.

shadcn (changes requested):
Added two comments. Otherwise this covers #243. Thanks @arlina-espinoza

modules/apigee_edge_teams/src/Form/TeamMembersFormBase.php
arlina-espinoza: @arshad I moved __construct() and create() to the base form to address this. Good catch :+1:

shadcn (approved):
Looks good to me. Thanks @arlina-espinoza

PR #257, opened 2019-08-12
[#244] Add team member form searches by email

### LINKED ISSUE #244: Unable to add developers to teams
**Describe the bug**
No result shows up when you try searching for developers on the add members page for teams.

**To Reproduce**
1. Go to teams page and select an existing team.
2. Click on add members
3. Try entering the name of a developer partially or completely 
4. No developer shows up

**Expected behavior**
Autocomplete should populate developer names matching the entered data.

**Screenshots**
Screen Shot 2019-07-24 at 3 38 16 PM

**Additional context**
None

**Version Info**
Apigee teams version: 8.x-1.0
Apigee edge version: 8.x-1.0

arlina-espinoza:
Fixes #244 . It makes the _"add team member"_ form _(teams/{team}/add-members)_ search by email, and display emails in the autocomplete dropdown.

### DIFF
--- a/modules/apigee_edge_teams/src/Plugin/EntityReferenceSelection/TeamMembersSelection.php
+++ b/modules/apigee_edge_teams/src/Plugin/EntityReferenceSelection/TeamMembersSelection.php
@@ -22,6 +22,7 @@ namespace Drupal\apigee_edge_teams\Plugin\EntityReferenceSelection;
 
 use Drupal\apigee_edge\Entity\Controller\DeveloperControllerInterface;
 use Drupal\apigee_edge_teams\TeamMembershipManagerInterface;
+use Drupal\Component\Utility\Html;
 use Drupal\Core\Database\Connection;
 use Drupal\Core\Entity\EntityManagerInterface;
 use Drupal\Core\Extension\ModuleHandlerInterface;
@@ -169,7 +170,13 @@ class TeamMembersSelection extends UserSelection {
    * {@inheritdoc}
    */
   protected function buildEntityQuery($match = NULL, $match_operator = 'CONTAINS') {
-    $query = parent::buildEntityQuery($match, $match_operator);
+    // Do not pass $match so that it does not add the "name" field to the query,
+    // as we will search by email instead.
+    $query = parent::buildEntityQuery(NULL, $match_operator);
+    if (isset($match)) {
+      $query->condition('mail', $match, $match_operator);
+    }
+
     $config = $this->getConfiguration();
     $team_name = $config['filter']['team'];
 
@@ -220,6 +227,23 @@ class TeamMembersSelection extends UserSelection {
     }
 
     return $query;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getReferenceableEntities($match = NULL, $match_operator = 'CONTAINS', $limit = 0) {
+    $options = parent::getReferenceableEntities($match, $match_operator, $limit);
+
+    $target_type = $this->getConfiguration()['target_type'];
+    $ids = array_keys($options[$target_type]);
+    $entities = $this->entityTypeManager->getStorage($target_type)->loadMultiple($ids);
+
+    /* @var \Drupal\user\UserInterface $entity */
+    foreach ($entities as $id => $entity) {
+      $options[$target_type][$id] = Html::escape($entity->getEmail());
+    }
+    return $options;
   }
 
 }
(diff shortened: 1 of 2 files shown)


modules/apigee_edge_teams/src/Plugin/EntityReferenceSelection/TeamMembersSelection.php
```diff
@@ -222,4 +229,21 @@ protected function buildEntityQuery($match = NULL, $match_operator = 'CONTAINS')
...
+  public function getReferenceableEntities($match = NULL, $match_operator = 'CONTAINS', $limit = 0) {
+    $options = parent::getReferenceableEntities($match, $match_operator, $limit);
+
+    $target_type = $this->getConfiguration()['target_type'];
+    $ids = array_keys($options[$target_type]);
+    $entities = $this->entityTypeManager->getStorage($target_type)->loadMultiple($ids);
+
+    /* @var \Drupal\user\UserInterface $entity */
+    foreach ($entities as $id => $entity) {
+      $options[$target_type][$id] = Html::escape($entity->getEmail());
```
mxr576: Why do you escape email addresses? Compared with a parent where a label() could be any kind of user input, I think we can safely assume that stored email addresses are always valid which mains they can not contain an XSS string.

modules/apigee_edge_teams/src/Plugin/EntityReferenceSelection/TeamMembersSelection.php
```diff
@@ -222,4 +229,21 @@ protected function buildEntityQuery($match = NULL, $match_operator = 'CONTAINS')
     return $query;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getReferenceableEntities($match = NULL, $match_operator = 'CONTAINS', $limit = 0) {
+    $options = parent::getReferenceableEntities($match, $match_operator, $limit);
```
mxr576: Do not call parent because if you do then you do all the stuff twice (including loading entities, which should be cached but if it does not it extremely slows down the autocomplete).

modules/apigee_edge_teams/src/Plugin/EntityReferenceSelection/TeamMembersSelection.php
```diff
@@ -169,7 +170,13 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
    * {@inheritdoc}
    */
   protected function buildEntityQuery($match = NULL, $match_operator = 'CONTAINS') {
-    $query = parent::buildEntityQuery($match, $match_operator);
+    // Do not pass $match so that it does not add the "name" field to the query,
+    // as we will search by email instead.
+    $query = parent::buildEntityQuery(NULL, $match_operator);
```
mxr576, line 174: As far as I see there is no better way than this only if you build your own implementation :(

modules/apigee_edge_teams/src/Plugin/EntityReferenceSelection/TeamMembersSelection.php
```diff
@@ -169,7 +170,13 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
...
+    if (isset($match)) {
```
mxr576: It is badly documented but `$match` could be `null` or a `string`. If it could be a string then it could be an empty string as well so I would handle that case here by calling `trim` on the input.

mxr576 (changes requested):

modules/apigee_edge_teams/src/Plugin/EntityReferenceSelection/TeamMembersSelection.php
arlina-espinoza: The parent implementation is not statically caching the results, so I moved the implementation into this class so it's not calling `parent` anymore.

modules/apigee_edge_teams/src/Plugin/EntityReferenceSelection/TeamMembersSelection.php
arlina-espinoza: Agreed.

modules/apigee_edge_teams/src/Plugin/EntityReferenceSelection/TeamMembersSelection.php
arlina-espinoza: Good catch.

modules/apigee_edge_teams/src/Plugin/EntityReferenceSelection/TeamMembersSelection.php
```diff
@@ -22,6 +22,7 @@
 
 use Drupal\apigee_edge\Entity\Controller\DeveloperControllerInterface;
 use Drupal\apigee_edge_teams\TeamMembershipManagerInterface;
+use Drupal\Component\Utility\Html;
```
shadcn: Import is never used. Can be removed.

shadcn (approved):
Added one comment. Otherwise looks good. Thanks @arlina-espinoza

modules/apigee_edge_teams/src/Plugin/EntityReferenceSelection/TeamMembersSelection.php
arlina-espinoza: Removed statement. Thanks for pointing it out. :wink:

shadcn (approved):
Thanks @arlina-espinoza LGTM

Issue #254, opened 2019-08-12
Changes to Edge connection settings page 


arunz6161:
**Is your feature request related to a problem? Please describe.**
There are a couple of changes needed to make the  Edge connection configuration simpler and error free. 

1. Apigee Edge Endpoint
The config page suggests that this value be left empty but does not mean when it should be left empty or be filled.

Value should be left empty ( so that the default can be applied) when one is pointing to the an org on [Public Cloud ](https://docs.apigee.com/api-platform/get-started/what-apigee-edge#cloudvonprem).

For a Cloud based installation of Edge, the URL is:
https://api.enterprise.apigee.com/v1
For a Private Cloud installation, it is in the form:
http://ms_IP_or_DNS:8080/v1 or
https://ms_IP_or_DNS:TLSport/v1

2. The authorization server endpoint for a SAML enabled edge org is in the form:

https:// {zonename}.login.apigee.com/oauth/token

**Describe the solution you'd like**
Same as above

**Describe alternatives you've considered**
None

**Additional context**
Attaching screenshot
![Screen Shot 2019-08-12 at 11 50 08 AM](https://user-images.githubusercontent.com/4381612/62897827-d99f6200-bd08-11e9-8eca-569f5bbac512.png)

PR #263, opened 2019-09-19
[#261] Allow authentication using a username

### LINKED ISSUE #261: Portal not compatible with External Auth on Edge
**Describe the bug**
When an Edge org (private cloud) is using External Auth, they are unable to connect their portal to Edge. The configuration page requires and email address, where External Auth requires a username (no @domain.com)

**To Reproduce**
Attempt to connect to an Edge organization using External Auth with a username (not email) and password. The portal will throw an error "Invalid email address" error. When attempting to connect with an email, you will receive a 401 error as Edge can only authenticate with a username.

**Expected behavior**
Dev portal should be able to connect to an Edge org with just a username - not only an email address.

**Version Info**
8.x-1.0

arlina-espinoza:
Fixes #261 . It changes the username field to a textfield, and does email validation only if using the default endpoint (for public cloud).

### DIFF
--- a/src/KeyEntityFormEnhancer.php
+++ b/src/KeyEntityFormEnhancer.php
@@ -94,7 +94,7 @@ final class KeyEntityFormEnhancer {
    *
    * @var \Drupal\Component\Utility\EmailValidatorInterface
    */
-  protected $emailValidator;
+  private $emailValidator;
 
   /**
    * KeyEntityFormEnhancer constructor.
@@ -512,7 +512,7 @@ final class KeyEntityFormEnhancer {
 
         // If on public cloud (using the default endpoint), the username should
         // be an email.
-        if ($key_type->getEndpointType($key) == 'default' && !$this->emailValidator->isValid($key_type->getUsername($key))) {
+        if ($key_type->getEndpointType($key) === EdgeKeyTypeInterface::EDGE_ENDPOINT_TYPE_DEFAULT && !$this->emailValidator->isValid($key_type->getUsername($key))) {
           $suggestion = $this->t('@fail_text The organization username should be a valid email.', [
             '@fail_text' => $fail_text,
           ]);
--- a/src/KeyEntityFormEnhancer.php
+++ b/src/KeyEntityFormEnhancer.php
@@ -29,6 +29,7 @@ use Drupal\apigee_edge\Plugin\EdgeKeyTypeInterface;
 use Drupal\apigee_edge\Plugin\KeyProviderRequirementsInterface;
 use Drupal\apigee_edge\Plugin\KeyType\ApigeeAuthKeyType;
 use Drupal\Component\Render\MarkupInterface;
+use Drupal\Component\Utility\EmailValidatorInterface;
 use Drupal\Component\Utility\Random;
 use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\DependencyInjection\DependencySerializationTrait;
@@ -89,6 +90,13 @@ final class KeyEntityFormEnhancer {
   private $configFactory;
 
   /**
+   * The email validator.
+   *
+   * @var \Drupal\Component\Utility\EmailValidatorInterface
+   */
+  protected $emailValidator;
+
+  /**
    * KeyEntityFormEnhancer constructor.
    *
    * @param \Drupal\apigee_edge\SDKConnectorInterface $connector
@@ -99,12 +107,15 @@ final class KeyEntityFormEnhancer {
    *   The entity type manager serivce.
    * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
    *   The config factory.
+   * @param \Drupal\Component\Utility\EmailValidatorInterface $email_validator
+   *   The email validator.
    */
-  public function __construct(SDKConnectorInterface $connector, OauthTokenStorageInterface $oauth_token_storage, EntityTypeManagerInterface $entity_type_manager, ConfigFactoryInterface $config_factory) {
+  public function __construct(SDKConnectorInterface $connector, OauthTokenStorageInterface $oauth_token_storage, EntityTypeManagerInterface $entity_type_manager, ConfigFactoryInterface $config_factory, EmailValidatorInterface $email_validator) {
     $this->connector = $connector;
     $this->entityTypeManager = $entity_type_manager;
     $this->oauthTokenStorage = $oauth_token_storage;
     $this->configFactory = $config_factory;
+    $this->emailValidator = $email_validator;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


cnovak (changes requested):
Looks good, but can you add a test for this service to validate the functionality?

arlina-espinoza:
@cnovak I've added a functional test, and also moved the validation from the main form submit to the _suggestions_, so that it is also checked when using the "test connection" button. Please review again. :slightly_smiling_face:

src/KeyEntityFormEnhancer.php
```diff
@@ -88,6 +89,13 @@ final class KeyEntityFormEnhancer {
    */
   private $configFactory;
 
+  /**
+   * The email validator.
+   *
+   * @var \Drupal\Component\Utility\EmailValidatorInterface
+   */
+  protected $emailValidator;
```
mxr576: Using `protected` in a `final` class does not make any sense.

src/KeyEntityFormEnhancer.php
```diff
@@ -498,10 +509,20 @@ private function createSuggestion(\Exception $exception, KeyInterface $key): Mar
...
       // the MGMT server returns HTTP 500 with an error instead of HTTP 401.
       if ($exception->getCode() === 401 || ($exception->getCode() === 500 && $exception->getEdgeErrorCode() === 'usersandroles.SsoInternalServerError')) {
-        $suggestion = $this->t('@fail_text The given username (%username) or password is incorrect.', [
-          '@fail_text' => $fail_text,
-          '%username' => $key_type->getUsername($key),
-        ]);
+
+        // If on public cloud (using the default endpoint), the username should
+        // be an email.
+        if ($key_type->getEndpointType($key) == 'default' && !$this->emailValidator->isValid($key_type->getUsername($key))) {
```
mxr576: Use strict comparison

src/Plugin/EdgeKeyTypeBase.php
```diff
@@ -62,6 +62,13 @@ public function getEndpoint(KeyInterface $key): string {
     return $key->getKeyValues()['endpoint'] ?? Client::DEFAULT_ENDPOINT;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getEndpointType(KeyInterface $key): string {
+    return $key->getKeyValues()['endpoint_type'] ?? 'default';
```
mxr576: I would probably consider introducing constants for `default` and `custom` in EdgeKeyTypeInterface because I keep seeing these magic words appearing in different places of the code. Ex.: in ApigeeAuthKeyForm.

mxr576 (changes requested):
Also `$values['endpoint_type']` is alway set [here](https://github.com/apigee/apigee-edge-drupal/blob/4dc2cb8ff1aaaf4c797660e7a1ca9ba849946d29/src/Plugin/KeyInput/ApigeeAuthKeyInput.php#L112) because of this [line](https://github.com/apigee/apigee-edge-drupal/blob/4dc2cb8ff1aaaf4c797660e7a1ca9ba849946d29/src/Plugin/KeyInput/ApigeeAuthKeyInput.php#L66).

src/KeyEntityFormEnhancer.php
mxr576: Does this change compatible with the environment variables API key storage? Because maybe the new entity_type field's definition is missing from [here](https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/src/Plugin/KeyType/ApigeeAuthKeyType.php#L42-L75) and thanks for that the environment credentials storage [does not expose](https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/src/Plugin/KeyProvider/EnvironmentVariablesKeyProvider.php#L115) an environment variable for this field. IOW, what happens if I set the value of the APIGEE_EDGE_USERNAME environment variable to an invalid email and the APIGEE_EDGE_ENDPOINT to a private org (without being able the set the endpoint type so it remains "default"). If I click on "Test connection, does it generate a suggestion that the username should be an email address?

If this is real bug then probably the same issue occurs for the `authorization_server_type` introduced in #254.

cnovak: @mxr576 can you explain why in this case? I am guessing you are thinking performance, but other than that I don't see an issue w/type juggling.

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -87,9 +87,9 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
       '#attributes' => ['autocomplete' => 'off'],
     ];
     $form['username'] = [
-      '#type' => 'email',
+      '#type' => 'textfield',
       '#title' => $this->t('Username'),
-      '#description' => $this->t("Organization user's email address that is used for authenticating with the endpoint."),
+      '#description' => $this->t("Organization username that is used for authenticating with the endpoint."),
```
cnovak:
```suggestion
      '#description' => $this->t("Apigee user's email address or identity provider username that is used for authenticating with the endpoint."),
```
Don't think we should change to username because you don't user the username if it is an Apigee User:
Screen Shot 2019-09-25 at 9 07 37 AM

cnovak (changes requested):
Need to address @mxr576 comments, and one wording change I put in.  Make sure to validate environment variables work properly.

cnovak:
@mxr576 not sure what you mean here, is this supposed to be commenting on a specific code line?:
> Also `$values['endpoint_type']` is alway set [here](https://github.com/apigee/apigee-edge-drupal/blob/4dc2cb8ff1aaaf4c797660e7a1ca9ba849946d29/src/Plugin/KeyInput/ApigeeAuthKeyInput.php#L112) because of this [line](https://github.com/apigee/apigee-edge-drupal/blob/4dc2cb8ff1aaaf4c797660e7a1ca9ba849946d29/src/Plugin/KeyInput/ApigeeAuthKeyInput.php#L66).

src/Plugin/EdgeKeyTypeBase.php
arlina-espinoza: Added the constants and updated appearances of `default` and `custom` to use them.

src/KeyEntityFormEnhancer.php
arlina-espinoza: Changed to strict comparison.

src/KeyEntityFormEnhancer.php
arlina-espinoza: Changed.

src/KeyEntityFormEnhancer.php
arlina-espinoza: I've updated the `EdgeKeyTypeBase::getEndpointType()` method so that if the `endpoint_type` property isn't defined, it still returns the correct type, that way we can avoid adding more environment variables.

https://github.com/apigee/apigee-edge-drupal/blob/ac7ba5cf219ac0807685af4f4acc4b652b444e80/src/Plugin/EdgeKeyTypeBase.php#L73-L77

`authorization_server_type` isn't an issue, because it isn't used outside of the `ApigeeAuthKeyInput` class/form.

arlina-espinoza:
I've addressed the recommended changes @cnovak / @mxr576, please take a look.

cnovak (approved):
LGTM

PR #266, opened 2019-09-20
[#262] Fix app analytics submit redirect

### LINKED ISSUE #262: Unable to retreive Analytics, Always showing page not found
pagenot found
After apigee edge connection setup
Below are working
1. Apps
2. Products
3. Catalogs

Not working is
1. Analytics. whenever i search for analytics its showing me page not found. Sometimes its showming me no errors other than page not found sometimes i got as error on [transaltion.[](url)](url)

arlina-espinoza:
Fixes #262 .

**Current bug**
When submitting an app analytics form (eg: _user/2/apps/test_app_1/analytics_), a redirect is issued that prepends the site's base path to the URL. If the site is installed under the root directory (like _example.com/_), then the redirect works, but if the site is under a subdirectory (like _/example.com/drupal/folder/_), then the redirect goes to a page not found _(example.com/drupal/folder/drupal/folder/user/2/apps/test_app_1/analytics)_.

**Expected behavior**
This PR fixes the redirect so it goes to the correct URL (ie, _example.com/drupal/folder/user/2/apps/test_app_1/analytics_).

### DIFF
--- a/src/Form/AppAnalyticsFormBase.php
+++ b/src/Form/AppAnalyticsFormBase.php
@@ -452,7 +452,7 @@ abstract class AppAnalyticsFormBase extends FormBase {
         'until' => $form_state->getValue('until')->getTimeStamp(),
       ],
     ];
-    $form_state->setRedirect('<current>', [],  $options);
+    $form_state->setRedirect('<current>', [], $options);
   }
 
   /**
--- a/src/Form/AppAnalyticsFormBase.php
+++ b/src/Form/AppAnalyticsFormBase.php
@@ -445,14 +445,16 @@ abstract class AppAnalyticsFormBase extends FormBase {
    * {@inheritdoc}
    */
   public function submitForm(array &$form, FormStateInterface $form_state) {
-    $url = $this->urlGenerator->generateFromRoute('<current>', [], [
+    $route = \Drupal::routeMatch();
+    $parameters = $route->getRawParameters()->all();
+    $options = [
       'query' => [
         'metric' => $form_state->getValue('metrics'),
         'since' => $form_state->getValue('since')->getTimeStamp(),
         'until' => $form_state->getValue('until')->getTimeStamp(),
       ],
-    ]);
-    $form_state->setRedirectUrl(Url::fromUri('internal:' . $url));
+    ];
+    $form_state->setRedirect($route->getRouteName(), $parameters, $options);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


cnovak (approved):
LGTM

src/Form/AppAnalyticsFormBase.php
```diff
@@ -445,14 +445,16 @@ protected function generateResponse(array &$form, AppInterface $app, string $met
    * {@inheritdoc}
    */
   public function submitForm(array &$form, FormStateInterface $form_state) {
-    $url = $this->urlGenerator->generateFromRoute('<current>', [], [
+    $route = \Drupal::routeMatch();
```
mxr576: Pls dependency inject the route matcher service.

mxr576 (changes requested):
Tbh, this fix looks odd to me because all Drupal core forms support subdirs but none of them using this `getRawParameters()` parameters trick which indicates to me that there should be a better way to resolve this problem. 
image

cnovak:
I see `NodeForm` `preview()` following this pattern but only to add parameters, not to pass parameters through to the redirect. Maybe you do not need to set the parameters & options again when using `$form_state->setRedirect`

> Tbh, this fix looks odd to me because all Drupal core forms support subdirs but none of them using this `getRawParameters()` parameters trick which indicates to me that there should be a better way to resolve this problem.
> image

src/Form/AppAnalyticsFormBase.php
cnovak: I wasn't sure if we need to inject services when using them for one off situations like this, but I do agree with @mxr576 after looking at [coding standards](https://www.drupal.org/docs/8/api/services-and-dependency-injection/services-and-dependency-injection-in-drupal-8#s-using-dependency-injection)

> Dependency injection is the preferred method for accessing and using services in Drupal 8 and should be used whenever possible.

cnovak (changes requested):
1. Add service through dependency injection
2. Verify need to pass parameters into $form_state->setRedirect

arlina-espinoza:
@cnovak Pushed an update to this PR. Figured out that setting the form redirect directly to` "<current>" `avoids having to pull out the route name and route parameters altogether, so there is no need to inject another service. Please review again.

>     1. Add service through dependency injection
> 
>     2. Verify need to pass parameters into $form_state->setRedirect

src/Form/AppAnalyticsFormBase.php
arlina-espinoza: This should not be needed  - see https://github.com/apigee/apigee-edge-drupal/pull/266#issuecomment-534248353.

cnovak (approved):
Looks good

PR #270, opened 2019-10-01
TravisCI: Update phpcs version

cnovak:
* Use latest phpcs version to allow line by line disable of sniffs
* Removed mxr576/docker-phpcs from docker-compose, no longer used
* Fixed phpcs issues in  EdgeExceptionSubscriberTest


arlina-espinoza (approved):
LGTM, thanks :)

PR #271, opened 2019-10-02
TravisCI: Run PHP Code Sniffer only once

cnovak:
Do not run the PHP Code Sniffer on every test, just once before tests begin.


cnovak:
The proof is in the [pudding](https://travis-ci.org/apigee/apigee-edge-drupal/builds/592646913?utm_source=github_status&utm_medium=notification)

arlina-espinoza (approved):
Thanks @cnovak :+1:

PR #273, opened 2019-10-03
[Bug][High priority] Do not try to update read-only API keys

mxr576:
The `apigee_edge_update_8002()` is broken if the environment credential storage is being used as an Apigee Edge credential storage. That storage is read-only, so calling `$key->save()` operation on the key throws a `KeyValueNotSetException` exception: `The selected key provider does not support setting a key value.` See: `\Drupal\key\Entity\Key::preSave()`

I am also referring to my original question on a related PR about whether these changes around the key field types are working properly if the environment credential storage is being used: https://github.com/apigee/apigee-edge-drupal/pull/263#discussion_r328165113

Please perform extra testing to verify the newly introduced functionality in 8.x-1.1 works properly with the environmental credential storage as well. For example, connect to a "special" org that accepts usernames instead of email address for authentication and verify the displayed error messages and suggestions on the Connection setting form are valid when you are trying to connect with:
* an invalid username
* with an invalid username that is an email address

### DIFF
--- a/apigee_edge.install
+++ b/apigee_edge.install
@@ -27,6 +27,7 @@ use Drupal\apigee_edge\Plugin\KeyType\ApigeeAuthKeyType;
 use Drupal\apigee_edge\OauthTokenFileStorage;
 use Drupal\Component\Serialization\Json;
 use Drupal\Core\Url;
+use Drupal\key\Plugin\KeyProviderSettableValueInterface;
 use Drupal\user\RoleInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


cnovak:
@arlina-espinoza I believe you checked everything as part of #263 , but let's make sure

arlina-espinoza (approved):
Good catch for the update hook. The keys `endpoint_type` and `authorization_server_type` are basically helpers for a nicer UI on the form, and they can be deduced from the actual `endpoint` and `authorization_server` keys. As for the testing scenarios:

- Empty `endpoint` or `endpoint_type` == default: This is the only use case when we validate the username is formatted as an email. A test was added as well to https://github.com/apigee/apigee-edge-drupal/blob/ce670d198ce48c39f9bf17d1a8491671904774ce/tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php#L285-L289

- If not using the default endpoint, we cannot assume a username is (or not) in the form of an email, and the default error message is shown (ie, _"the given username (%username) or password is incorrect"_) - this was already covered in the tests.

Issue #268, opened 2019-10-01
Change apigee settings page to accommodate hybrid org


arunz6161:
Change apigee settings page to accommodate hybrid org- by adding hybrid as an option which defaults to https://apigee.googleapis.com/v1/ as the endpoint and allow user to upload service account key (json key) for authentication.

arlina-espinoza:
Depends on https://github.com/apigee/apigee-client-php/issues/64.

arlina-espinoza:
Resolved by #275 .

PR #275, opened 2019-10-08
[#268] Change apigee settings page to accommodate hybrid org

arlina-espinoza:
@cnovak / @arunz6161  Let me know your thoughts on the changes that would be done to the Edge settings page. The idea is to only present the relevant options for each Apigee instance type (public, private or hybrid).

If the user selects "**public cloud**", then there is not an option to configure a custom endpoint - it will use the default (https://api.enterprise.apigee.com/v1). It does give the user the option to configure authentication type (HTTP basic or oauth):
image

If the user selects "**private cloud**", they must fill in the "Apigee Edge endpoint" textfield - again, they have the option to configure authentication type (HTTP basic or oauth):
image

If the user selects "**hybrid cloud**", then the only options are the organization name, and the textarea to paste the contents of the JSON account key:
image


arunz6161:
@arlina-espinoza this looks good. We do need to clarify more on the userrole of service account of which we are asking the account key. Am working on this and will update the thread soon.

cnovak:
Agree with @arunz6161, this looks like a great approach to the settings UX

phpcs.xml.dist
```diff
@@ -18,5 +18,9 @@
   <rule ref="Drupal.NamingConventions.ValidVariableName.LowerCamelName">
     <severity>0</severity>
   </rule>
+  <!-- Deprecation notices do not need a link to drupal nodes. -->
```
cnovak:
```suggestion
  <!-- Deprecation notices do not need a link to drupal nodes since we use GitHub issue queue. -->
```

src/Plugin/EdgeKeyTypeInterface.php
```diff
@@ -45,13 +65,17 @@ interface EdgeKeyTypeInterface extends KeyTypeMultivalueInterface, KeyTypeAuthen
    * The endpoint type for default.
    *
    * @var string
+   *
+   * @deprecated
```
cnovak: Can you add why this is deprecated since we don't have an issue to link to, or maybe description and link to the issue #268

src/Plugin/EdgeKeyTypeInterface.php
```diff
@@ -45,13 +65,17 @@ interface EdgeKeyTypeInterface extends KeyTypeMultivalueInterface, KeyTypeAuthen
...
    */
   const EDGE_ENDPOINT_TYPE_DEFAULT = 'default';
 
   /**
    * The endpoint type for custom.
    *
    * @var string
+   *
+   * @deprecated
```
cnovak: Can you add why this is deprecated since we don't have an issue to link to, or maybe description and link to the issue #268

arlina-espinoza:
This is ready for review. It updates the Hybrid settings page and authentication method to make it compatible with Hybrid.
This PR depends on [apigee/apigee-client-php #63](https://github.com/apigee/apigee-client-php/pull/71). Instructions for setting up a test environment using this PR and the php-client's branch and dependencies can be found in this gist:
https://gist.github.com/arlina-espinoza/e6b299d64f781e89bf54dc5596be1de2

apigee_edge.install
```diff
@@ -27,6 +27,7 @@ use Drupal\apigee_edge\Plugin\KeyType\ApigeeAuthKeyType;
 use Drupal\apigee_edge\OauthTokenFileStorage;
 use Drupal\Component\Serialization\Json;
 use Drupal\Core\Url;
+use Drupal\key\Plugin\KeyProviderSettableValueInterface;
```
cnovak: Not used in file

src/HybridAuthentication.php
```diff
@@ -0,0 +1,41 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+namespace Drupal\apigee_edge;
```
cnovak: Seems like we are adding a lot of credentials/authentication classes in the root namespace, how about we put them in a namespace such as `authentication` or `connector`?

src/HybridCredentials.php
```diff
@@ -0,0 +1,52 @@
...
+ * with this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+namespace Drupal\apigee_edge;
+
+use Drupal\apigee_edge\Exception\InvalidArgumentException;
+use Drupal\apigee_edge\Plugin\EdgeKeyTypeInterface;
+use Drupal\key\KeyInterface;
+use Http\Message\Authentication;
```
cnovak: import not used

src/KeyEntityFormEnhancer.php
```diff
@@ -246,9 +248,18 @@ public function alterForm(array &$form, FormStateInterface $form_state): void {
         ],
         '#states' => [
           'enabled' => [
-            ':input[name="key_input_settings[password]"]' => ['filled' => TRUE],
-            ':input[name="key_input_settings[organization]"]' => ['filled' => TRUE],
-            ':input[name="key_input_settings[username]"]' => ['filled' => TRUE],
+            [
```
cnovak, line 252: The "send request" button is greyed out whenever you put text in the `account_json_key` field
send-request-disabled

src/KeyEntityFormEnhancer.php
```diff
@@ -581,16 +598,27 @@ private function createSuggestion(\Exception $exception, KeyInterface $key): Mar
   private function createDebugText(\Exception $exception, KeyInterface $key): string {
     $key_type = $key->getKeyType();
 
-    $credentials = !($key_type instanceof EdgeKeyTypeInterface) ? [] : [
-      'endpoint' => $key_type->getEndpoint($key),
-      'organization' => $key_type->getOrganization($key),
-      'username' => $key_type->getUsername($key),
-    ];
+    if ($key_type instanceof EdgeKeyTypeInterface
```
cnovak: This logic is hard to read, can we build up the response instead? Such as set `$credentials` to empty array, then add values depending on key type instance?

src/KeyEntityFormEnhancer.php
```diff
@@ -581,16 +598,27 @@ private function createSuggestion(\Exception $exception, KeyInterface $key): Mar
...
+      && $key_type->getInstanceType($key) === EdgeKeyTypeInterface::INSTANCE_TYPE_HYBRID) {
+      $credentials = [
+        'endpoint' => $key_type->getEndpoint($key),
+        'organization' => $key_type->getOrganization($key),
+      ];
+      $keys = [];
```
cnovak: We should still be able to output key type and auth type in debug if it is hybrid

src/Plugin/EdgeKeyTypeBase.php
```diff
@@ -49,6 +50,10 @@ public function unserialize($value) {
    * {@inheritdoc}
    */
   public function getAuthenticationType(KeyInterface $key): string {
+    if ($this->getInstanceType($key) === EdgeKeyTypeInterface::INSTANCE_TYPE_HYBRID) {
+      return 'hybrid';
```
cnovak: It seems that the authentication type is `jwt` since you can use [JWT directly](https://developers.google.com/identity/protocols/OAuth2ServiceAccount#jwt-auth) or OAuth.

src/Plugin/EdgeKeyTypeBase.php
```diff
@@ -59,24 +64,40 @@ public function getAuthenticationType(KeyInterface $key): string {
    * {@inheritdoc}
    */
   public function getEndpoint(KeyInterface $key): string {
+    if ($this->getInstanceType($key) === EdgeKeyTypeInterface::INSTANCE_TYPE_HYBRID) {
+      return ClientInterface::HYBRID_ENDPOINT;
+    }
     return $key->getKeyValues()['endpoint'] ?? Client::DEFAULT_ENDPOINT;
```
cnovak: Changing `Client::DEFAULT_ENDPOINT` to be `Client::PUBLIC_CLOUD_ENDPOINT` should make things more clear since there isn't a default endpoint anymore.

src/Plugin/EdgeKeyTypeBase.php
```diff
@@ -59,24 +64,40 @@ public function getAuthenticationType(KeyInterface $key): string {
    * {@inheritdoc}
    */
   public function getEndpoint(KeyInterface $key): string {
+    if ($this->getInstanceType($key) === EdgeKeyTypeInterface::INSTANCE_TYPE_HYBRID) {
```
cnovak, line 67: The logic here seems to be more clear if we stick to the way the form validation works:

If instance type is:
1. hybrid, return hybrid endpoint
2. public, return public endpoint
3. private, return `$key->getKeyValues()['endpoint']`

src/KeyEntityFormEnhancer.php
```diff
@@ -555,7 +572,7 @@ private function createSuggestion(\Exception $exception, KeyInterface $key): Mar
 
         // If SDKConnector::testConnection() fails to retrieve a valid org,
         // then this exception is thrown.
-        elseif ($exception instanceof InvalidArgumentException) {
+        elseif ($exception instanceof InvalidArgumentException || $exception instanceof NotEncodableValueException) {
```
cnovak: How do you get the system to throw a `NotEncodableValueException`?

src/KeyEntityFormEnhancer.php
```diff
@@ -581,16 +598,27 @@ private function createSuggestion(\Exception $exception, KeyInterface $key): Mar
...
-    ];
+    if ($key_type instanceof EdgeKeyTypeInterface
+      && $key_type->getInstanceType($key) === EdgeKeyTypeInterface::INSTANCE_TYPE_HYBRID) {
+      $credentials = [
+        'endpoint' => $key_type->getEndpoint($key),
+        'organization' => $key_type->getOrganization($key),
+      ];
+      $keys = [];
+    }
+
```
cnovak, line 626: extra space

src/KeyEntityFormEnhancer.php
```diff
@@ -581,16 +598,27 @@ private function createSuggestion(\Exception $exception, KeyInterface $key): Mar
...
+        'endpoint' => $key_type->getEndpoint($key),
+        'organization' => $key_type->getOrganization($key),
+        'username' => $key_type->getUsername($key),
+      ];
 
-    $keys = [
-      'auth_type' => ($key_type instanceof EdgeKeyTypeInterface) ? $key_type->getAuthenticationType($key) : 'invalid credentials',
-      'key_provider' => get_class($key->getKeyProvider()),
-    ];
+      $keys = [
```
cnovak: The `$keys` array should still be populated with Auth Type and Key Provider for hybrid

src/Plugin/EdgeKeyTypeInterface.php
```diff
@@ -77,17 +101,29 @@ public function getAuthenticationType(KeyInterface $key): string;
...
   public function getEndpoint(KeyInterface $key): string;
 
+  /**
+   * Gets the instance type (public, private or hybrid).
+   *
+   * @param \Drupal\key\KeyInterface $key
+   *   The key entity.
+   *
+   * @return string
+   *   The instance type.
```
cnovak:
```suggestion
   *   The instance type, either `public`, `private` or `hybrid`.
```

src/Plugin/EdgeKeyTypeBase.php
```diff
@@ -128,4 +149,16 @@ public function getClientSecret(KeyInterface $key): string {
     return $key->getKeyValues()['client_secret'] ?? Oauth::DEFAULT_CLIENT_SECRET;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getAccountKey(KeyInterface $key): array {
```
cnovak, line 158: Should Hybrid be its own class since the attributes are so different? We have username/pass that don't make sense for Hybrid but are stored along w/the key data, which may be another side effect of using the same class.

src/Plugin/EdgeKeyTypeInterface.php
```diff
@@ -157,4 +193,15 @@ public function getClientId(KeyInterface $key): string;
...
+  /**
+   * Return the JSON account key decoded as an array.
+   *
+   * @param \Drupal\key\KeyInterface $key
+   *   The key entity.
+   *
+   * @return array
+   *   The account key as an array.
+   */
+  public function getAccountKey(KeyInterface $key): array;
```
cnovak, line 221: See my comment on implementation about breaking Hybrid into its own class

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -63,9 +62,32 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
...
+      ':input[name="key_input_settings[instance_type]"]' => ['value' => EdgeKeyTypeInterface::INSTANCE_TYPE_PRIVATE],
+    ];
+    $state_for_hybrid = [
+      ':input[name="key_input_settings[instance_type]"]' => ['value' => EdgeKeyTypeInterface::INSTANCE_TYPE_HYBRID],
+    ];
+
+    $form['instance_type'] = [
+      '#type' => 'radios',
+      '#title' => $this->t('Apigee instance type'),
+      '#description' => $this->t('Select the Apigee instance type you are connecting to. More information can be found in the <a href="@link" target="_blank">docs</a>.', [
```
cnovak:
```suggestion
      '#description' => $this->t('Select the Apigee instance type you are connecting to. More information can be found in the <a href="@link" target="_blank">Apigee documentation</a>.', [
```

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -63,9 +62,32 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
...
+        '@link' => 'https://docs.apigee.com/api-platform/get-started/what-apigee-edge#cloudvonprem',
```
cnovak: Assuming that @jbennett13 gave you this as the URL to use that will be updated w/hybrid info in the future

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -90,51 +115,50 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
       '#type' => 'textfield',
       '#title' => $this->t('Username'),
       '#description' => $this->t("Apigee user's email address or identity provider username that is used for authenticating with the endpoint."),
-      '#required' => TRUE,
       '#default_value' => $values['username'] ?? '',
       '#attributes' => ['autocomplete' => 'off'],
+      '#states' => [
+        'visible' => [$state_for_public, 'xor', $state_for_private],
```
cnovak: Wouldn't these states work with `or` instead of `xor`? Thinking this and other fields should be visible if private `or` public, those are already mutually exclusive.

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -90,51 +115,50 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
...
-      '#type' => 'radios',
-      '#required' => TRUE,
-      '#default_value' => $values['endpoint_type'],
-      '#options' => [
-        EdgeKeyTypeInterface::EDGE_ENDPOINT_TYPE_DEFAULT => $this->t('Default'),
-        EdgeKeyTypeInterface::EDGE_ENDPOINT_TYPE_CUSTOM => $this->t('Custom'),
+    $form['account_json_key'] = [
+      '#type' => 'textarea',
+      '#title' => $this->t('Account key'),
+      '#description' => $this->t("Paste the contents of the account key JSON file."),
```
cnovak:
```suggestion
      '#description' => $this->t("Paste the contents of the GCP service account key JSON file."),
```

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -90,51 +115,50 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
...
-      '#title' => $this->t('Apigee Edge endpoint'),
-      '#type' => 'radios',
-      '#required' => TRUE,
-      '#default_value' => $values['endpoint_type'],
-      '#options' => [
-        EdgeKeyTypeInterface::EDGE_ENDPOINT_TYPE_DEFAULT => $this->t('Default'),
-        EdgeKeyTypeInterface::EDGE_ENDPOINT_TYPE_CUSTOM => $this->t('Custom'),
+    $form['account_json_key'] = [
+      '#type' => 'textarea',
+      '#title' => $this->t('Account key'),
```
cnovak:
```suggestion
      '#title' => $this->t('GCP service account key'),
```

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -218,13 +263,20 @@ public function processSubmittedKeyValue(FormStateInterface $form_state) {
...
-    if (empty($input_values['endpoint_type']) || $input_values['endpoint_type'] == EdgeKeyTypeInterface::EDGE_ENDPOINT_TYPE_DEFAULT) {
+    if ($input_values['instance_type'] == EdgeKeyTypeInterface::INSTANCE_TYPE_PUBLIC) {
       $input_values['endpoint'] = '';
     }
     if (empty($input_values['authorization_server_type']) || $input_values['authorization_server_type'] == 'default') {
       $input_values['authorization_server'] = '';
     }
 
+    // Remove unneeded values if on a Hybrid instance.
+    if ($input_values['instance_type'] == EdgeKeyTypeInterface::INSTANCE_TYPE_HYBRID) {
```
cnovak: What about `username`, `password`, `client id`, `client secret`? Shouldn't those values get removed also?
I guess maybe not so that you can flip between radio buttons and not loose data?

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -218,13 +263,20 @@ public function processSubmittedKeyValue(FormStateInterface $form_state) {
     $input_values = $form_state->getValues();
 
     // Make sure the endpoint defaults are not overridden by other values.
-    if (empty($input_values['endpoint_type']) || $input_values['endpoint_type'] == EdgeKeyTypeInterface::EDGE_ENDPOINT_TYPE_DEFAULT) {
+    if ($input_values['instance_type'] == EdgeKeyTypeInterface::INSTANCE_TYPE_PUBLIC) {
```
cnovak: Shouldn't public and private remove `account_json_key`?

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
```diff
@@ -282,8 +310,8 @@ protected function validateForm(callable $visitFormAsAdmin): void {
     $web_assert->elementNotContains('css', 'textarea[data-drupal-selector="edit-debug-text"]', $random_pass);
     $page->fillField('Password', $this->password);
 
-    // Test invalid username when using default endpoint.
-    $page->selectFieldOption('key_input_settings[endpoint_type]', 'default');
+    // Test invalid username when using default public cloud endpoint.
```
cnovak:
```suggestion
    // Test invalid username when using public cloud endpoint.
```

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
```diff
@@ -141,6 +141,36 @@ public function testKeyAddForm() {
...
+    // Test the authentication form.
+    $this->drupalLogin($this->rootUser);
+    $this->drupalGet(Url::fromRoute('entity.key.add_form'));
+    $this->visitKeyAddForm();
+
+    $page = $this->getSession()->getPage();
+
+    $page->selectFieldOption('key_input_settings[instance_type]', EdgeKeyTypeInterface::INSTANCE_TYPE_HYBRID);
+    $page->fillField('Organization', $organization);
+    $page->fillField('Account key', $account_key);
```
cnovak: Make sure to update this line if you take my field title change above

cnovak (changes requested):
1.  In `src/KeyEntityFormEnhancer.php` you need to call [oauthTokenStorage->checkRequirements();](https://github.com/apigee/apigee-edge-drupal/pull/275/files#diff-f42d077926d039ab00f544a6e367f695R342) during `validateForm()` or an exception is thrown if you don't have the private filesystem set. 
2. In the case the JWT token is invalid, a \DomainException is thrown and message "Failed to connect to Apigee Edge. Error message: OpenSSL unable to sign data" is displayed. May be good to change \Drupal\apigee_edge\KeyEntityFormEnhancer::createSuggestion to display message that says the private key in JSON string is invalid.
3. If you use "Public Cloud" as the instance type and set username and password, then switch to "Hybrid", the username and pass are still saved to the Key Provider.  This seems odd since you would expect that data to be cleared. I would think we would not want to save old values, for example if you used a username/pass and switched to hybrid, you would think those values are no longer saved in the system.

src/Plugin/EdgeKeyTypeInterface.php
arlina-espinoza: Changed to`@deprecated See EdgeKeyTypeInterface::getEndpointType().`.The referenced (also deprecated) method contains the link to issue #268.

src/Plugin/EdgeKeyTypeInterface.php
arlina-espinoza: Changed to`@deprecated See EdgeKeyTypeInterface::getEndpointType().`.The referenced (also deprecated) method contains the link to issue #268.

apigee_edge.install
arlina-espinoza: Fixed.

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
arlina-espinoza: Updated field name :+1:

src/HybridCredentials.php
arlina-espinoza: Fixed.

src/Plugin/EdgeKeyTypeBase.php
arlina-espinoza: This is a great suggestion, but it has to do with the PHP client library. I've added it as a new issue: https://github.com/apigee/apigee-client-php/issues/76

src/KeyEntityFormEnhancer.php
arlina-espinoza: Fixed.

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
arlina-espinoza: Added that:
https://github.com/arlina-espinoza/apigee-edge-drupal/blob/f5b7a40c444e95471f2ac8f496ed1d97ea092441/src/Plugin/KeyInput/ApigeeAuthKeyInput.php#L284-L287

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
arlina-espinoza: I've added this too:
https://github.com/arlina-espinoza/apigee-edge-drupal/blob/f5b7a40c444e95471f2ac8f496ed1d97ea092441/src/Plugin/KeyInput/ApigeeAuthKeyInput.php#L274-L283

src/HybridAuthentication.php
arlina-espinoza: Good idea, I've moved them to the "connector" namespace. :+1:

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
arlina-espinoza: I've pointed it to the drupal.org documentation page that she gave me. It is still not pointing to a particular header, but that would be the right page.

src/Plugin/EdgeKeyTypeBase.php
arlina-espinoza: @cnovak I think it would further complicate the user experience of setting up the connection by having a different key type for Hybrid, as it would involve exposing the key type on the module configuration page (admin/config/apigee-edge/settings), and then it would not be much different than setting up a key for use by the Edge module directly on the keys module settings (admin/config/system/keys/add).
I've added more sanitization upon saving the key so that only the relevant values are saved (ie, removing the unused values), so we can keep using this same class. Let me know if you'd like to discuss this further.

src/Plugin/EdgeKeyTypeInterface.php
arlina-espinoza: See https://github.com/apigee/apigee-edge-drupal/pull/275/files#r339764420.

src/KeyEntityFormEnhancer.php
arlina-espinoza: Should be fixed now.

src/Plugin/EdgeKeyTypeBase.php
arlina-espinoza: Added that:
https://github.com/arlina-espinoza/apigee-edge-drupal/blob/ef7cfb37ce769bfa40be985234b032b5e970466b/src/Plugin/EdgeKeyTypeBase.php#L52-L55

src/KeyEntityFormEnhancer.php
arlina-espinoza: Refactored so it's hopefully clearer:
https://github.com/arlina-espinoza/apigee-edge-drupal/blob/83a4cdd4131f8d6363133ab703e647b7c273dce3/src/KeyEntityFormEnhancer.php#L604-L627

src/KeyEntityFormEnhancer.php
arlina-espinoza: Fixed.

src/Plugin/EdgeKeyTypeBase.php
arlina-espinoza: Agreed, added that.

src/KeyEntityFormEnhancer.php
arlina-espinoza: Fixed.

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
arlina-espinoza: You are right, I've changed them to "or".

arlina-espinoza (commented):
>     1. In `src/KeyEntityFormEnhancer.php` you need to call [oauthTokenStorage->checkRequirements();](https://github.com/apigee/apigee-edge-drupal/pull/275/files#diff-f42d077926d039ab00f544a6e367f695R342) during `validateForm()` or an exception is thrown if you don't have the private filesystem set.
Fixed.
> 
>     2. In the case the JWT token is invalid, a \DomainException is thrown and message "Failed to connect to Apigee Edge. Error message: OpenSSL unable to sign data" is displayed. May be good to change \Drupal\apigee_edge\KeyEntityFormEnhancer::createSuggestion to display message that says the private key in JSON string is invalid.
Added the suggestion.
> 
>     3. If you use "Public Cloud" as the instance type and set username and password, then switch to "Hybrid", the username and pass are still saved to the Key Provider.  This seems odd since you would expect that data to be cleared. I would think we would not want to save old values, for example if you used a username/pass and switched to hybrid, you would think those values are no longer saved in the system.
Fixed so only the relevant values are saved according to the instance type you are connecting to.

src/KeyEntityFormEnhancer.php
arlina-espinoza: @cnovak I was running into it last week when working with it, but now that I try to reproduce it I can't get it to throw that exception, so I've removed it - might have been something on the way I had both this module and the php client as work-in-progress at the time.

arlina-espinoza:
@cnovak Thanks for your detailed review! I've gone through it and it should be ready for your feedback again.

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
```diff
@@ -20,7 +20,8 @@
 namespace Drupal\Tests\apigee_edge\FunctionalJavascript\Form;
 
 use Drupal\apigee_edge\Form\AuthenticationForm;
-use Drupal\apigee_edge\OauthTokenFileStorage;
+use Drupal\apigee_edge\Connector\OauthTokenFileStorage;
```
mxr576: You cannot move/rename classes because that is a BC break. You can create a new class in new locations and extend the new class in the old class and mark the old class as deprecated. 
Although, as the actual implementation did not change, I do not think this is necessary change.

mxr576 (changes requested):

mxr576:
https://github.com/apigee/apigee-client-php/pull/71 is the related PR in the PHP API client library.

src/KeyEntityFormEnhancer.php
```diff
@@ -452,8 +465,19 @@ private function createSuggestion(\Exception $exception, KeyInterface $key): Mar
     /** @var \Drupal\apigee_edge\Plugin\KeyType\ApigeeAuthKeyType $key_type */
     $key_type = $key->getKeyType();
 
+    if ($exception instanceof AuthenticationKeyException) {
+      $suggestion = $this->t('@fail_text Verify the Apigee Edge connection settings.', [
+        '@fail_text' => $fail_text,
+      ]);
+    }
+    // Invalid key / OpenSSL unable to sign data.
+    elseif ($exception instanceof DomainException) {
```
mxr576: This should be captured and handled in the API client level first: https://github.com/apigee/apigee-client-php/pull/71/files/3a50936a4198dfaf9377f674725ba59e74f3234e#r341003440 The client should return a `HybridOauth2AuthenticationException` and the previous exception should be the `DomainException` that you can capture here if you want.

src/Plugin/EdgeKeyTypeBase.php
mxr576: Related question: Can we use the Apigee Edge Environment Variables key storage to store hybrid credentials? It can be used to store OAuth credentials and the token is stored elsewhere. Hybrid should work the same way and the environment variables should be quite similar to OAuth (or preferably the same) that a user can use to store hybrid credentials (email, private key, auth server, etc).

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -218,13 +263,29 @@ public function processSubmittedKeyValue(FormStateInterface $form_state) {
...
```
cnovak: Need to check for $input_values['instance_type'] being set to avoid `Undefined index`

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -218,13 +263,29 @@ public function processSubmittedKeyValue(FormStateInterface $form_state) {
...
       $input_values['endpoint'] = '';
     }
     if (empty($input_values['authorization_server_type']) || $input_values['authorization_server_type'] == 'default') {
       $input_values['authorization_server'] = '';
     }
 
+    // Remove unneeded values if on a Hybrid instance.
+    if ($input_values['instance_type'] == EdgeKeyTypeInterface::INSTANCE_TYPE_HYBRID) {
```
cnovak: Need to check for $input_values['instance_type'] being set to avoid `Undefined index`

src/KeyEntityFormEnhancer.php
arlina-espinoza: Done :wink:

src/Plugin/EdgeKeyTypeBase.php
cnovak: We may want to start using the environment variables key provider that is now part of the Key module instead. I see that `APIGEE_EDGE_ACCOUNT_JSON_KEY` exists in  Apigee Edge Environment Variables key provider, but having trouble getting formatting correct for key when stored as an environment variable in nginx.

cnovak (changes requested):
Looks good, added one issue I found that needs addressing, and comments @mxr576 added

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
arlina-espinoza: Fixed!

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
arlina-espinoza: Fixed :+1:

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
arlina-espinoza: I've reverted this change.

arlina-espinoza:
@cnovak About https://github.com/apigee/apigee-edge-drupal/pull/275#discussion_r341400636:
The module is already using that, you should be able to set the `APIGEE_EDGE_ACCOUNT_JSON_KEY` environment var to the JSON encoded account key - I haven't tried it in nginx, but in apache it works correctly if I just run the JSON key through a minifier to remove whitespaces and line breaks.

src/Plugin/EdgeKeyTypeBase.php
arlina-espinoza: Hybrid implementation should also work with the Apigee Edge Environment Variables key storage. The `APIGEE_EDGE_ACCOUNT_JSON_KEY` environment var should be set to the JSON encoded account key - I haven't tried it in nginx, but in apache it works correctly if I just run the JSON key through a minifier to remove whitespaces and line breaks.

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
```diff
@@ -113,6 +138,10 @@ public function testAuthenticationForm() {
    * form is a customized Key edit form.
    */
   public function testKeyAddForm() {
+    if ($this->instanceType == EdgeKeyTypeInterface::INSTANCE_TYPE_HYBRID) {
+      $this->markTestSkipped('Skipping "testAuthenticationForm": can only be tested with public/private cloud credentials.');
```
cnovak: Should be “skipping “testKeyAddForm”

cnovak (changes requested):
One small change

tests/src/FunctionalJavascript/Form/AuthenticationFormJsTest.php
arlina-espinoza: Fixed!

cnovak (approved):
LGTM

PR #277, opened 2019-10-09
[#276] Update the anonymous user first_name and last_name fields

### LINKED ISSUE #276: Recent log messages shows incorrect user
## Description
Related to apigee/apigee-devportal-kickstart-drupal#270, the Recent log messages page shows wrong user for anonymous user logs, but detail page shows correct user. 

## Steps to Reproduce
1. Go to **Recent log messages**
2. Filter by type **cron**

** Actual Behavior **
The **User** column shows Firstname Lastname (not verified) as the username, which is the unfortunate name by default of the user uid 1.
Screen Shot 2019-10-02 at 10 52 45 AM

**Expected behavior**
The username should be anonymous, as you can see when clicking into detail of the message:
Screen Shot 2019-10-02 at 10 53 15 AM

## Version Info
apigee/apigee_devportal_kickstart        1.5.0

shadcn:
Fixes #276

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -1121,6 +1121,11 @@ function template_preprocess_app_credential(array &$variables) {
  * it has been in Apigee Edge.
  */
 function apigee_edge_user_presave(UserInterface $account) {
+  // Do not sync the anonymous user.
+  if ($account->isAnonymous()) {
+    return;
+  }
+
   // If the developer-user synchronization is in progress, then saving
   // developers while saving Drupal user should be avoided.
   if (_apigee_edge_is_sync_in_progress()) {
--- a/apigee_edge.install
+++ b/apigee_edge.install
@@ -86,6 +86,9 @@ function apigee_edge_install() {
       'analytics own developer_app',
     ];
     user_role_grant_permissions(RoleInterface::AUTHENTICATED_ID, $authenticated_user_permissions);
+
+    // Update the anonymous user first_name and last_name fields.
+    _apigee_edge_rename_anonymous_user();
   }
 }
 
@@ -176,5 +179,25 @@ function apigee_edge_update_8002() {
       $key->setKeyValue(Json::encode($values));
       $key->save();
     }
+  }
+}
+
+/**
+ * Rename the anonymous user from "Firstname Lastname" to "Anonymous".
+ */
+function apigee_edge_update_8003() {
+  _apigee_edge_rename_anonymous_user();
+}
+
+/**
+ * Helper to rename the anonymous user.
+ */
+function _apigee_edge_rename_anonymous_user() {
+  /** @var \Drupal\user\UserInterface $anonymous */
+  if ($anonymous = Drupal::entityTypeManager()->getStorage('user')->load(0)) {
+    // Use the value from config.
+    $anonymous->first_name->value = \Drupal::config('user.settings')->get('anonymous');
+    $anonymous->last_name->value = '';
+    $anonymous->save();
   }
 }


mxr576 (changes requested):
I think this is a completely unnecessary overhead that you are doing in this PR.

Just add and extra condition to `apigee_edge_user_format_name_alter()` that ensures it does not alter the formatted username of the anonymous user. 

https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/apigee_edge.module#L140

apigee_edge.module
```diff
@@ -1121,6 +1121,11 @@ function template_preprocess_app_credential(array &$variables) {
  * it has been in Apigee Edge.
  */
 function apigee_edge_user_presave(UserInterface $account) {
+  // Do not sync the anonymous user.
+  if ($account->isAnonymous()) {
```
mxr576, line 1125: This should not happen... nobody should update/delete the definition of the anonymous user. This only happens because you are doing something that you should not in  `_apigee_edge_rename_anonymous_user()`.

shadcn: There are situations other than ` _apigee_edge_rename_anonymous_user` where this could happen. There's a check, which you added, for the anonymous user in `apigee_edge_user_delete` (https://github.com/apigee/apigee-edge-drupal/commit/9de5c40b58730dd5ceb8e6afb47ac0d54aca3a70) so I don't see why there should not be one during presave.

shadcn:
> I think this is a completely unnecessary overhead that you are doing in this PR.
> 
> Just add and extra condition to `apigee_edge_user_format_name_alter()` that ensures it does not alter the formatted username of the anonymous user.
> 
> https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/apigee_edge.module#L140

This has been discussed and is still another viable solution. So I don't see this as being "completely unnecessary". Just another solution as @kscheirer pointed out in https://github.com/apigee/apigee-edge-drupal/issues/276#issuecomment-538511701

apigee_edge.module
mxr576: Okay, as an extra sanity check, you can also make sure here that someone did not call mistakenly the save() on the anonymous user, but the sync already excludes the anonymous user from the synching this is the reason why we did not need this extra sanity check before, only now, when you added `_apigee_edge_rename_anonymous_user()`.

This condition also excludes the anonymous user: https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/src/Job/DeveloperSync.php#L85

mxr576:
The first name and last name [_base fields_ ](https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/apigee_edge.module#L248-L276)are only added to the user because without them a Drupal user cannot be saved to Apigee Edge as a developer. These are required properties on Apigee Edge, so without them you cannot create a developer. This is another reason why these two base fields have to have a default value, "First name", "Last name".  For some reason, Drupal 8 persist these default field values to the database, this is what @kscheirer pointed out. Although, the problem that @cnovak is reported is about how Drupal formats the username of a user on the UI.
I do not think that we should do anything with the value in the database, because nobody is going to use that for the anonymous user, rather we should fix the formatting of its username.

_We added this alter hook to the module because we believe that on professional developer portals displaying a user's first name and last like (like "John Dorian") is a better solution than its favorite, sometimes meaningless username ( like: "asddsa1990")._

apigee_edge.module
mxr576: Not having an extra condition here is actually useful, because the actual requirement is not being a logged-in user, rather having all required fields (email, first name, last name, username) by Apigee Edge filled in when someone/something saves a user. If you have ever integrated SAML with the Apigee Edge module you know that most SAML modules try to create an empty stub user first with invalid, missing data, usually with a fake email address, that could lead to possible errors in the future if those stub users get saved to Apigee Edge. The current implements prevent this.

shadcn:
> The first name and last name [_base fields_ ](https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/apigee_edge.module#L248-L276)are only added to the user because without them a Drupal user cannot be saved to Apigee Edge as a developer. These are required properties on Apigee Edge, so without them you cannot create a developer. This is another reason why these two base fields have to have a default value, "First name", "Last name". For some reason, Drupal 8 persist these default field values to the database, this is what @kscheirer pointed out. Although, the problem that @cnovak is reported is about how Drupal formats the username of a user on the UI.
> I do not think that we should do anything with the value in the database, because nobody is going to use that for the anonymous user, rather we should fix the formatting of its username.
> 
> _We added this alter hook to the module because we believe that on professional developer portals displaying a user's first name and last like (like "John Dorian") is a better solution than its favorite, sometimes meaningless username ( like: "asddsa1990")._

Deszo I don't see why you're trying to explain the obvious here. This PR was meant to be a possible solution. There are many situations where the anonymous user could be pulled (`Drupal::entityTypeManager()->getStorage('user')->loadMultiple()`) and "Firstname and Lastname" will show up. But if you want to be right, be right. I'll close this PR and open a new one  (https://github.com/apigee/apigee-edge-drupal/pull/278). I just don't have the energy or time to do this anymore.

mxr576:
Sorry, I did not want to be offensive at all and "explain the obvious here". I have tried to prove my point and also give you a room to prove yours and with that give more context to others. Sorry, I do not have time to check the issue queue all the time, otherwise, I could have raised these concerns and share my idea in #276 before you would have started to implement anything.

>  the anonymous user could be pulled (Drupal::entityTypeManager()->getStorage('user')->loadMultiple()) 

That is correct, but what do you do 99,9% of the times? Exclude the anonymous user from the query because it is not a real user, it is just a classification for non-logged in/deleted users. When you run an update hook to set a field value on all users you also exclude the anonymous user because there is no need to update that user.

cnovak:
I want to point out that we should be using welcoming and inclusive language during our conversations per [our code of conduct](https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/CODE_OF_CONDUCT.md). 

Please stay positive with all differing viewpoint conversations, thanks.

PR #278, opened 2019-10-09
[#276] Exclude anonymous user from format

### LINKED ISSUE #276: Recent log messages shows incorrect user

shadcn:
Fixes #276

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -139,7 +139,7 @@ function apigee_edge_theme() {
  */
 function apigee_edge_user_format_name_alter(&$name, AccountInterface $account) {
   // Ignore Anonymous user.
-  if ($account instanceof UserInterface) {
+  if ($account instanceof UserInterface && !$account->isAnonymous()) {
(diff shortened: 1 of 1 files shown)


arlina-espinoza (approved):
Tested and LGTM. Thanks @arshad :)

Issue #280, opened 2019-10-14
Stop changing username to "Firstname Lastname"


cnovak:
## Description
Issue https://github.com/apigee/apigee-devportal-kickstart-drupal/issues/270 explains how changing the username to display user's firstname and lastname has unintended consequences. Instead of fixing this in Kickstart with https://github.com/apigee/apigee-devportal-kickstart-drupal/pull/274, we should properly fix this in this base module.

## To Reproduce
1. Create a new user with **First Name** "Test", **Last Name** "User", **Email** test@example.com".
2. View the user's profile page, the username will be 
<img width="484" alt="user-view" src="https://user-images.githubusercontent.com/613093/66067338-385eac00-e500-11e9-9585-fbe9a457ea84.png">

This will also happen on the user listing page in the admin screens.

## Expected behavior
Username should display the user's username.

## Version Info
Apigee Edge 8.x-1.1 and below

minnur:
@cnovak I created a new PR for the Apigee Edge and modified Kickstarter PR created by Arshad (we needed that one to display Full name and usernames on profile pages).

1. https://github.com/apigee/apigee-edge-drupal/pull/281
2. https://github.com/apigee/apigee-devportal-kickstart-drupal/pull/274

I also asked Arlina to review those PRs.

Thanks,

Minnur

mxr576:
Okay, so if I understand it correctly from all references issues, you also did something in Kickstarter which:
>The current release 1.5.0 hides the username from the system, and must autogenerate it in the background. 

Consequently, when a user registers *in a kickstart installation* then it does not set its username and when it would need to know its username then it cannot find it on the site because:

> This makes it very hard to use the Expose REST APIs to interact with API Docs instructions since the basic_auth module uses the username to authenticate. The username is hidden in the edit user form and shown in the view user profile incorrectly (displays First Name Last Name)

Neither an admin cannot tell that to a user because on the admin/people page even an admin cannot see it?

https://github.com/apigee/apigee-devportal-kickstart-drupal/issues/270#issuecomment-540690858

Now I understand why this username `apigee_edge_user_format_name_alter()` implementation is a problem *in a Kickstarter* installation. It is not a problem on regular installations when only the Apigee Edge module is installed, because the username is always visible on the registration form and it could be visible on the user view page.

Based on this, #281 is just a workaround to the problem, but it does not solve the real one.

I see two possible options to fix the real issue that you have:
[...]

Issue #272, opened 2019-10-03
Travis: Restore PHPUnit test running and highest-lowest testing


mxr576:
**Describe the bug**
Since #271, Travis CI does not run PHPUnit tests included in the main Apigee Edge module and in its submodules, not even, when a PR merged to 8.x-1.x. https://travis-ci.org/apigee/apigee-edge-drupal/builds/592688085 It only checks code style.

If we do run PHPUnit tests for PRs, that is fine, we documented that PR authors should run PHPUnit tests with their own Apigee trial accounts on their own Travis instances. But if we do not run PHPUnit tests to validate that the content of 8.x-1.x is stable and always ready for an immediate release, then how we can ensure real code quality?

mxr576:
Bump...
https://github.com/apigee/apigee-edge-drupal/pull/281#pullrequestreview-309188201

PR #281, opened 2019-10-16
[#280 ] Stop changing username to "Firstname Lastname".

minnur:
Work in progress.

### DIFF
--- a/src/UserDeveloperConverter.php
+++ b/src/UserDeveloperConverter.php
@@ -104,7 +104,11 @@ class UserDeveloperConverter implements UserDeveloperConverterInterface {
       $setter = 'set' . ucfirst($developer_prop);
       $getter = 'get' . ucfirst($developer_prop);
       if ($user->get($base_field)->value !== $developer->{$getter}()) {
-        $developer->{$setter}($user->get($base_field)->value);
+        $value = $user->get($base_field)->value;
+        // Make sure we don't set first name and last name when nothing to set.
+        if (!in_array($base_field, ['first_name', 'last_name']) && empty($value)) {
+          $developer->{$setter}($value);
+        }
         $successful_changes++;
       }
     }
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -135,21 +135,6 @@ function apigee_edge_theme() {
 }
 
 /**
- * Implements hook_user_format_name_alter().
- */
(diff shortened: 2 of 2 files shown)


arlina-espinoza (commented):
@cnovak Do we want to ultimately delete all _First Last_ formatting from this module? This PR currently just removes the initial default values, but keeps `apigee_edge_user_format_name_alter()`, which will alter the display name of a user (ie, basically the user entity's label). My vote would be to remove that hook from `apigee_edge` as it is presentation logic that is not related to this module and is just making things confusing.

mxr576:
Well, you should keep in mind that at least 500+ sites are already using this module with the old implementation. Because the module is stable you should provide a backward-compatibility layer for those sites, IOW, existing sites still should work the same as before, only new installations would not use first name + last name.

Btw, does not this change breaks the developer synchronization?

minnur:
I made the change so first name and last name values would be kept. No that I think even without the latest commit old sites would continue to work .

src/UserDeveloperConverter.php
```diff
@@ -104,7 +104,11 @@ public function convertUser(UserInterface $user): UserToDeveloperConversionResul
       $setter = 'set' . ucfirst($developer_prop);
       $getter = 'get' . ucfirst($developer_prop);
       if ($user->get($base_field)->value !== $developer->{$getter}()) {
-        $developer->{$setter}($user->get($base_field)->value);
+        $value = $user->get($base_field)->value;
+        // Make sure we don't set first name and last name when nothing to set.
+        if (!in_array($base_field, ['first_name', 'last_name']) && empty($value)) {
```
mxr576, line 109: This is not going to work. As I described in https://github.com/apigee/apigee-edge-drupal/pull/277#issuecomment-539979633, you cannot create a developer in Apigee Edge without a first name or a last name.

Install a clean Drupal site. Create n+1 users, install the Apigee Edge module and configure the connection. Try to run the developer sync on the UI, it will fail with errors like this:

image

image

(users created earlier than someone enabled the Apigee Edge module does not have a first name and last name that the system could send to Apigee Edge when it tries to create/update developers.)

mxr576 (changes requested):
I had to test this manually as I the CI does not run tests anymore... not even on PR creator's Travis instance :roll_eyes:, @cnovak please restore this feature. See #272

Issue #284, opened 2019-10-23
Analytics "chart" shows ranges but no series


kford:
**Describe the bug**
Developer App Analytics shows the appropriate range for the metric selected, but there are no "series" plotted onto it.  Export to CSV does appear to have the right values

**To Reproduce**
1. Create API Proxy with API Key Verification
2. Create API Product with API Proxy
3. Login to kickstart portal as user, create App, consume Product
4. Make calls to API Proxy, using API Key
5. View App Analytics
6. Chart has ranges but no "series"

**Expected behavior**
Expect to see Chart with series depicting metrics over time

**Screenshots**
attached

**Additional context**
---

**Version Info**
COMPOSER SHOW OUTPUT BELOW
-----------------------------------------------------------
drupal@9af78098554a:~/project$ cat composer_show.out 
apigee/apigee-client-php                 2.0.3        Client library for connecting to the Apigee Edge API.
apigee/apigee_devportal_kickstart        1.5.0        A fast demo and starting point for Apigee Developer Portals.
asm89/stack-cors                         1.2.0        Cross-origin resource sharing library and stack middleware
brumann/polyfill-unserialize             v1.0.4       Backports unserialize options introduced in PHP 7.0 to older PHP versions.
chi-teck/drupal-code-generator           1.30.4       Drupal code generator
clue/stream-filter                       v1.4.1       A simple and modern approach to stream filtering in PHP
commerceguys/addressing                  v1.0.6       Addressing library powered by CLDR and Google's address data.
commerceguys/intl                        v1.0.4       Internationalization library powered by CLDR data.
composer/installers                      v1.7.0       A multi-framework Composer library installer
composer/semver                          1.5.0        Semver library that offers utilities, version constraint parsing and validation.
consolidation/annotated-command          2.12.0       Initialize Symfony Console commands from annotated command class methods.
consolidation/config                     1.2.1        Provide configuration services for a commandline tool.
consolidation/filter-via-dot-access-data 1.0.0        This project uses dflydev/dot-access-data to provide simple output filtering for applications built with annotated-command / Robo.
consolidation/log                        1.1.1        Improved Psr-3 / Psr\Log logger based on Symfony Console components.
consolidation/output-formatters          3.5.0        Format text by applying transformations provided by plug-in formatters.
consolidation/robo                       1.4.10       Modern task runner
consolidation/self-update                1.1.5        Provides a self:update command for Symfony Console applications.
[...]

PR #286, opened 2019-10-24
[#284] Analytics "chart" shows ranges but no series.

minnur:
Fix for https://github.com/apigee/apigee-edge-drupal/issues/284

Changed routine to determine earliest and the latest timestamps.

### DIFF
--- a/src/Form/AppAnalyticsFormBase.php
+++ b/src/Form/AppAnalyticsFormBase.php
@@ -385,7 +385,7 @@ abstract class AppAnalyticsFormBase extends FormBase {
       // must be JSON encoded before passing.
       // @see: https://developers.google.com/chart/interactive/docs/gallery/linechart#configuration-options
       $visualization_options = [
-        'width' => '900',
+        'width' => '100%',
         'legend' => 'none',
         'interpolateNulls' => 'true',
         'hAxis' => [
--- a/src/Form/AppAnalyticsFormBase.php
+++ b/src/Form/AppAnalyticsFormBase.php
@@ -373,37 +373,27 @@ abstract class AppAnalyticsFormBase extends FormBase {
       $this->store->set($data_id = Crypt::randomBytesBase64(), $analytics);
       $form['export_csv']['#url'] = Url::fromRoute('apigee_edge.export_analytics.csv', ['data_id' => $data_id]);
 
+      $values = $analytics['stats']['data'][0]['metric'][0]['values'];
+
       $form['#attached']['drupalSettings']['analytics']['metric'] = $form['controls']['metrics']['#options'][$metric];
       $form['#attached']['drupalSettings']['analytics']['timestamps'] = $analytics['TimeUnit'];
-      $form['#attached']['drupalSettings']['analytics']['values'] = $analytics['stats']['data'][0]['metric'][0]['values'];
-      $form['#attached']['drupalSettings']['analytics']['skip_zero_values'] = TRUE;
+      $form['#attached']['drupalSettings']['analytics']['values'] = $values;
+      $form['#attached']['drupalSettings']['analytics']['skip_zero_values'] = FALSE;
       $form['#attached']['drupalSettings']['analytics']['language'] = $this->currentUser()
         ->getPreferredLangcode();
       $form['#attached']['drupalSettings']['analytics']['chart_container'] = $form['chart']['#attributes']['id'];
-
-      $view_window_min = $view_window_max = 0;
-      if (array_key_exists('TimeUnit', $analytics) && is_array($analytics['TimeUnit'])) {
-        for ($i = count($analytics['TimeUnit']) - 1; $i > 0; $i--) {
-          if ($analytics['stats']['data'][0]['metric'][0]['values'][$i] !== 0) {
-            $view_window_min = $i;
-          }
-          if ($view_window_max === 0 && $analytics['stats']['data'][0]['metric'][0]['values'][$i] !== 0) {
-            $view_window_max = $i;
-          }
-        }
-      }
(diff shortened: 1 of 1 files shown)


src/Form/AppAnalyticsFormBase.php
```diff
@@ -376,34 +376,22 @@ protected function generateResponse(array &$form, AppInterface $app, string $met
...
-          }
-        }
-      }
-
       // Visualization options for Google Charts draw() function,
       // must be JSON encoded before passing.
       // @see: https://developers.google.com/chart/interactive/docs/gallery/linechart#configuration-options
       $visualization_options = [
-        'width' => '100%',
+        'width' => '900',
```
mxr576: How this change is related to the fix of the problem? Does not this break the responsive support of the chart?

src/Form/AppAnalyticsFormBase.php
```diff
@@ -376,34 +376,22 @@ protected function generateResponse(array &$form, AppInterface $app, string $met
...
         'legend' => 'none',
         'interpolateNulls' => 'true',
         'hAxis' => [
           'viewWindow' => [
-            'min' => $analytics['TimeUnit'][$view_window_min],
-            'max' => $analytics['TimeUnit'][$view_window_max],
+            'min' => !empty($analytics['TimeUnit']) ? min($analytics['TimeUnit']) : 0,
```
mxr576, line 393: Can you confirm that this fix works every time? It seems odd that a simple change like this can replace a complex calculation.

src/Form/AppAnalyticsFormBase.php
minnur: 100% doesn't make it responsive.

src/Form/AppAnalyticsFormBase.php
minnur: No need for a complex routine here. We just need to find minimum and maximum value from the data list and set those as a threshold

src/Form/AppAnalyticsFormBase.php
mxr576: But it is still a relative value and if this change is not related to the fix of the bug we should rather keep it untouched.

minnur: 900 vs 100% is a big difference in what you see in on the page. We now display 0 values in the chart, not just non-zero values.

mxr576: Can you demonstrate the difference with screenshots?

mxr576: > Width of the chart, in pixels.Type: numberDefault: width of the containing element

So what happens if we do not specify the width?

minnur: Same as if it would be with 100% value.

mxr576: Yeah, this is what I thought so, but I am still curious about what difference a user can see if the value is 900 or "with of the containing element", for example with the Bartik theme.

cnovak: I agree with @mxr576, I do not see any issues on how it renders on different themes using 100%.  If we stick to fixed width, smaller and larger sizes will be stuck using 900px. What problem is a fixed width fixing?

Here is 100% on a few themes, it is rendering correctly as the screen width changes:

Screen Shot 2019-10-28 at 5 46 50 PM
Screen Shot 2019-10-28 at 5 47 39 PM

cnovak (changes requested):
Works great showing proper 0 values, here is my data:

```
Date,"Message count"
"2019-10-29 00:00:00 +00:00",8
"2019-10-28 23:00:00 +00:00",12
"2019-10-28 22:00:00 +00:00",12
"2019-10-28 21:00:00 +00:00",10
"2019-10-28 20:00:00 +00:00",45
"2019-10-28 19:00:00 +00:00",0
"2019-10-28 18:00:00 +00:00",0
"2019-10-28 17:00:00 +00:00",0
"2019-10-28 16:00:00 +00:00",1
"2019-10-28 15:00:00 +00:00",0
"2019-10-28 14:00:00 +00:00",0
"2019-10-28 13:00:00 +00:00",0
"2019-10-28 12:00:00 +00:00",0
"2019-10-28 11:00:00 +00:00",0
"2019-10-28 10:00:00 +00:00",0
"2019-10-28 09:00:00 +00:00",0
"2019-10-28 08:00:00 +00:00",0
"2019-10-28 07:00:00 +00:00",0
"2019-10-28 06:00:00 +00:00",0
"2019-10-28 05:00:00 +00:00",0
"2019-10-28 04:00:00 +00:00",0
"2019-10-28 03:00:00 +00:00",0
"2019-10-28 02:00:00 +00:00",0
"2019-10-28 01:00:00 +00:00",0
"2019-10-28 00:00:00 +00:00",0

```

Screen Shot 2019-10-28 at 5 54 28 PM


However, we should not be changing the width to a fixed width as far as I can see.

minnur:
@cnovak it wasn't showing the whole chart for me with 100% in the kickstarter theme. Making it 900px worked better. I guess this is something that should be address in the kickstarter

minnur:
I made the change and set the width back to 100%

cnovak:
> @cnovak it wasn't showing the whole chart for me with 100% in the kickstarter theme. Making it 900px worked better. I guess this is something that should be address in the kickstarter

Yes, agreed, I put in https://github.com/apigee/apigee-devportal-kickstart-drupal/issues/278 about this. I did not see the chart getting cut off on my screen, if you have a screenshot please add to that ticket, thx.

cnovak (approved):

PR #287, opened 2019-10-25
[#285] Fix access check on the user/[uid]/create-app

### LINKED ISSUE #285: Users are able to create applications for other users
**Describe the bug**
Currently if you are logged in as a user and navigate to `/user/{userId}/create-app`, you can change the `userId` to one that is not your own and you can create applications for other users.

**To Reproduce**
1. Create 2 users
2. Log in as one user
3. Create a developer app for another user by changing the `userId` to another user's `userId` (/user/{userId}/create-app)

**Expected behavior**

There should be separate permissions for "Create any developer app" and "Create own developer app". Currently any user can create developer apps for any other user, since the permission is only for "Create developer app".

**Screenshots**
Screen Shot 2019-10-23 at 4 37 59 PM

**Additional context**
The endpoint for creating a developer app is documented [here](https://apidocs.apigee.com/management/apis/post/organizations/%7Borg_name%7D/developers/%7Bdeveloper_email_or_id%7D/apps). It looks like the endpoint only takes in an email address without any additional validation. This looks intentional since Admin accounts should be able to create applications for all users, however most users should *not* have this ability.

**Version Info**
Version 8.x-1.0

mxr576:
Closes #285

Fixes access check on the mentioned route. As it is a critical issue, I believe the fix should be merged as soon as possible.

### DIFF
--- a/src/Access/AppsPageAccessCheck.php
+++ b/src/Access/AppsPageAccessCheck.php
@@ -53,4 +53,27 @@ class AppsPageAccessCheck implements AccessInterface {
     );
   }
 
+  /**
+   * Grant access to the user/[uid]/create-apps page.
+   *
+   * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
+   *   The parametrized route.
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The currently logged in account.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   */
+  public function createAppAccess(RouteMatchInterface $route_match, AccountInterface $account) {
+    /** @var \Drupal\user\UserInterface $user */
+    $user = $route_match->getParameter('user');
+    if ($user === NULL) {
+      return AccessResult::forbidden('User is missing from route.');
+    }
+    return AccessResultAllowed::allowedIf(
+      ($account->id() === $user->id() && $account->hasPermission('create developer_app')) ||
+      ($account->hasPermission('administer developer_app'))
+    );
+  }
+
 }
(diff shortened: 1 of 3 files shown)


tests/src/Functional/DeveloperAppUITest.php
```diff
@@ -593,4 +594,27 @@ public function testBreadcrumbOnDeveloperAppPages() {
     $this->assertEquals(Url::fromRoute('entity.developer_app.collection_by_developer', ['user' => $user->id()])->toString(), $breadcrumb_links[2]->getAttribute('href'));
   }
 
+  /**
+   * Validates access on the user/[uid]/create-app route.
+   *
+   * TODO Move this to \Drupal\Tests\apigee_edge\Functional\DeveloperAppPermissionTest
```
mxr576: "Moving" means, refactor the mentioned test to validate this properly, because it should have validated this.

mxr576:
This fix can be applied with Composer Patches by using the GH diff URL https://github.com/apigee/apigee-edge-drupal/pull/287.diff, but in that case, you have to install the latest dev version of the module from Github. If you do not want to switch from the latest stable, here is a patch that could be applied to that without additional workarounds.


[apigee-edge-fix-access-on-create-app-form-GH285.patch.txt](https://github.com/apigee/apigee-edge-drupal/files/3770834/apigee-edge-fix-access-on-create-app-form-GH285.patch.txt)

cnovak (approved):
Thanks much @mxr576 for the quick fix

mxr576:
@cnovak Yw, can you create a new release wirh this fix and #273? These are important fixes that should be shipped to customers immediately.

PR #288, opened 2019-10-30
[#279] Allow disabling username override w/ first name + last name

### LINKED ISSUE #279: Stop using "First Name" "Last Name" as default values
* The UID 1 user is always "Firstname" "Lastname", change that to be something better.
* Stop using "Firstname" "Lastname", but instead use username so that the name is clear when displayed. User can always edit to fix the name.
* We could use username in both fields, or if the username has a dot in it, split by that dot. For example, `jon.doe` can be Firstname: `Jon`, Lastname: `Doe`.  Not perfect, but keeps us from having all kinds of "firstname lastname" accounts.

We should test:
1. New install, UID 1 
2. What happens when there are 1 or more new accounts in Drupal but not in Edge and the Apigee Edge module gets installed?

_Originally posted by @kscheirer in https://github.com/apigee/apigee-edge-drupal/issues/276#issuecomment-538511701_

mxr576:
Possible additional configuration options that could be implemented
* only override if the first name and last name fields are actually set by the user (IOW, their values are not the default "Firstname" and "Lastname"). This could generate mixed results on the admin/people page that could be confusing.
* Add username to the formatted username, ex.: "John Doe" => "John Doe (superjohn)". The result would look better on the admin/people page, but other UIs where the username is displayed should be also considered.

Closes: #279 #280 #281

### DIFF
--- a/apigee_edge.install
+++ b/apigee_edge.install
@@ -178,3 +178,14 @@ function apigee_edge_update_8002() {
     }
   }
 }
+
+/**
+ * Initialize new developer_settings.override_usernames configuration.
+ */
+function apigee_edge_update_8003() {
+  $config_factory = \Drupal::configFactory();
+
+  $developer_settings = $config_factory->getEditable('apigee_edge.developer_settings');
+  $developer_settings->set('override_usernames', TRUE)
+    ->save(TRUE);
+}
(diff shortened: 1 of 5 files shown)


mxr576:
If the idea is acceptable, I can also provide a test coverage to this feature.

mxr576:
https://github.com/apigee/apigee-devportal-kickstart-drupal/issues/270 is also related and could be solved if the Kickstarter sets the new config to FALSE when it installs the module.

cnovak:
Fixed by #307 already, closing

Issue #267, opened 2019-09-25
Rules Module Integration


giteshk:
Add integration to [Business Rules module](https://www.drupal.org/project/business_rules) so that users can create rules around events.

## Events to react on
* App
  * Created
  * Updated
  * Deleted
* Teams
  * Team Created 
  * Team Updated
  * Team Deleted
  * Team Member Added
  * Team Member Removed
* User (these may already exist)
  * Created 
  * Updated
  * Deleted

## Example use cases we want to cover
  * Send out email to **messaging** admins when the **messaging** API Product is added to an app
  * Send out email to admins when new app is created
  * Approve apps that are created by developers that have an @foo.com email address
  * Send out notification to a developer when their app is approved

mxr576:
Rules is unstable and still under active development. It is not recommended to be used on production sites, I think it is too early to validate the compatibility because of these.
https://www.drupal.org/project/rules/releases/8.x-3.0-alpha5

(This should not be a bug ticket)

giteshk:
If not Rules can we suggest any other approach which is repeatable? 
eg. Send email on app creation, app deletion, Team creation, Team invitation

Maybe add something to the documentation about possible approaches customers can try out.

arunz6161:
To clarify the ask here is to evaluate if rules or business_rules will work better with D8P and document steps on how a user can setup email notification on common events like app creation, app deletion, team creation, team invitation.

mxr576:
I have just wanted to suggest Business Rules as an alternative. At this moment, it seems more promising than the old-friend, Rules.

minnur:
Business Rules integration is the module that works better than the Rules module.  For the full Business Rules module integration we would need to create more Event plugins and dispatch them where needed. Business Rules module RuleOn plugin is listening to those events and triggering actions. So far I see only about 4 events in the Apigee Edge module and almost all of them related to the credentials. Like new app credential created/deleted etc.

Actions to do:
1. Determine events that we need
2. Create Event plugins for each of the actions
3. Implement RuleOn Business Rules plugins. This would allow add actions such as send email, show message etc..

mxr576:
> So far I see only about 4 events in the Apigee Edge module and almost all of them related to the credentials.

[...]

PR #289, opened 2019-10-31
[#267] Initial integration for the Business Rules module.

minnur:
Work in progress.

### DIFF
--- a/src/Plugin/BusinessRulesReactsOn/AppCredentialAddApiProduct.php
+++ b/src/Plugin/BusinessRulesReactsOn/AppCredentialAddApiProduct.php
@@ -1,0 +1,43 @@
+<?php
+
+/**
+ * Copyright 2018 Google Inc.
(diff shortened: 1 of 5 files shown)


minnur:
This is ticket requires more events integration.

1. Create events and trigger them on certain events.
2. Create Business Rules integration plugins.

arlina-espinoza:
From my understanding, the following is what needs to be done:

1. Create `BusinessRulesReactsOn` plugin definitions for the events we need to handle - the current PR has some of that (or all?).
2. Throw events of type `BusinessRulesEvent` that use the "reacts_on" definitions from step 1 during the execution of the module. See as an example [`business_rules_user_login()`](https://git.drupalcode.org/project/business_rules/blob/8.x-1.x/business_rules.module#L348-367).

shadcn:
Closing this since #360 was merged.

PR #290, opened 2019-10-31
[#156] Display expired credentials

### LINKED ISSUE #156: DRUP-667 Display expired credentials as such
| Module Version              | 8.x-1.0-beta2
| PHP Client version          | 2.0.1

#### Actual Behavior

When a credential is expired, it is displayed on the app details page like:
```
Expires: 17 minutes 25 seconds ago
Key Status: approved
```
I understand that this is what the management server returns, but it's still confusing to the end users as they can't use this credential.

#### Expected Behavior

```
Key Status: expired
```

#### Steps to Reproduce the behavior

- On Edge's web UI, create a credential for an app with the expiration set to duration/2 minutes.
- Wait at least those 2 minutes.
- (Re)Load this app's details page in Drupal.

#### Proposed Solution

`template_preprocess_app_credential()` in `apigee_edge.module:998` could be changed to display `expired` as `#value` and `StatusPropertyElement::INDICATOR_STATUS_ERROR` as `#indicator_status` when the credential is expired.

shadcn:
Fixes #156 

This PR displays expired credentials on the app page under key status and adds a warning on the app listing page for expired credentials.

ac-test-app-for-expired_app___Apigee

Apps___Apigee

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -1081,6 +1081,10 @@ function template_preprocess_app_credential(array &$variables) {
       ];
     }
     elseif ($def['value_type'] === 'status') {
+      // Check if expired.
+      if ($normalized['expiresAt'] !== -1 && \Drupal::time()->getRequestTime() - intval($normalized['expiresAt'] / 1000) > 0) {
+        $value = t('Expired');
+      }
       $variables['primary_wrapper'][$property]['value'] = [
         '#type' => 'status_property',
         '#value' => $value,
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -1082,7 +1082,7 @@ function template_preprocess_app_credential(array &$variables) {
     }
     elseif ($def['value_type'] === 'status') {
       // Check if expired.
-      if ($normalized['expiresAt'] !== -1 && \Drupal::time()->getRequestTime() - intval($normalized['expiresAt'] / 1000) > 0) {
+      if ($normalized['expiresAt'] !== -1 && \Drupal::time()->getRequestTime() - (int)($normalized['expiresAt'] / 1000) > 0) {
         $value = t('Expired');
       }
       $variables['primary_wrapper'][$property]['value'] = [
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -1082,7 +1082,7 @@ function template_preprocess_app_credential(array &$variables) {
     }
     elseif ($def['value_type'] === 'status') {
       // Check if expired.
-      if ($normalized['expiresAt'] !== -1 && \Drupal::time()->getRequestTime() - (int)($normalized['expiresAt'] / 1000) > 0) {
+      if ($normalized['expiresAt'] !== -1 && \Drupal::time()->getRequestTime() - (int) ($normalized['expiresAt'] / 1000) > 0) {
         $value = t('Expired');
       }
       $variables['primary_wrapper'][$property]['value'] = [
--- a/src/Entity/ListBuilder/AppListBuilder.php
+++ b/src/Entity/ListBuilder/AppListBuilder.php
@@ -215,7 +215,7 @@ class AppListBuilder extends EdgeEntityListBuilder {
     // - any credentials of the app is in revoked status
     // - any products of any credentials of the app is in revoked or
     //   pending status.
-    if ($app->getStatus() === AppInterface::STATUS_APPROVED && ($warnings['revokedCred'] || $warnings['revokedOrPendingCredProduct'])) {
+    if ($app->getStatus() === AppInterface::STATUS_APPROVED && ($warnings['revokedCred'] || $warnings['revokedOrPendingCredProduct'] || $warnings['expiredCred'])) {
       $build['status'] = $rows[$info_row_css_id]['data']['status']['data'];
       $build['warning'] = [
         '#type' => 'html_tag',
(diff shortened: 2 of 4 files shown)


src/Entity/ListBuilder/AppListBuilder.php
```diff
@@ -316,6 +327,14 @@ protected function checkAppCredentialWarnings(AppInterface $app): array {
         }
         break;
       }
+
+      // Check for expired credentials.
+      if (($expired_date = $credential->getExpiresAt()) && \Drupal::time()->getRequestTime() - intval($expired_date->getTimestamp() / 1000) > 0) {
```
mxr576: DI the time service.

apigee_edge.module
```diff
@@ -1081,6 +1081,10 @@ function template_preprocess_app_credential(array &$variables) {
       ];
     }
     elseif ($def['value_type'] === 'status') {
+      // Check if expired.
+      if ($normalized['expiresAt'] !== -1 && \Drupal::time()->getRequestTime() - intval($normalized['expiresAt'] / 1000) > 0) {
```
mxr576: You should set the "Expired" special value for the `expiredAt` field earlier, here:

```php
<?php
    if ($def['value_type'] == 'date') {
      // TODO Should we make format configurable?
      /** @var \DateTimeInterface $value */
      if ($value !== -1) {
        $time_diff = \Drupal::time()->getRequestTime() - intval($value / 1000);
        if ($time_diff > 0) {
          $value = t('@time ago', ['@time' => $dateFormatter->formatTimeDiffSince(intval($value / 1000))]);
        }
        else {
          $value = t('@time hence', ['@time' => $dateFormatter->formatTimeDiffUntil((int) ($value / 1000))]);
        }
      }
      else {
        $value = t('Never');
      }
    }

```

src/Entity/ListBuilder/AppListBuilder.php
mxr576: Do not use inline variable assignment in conditions.
`($expired_date = $credential->getExpiresAt()) ` 
Also, the newly introduced variable is just a shortcut for `$credential->getExpiresAt() `.

mxr576 (changes requested):
QA: Replace `intval()` with typecasting `(int)` in the modified functions.

src/Entity/ListBuilder/AppListBuilder.php
```diff
@@ -215,7 +215,7 @@ protected function buildWarningRow(AppInterface $app, array &$rows) {
     // - any credentials of the app is in revoked status
```
mxr576: revoked status or expired

src/Entity/ListBuilder/AppListBuilder.php
shadcn: Done.

src/Entity/ListBuilder/AppListBuilder.php
shadcn: Nice catch. Updated the comment.

apigee_edge.module
shadcn: There was no comment here so I was not sure if this code is to display the expired date only or can `$value` be used here.

Wouldn't this `if ($def['value_type'] == 'date') {` be TRUE for `issuedAt` and `expiresAt` only?

src/Entity/ListBuilder/AppListBuilder.php
shadcn: `$credential->getExpiresAt()` can be null.

apigee_edge.module
```diff
@@ -1081,6 +1081,10 @@ function template_preprocess_app_credential(array &$variables) {
       ];
     }
     elseif ($def['value_type'] === 'status') {
+      // Check if expired.
+      if ($normalized['expiresAt'] !== -1 && \Drupal::time()->getRequestTime() - (int)($normalized['expiresAt'] / 1000) > 0) {
```
kscheirer: A constant instead of -1 would be a little clearer.

shadcn: There's already a few places where -1 is used so I did not want to change that but I see how a constant could be better. We can do that in another fix. Thanks @kscheirer

src/Entity/ListBuilder/AppListBuilder.php
mxr576: Yes, it can be `null` but it should not prevent you from moving this condition above where all values are calculated, above this line:

```php
    // Below, $value is expected to be a string in some places, and it might be
    // TranslatableMarkup or a string. If it is not a string, then in some
    // cases warnings will be generated. This way it is always a string,
    // removing ambiguity.
    $value = (string) $value;
```

The code block where it is now is responsible for the rendering of the (calculated) `$value`, this is my only problem with it.

src/Entity/ListBuilder/AppListBuilder.php
```diff
@@ -246,12 +258,22 @@ protected function buildWarningRow(AppInterface $app, array &$rows) {
...
       if ($warnings['revokedCred']) {
-        $row['data']['info']['data'] = $warnings['revokedCred'];
+        $items[] = $warnings['revokedCred'];
       }
       elseif ($warnings['revokedOrPendingCredProduct']) {
-        $row['data']['info']['data'] = $warnings['revokedOrPendingCredProduct'];
+        $items[] = $warnings['revokedOrPendingCredProduct'];
+      }
+
+      if ($warnings['expiredCred']) {
```
kscheirer, line 269: should this be an `elseif` or `switch`, or does it need ti be separate from the previous checks in $warnings?

kscheirer (approved):
LGTM, just a minor question

src/Entity/ListBuilder/AppListBuilder.php
shadcn: @kscheirer It has to be separate because we are always showing expiredCred warnings if they exist.

kscheirer: thanks 👍

PR #305, opened 2019-12-10
#304 Add compatibility with Drupal core 8.8.0

### LINKED ISSUE #304: [Urgent] Allow installation on Drupal  >= 8.8.0
We used to validate the compatibility of the module with the next Drupal core release when the BETA1/RC1 was out. This is not happening for a while and since #195 the module can be installed only on Drupal 8.7.x which only gets security updates since 8.8.0 is out.

https://www.drupal.org/project/drupal/releases/8.8.0

Tasks
- [ ] Validate compatibility with Drupal 8.8.0
- [ ] Allow installation on 8.8.0 and on 8.7.x still.

arlina-espinoza:
Closes #304 . Fixes compatibility issues with core 8.8.0.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "php": ">=7.1",
         "apigee/apigee-client-php": "^2.0.4",
         "cweagans/composer-patches": "^1.6.5",
-        "drupal/core": "~8.7.0",
+        "drupal/core": "~8.7",
         "drupal/entity": "^1.0",
         "drupal/key": "^1.8",
         "php-http/guzzle6-adapter": "^1.1.1"
@@ -16,7 +16,7 @@
         "behat/mink" : "dev-master#9ea1cebe",
         "drush/drush": "^9.0",
         "phpunit/phpunit": "^6.5",
-        "webflo/drupal-core-require-dev": "^8.7.0"
+        "webflo/drupal-core-require-dev": "^8.7"
     },
     "config": {
         "sort-packages": true
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "behat/mink" : "dev-master#9ea1cebe",
         "drush/drush": "^9.0",
         "phpunit/phpunit": "^6.5",
-        "webflo/drupal-core-require-dev": "^8.7"
+        "drupal/core-dev": "^8.7"
     },
     "config": {
         "sort-packages": true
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -16,7 +16,7 @@
         "behat/mink" : "dev-master#9ea1cebe",
         "drush/drush": "^9.0",
         "phpunit/phpunit": "^6.5",
-        "webflo/drupal-core-require-dev": "^8.7.0"
+        "webflo/drupal-core-require-dev": "^8.7"
```
minnur: Package webflo/drupal-core-strict is abandoned, you should avoid using it. Use drupal/core-recommended instead.

This might apply to webflo/drupal-core-require-dev too. Maybe we should check this.

composer.json
```diff
@@ -7,7 +7,7 @@
         "php": ">=7.1",
         "apigee/apigee-client-php": "^2.0.4",
         "cweagans/composer-patches": "^1.6.5",
-        "drupal/core": "~8.7.0",
+        "drupal/core": "~8.7",
```
cnovak, line 10: I am thinking that we do want to make this `~8.8.0`, following [drupal-composer project](https://github.com/drupal-composer/drupal-project/blob/8.x/composer.json#L23)

composer.json
cnovak: Looks like drupal-composer project is using [drupal/core-composer-scaffold](https://github.com/drupal/core-composer-scaffold) for this, which is what we patterned our composer project after.

cnovak (changes requested):
Changes needed to composer.json

composer.json
shadcn: +1 for this. I'm thinking we should probably do this for kickstart as well. `"drupal/core": "~8.8.0",`

arlina-espinoza: @cnovak / @arshad  Doing `~8.8.0` would prevent anyone on core 8.7.* from updating the module. Do we really want that? I like the idea for Kickstart though.

composer.json
arlina-espinoza: @minnur  I changed `webflo/drupal-core-require-dev` for `drupal/core-dev`. Please note that this is a `require-dev` dependency within the module, which mean it's not installed when someone download the module - I believe we only keep it there for travis tests, which runs `wikimedia/composer-merge-plugin` and uses those dev requirements on the test environment.

arlina-espinoza: > Looks like drupal-composer project is using [drupal/core-composer-scaffold](https://github.com/drupal/core-composer-scaffold) for this, which is what we patterned our composer project after.

@cnovak Isn't that for the Kickstart project template? I don't see that applying to the `apigee_edge` module.

composer.json
cnovak: @arlina-espinoza you are right, this is Apigee Edge module, so never mind this suggestion.  Added https://github.com/apigee/devportal-kickstart-project-composer/issues/20 for this

cnovak (approved):
LGTM

PR #340, opened 2020-02-25
[#339] Add a pagination fallback method in case call with "expand=true" fails.

### LINKED ISSUE #339: In one api fails, all other edge functionality gets blocked
**Describe the bug**
APIGEE edge executes management APIs to fetch API products, developers etc.
If the endpoint for API product fails with some particular product detail fetch somehow it is not allowing developers to do anything else on the page.

**To Reproduce**
1- There is a private product on the instance that has some issue on APIGEE cloud.
2- When I enable the module it fetches all of the products available on apigee orgs
3- When a particular product comes in a loop system start throwing en error and page doesn't get executed.

**Expected behavior**
If there is an error on a particular product should not be synced to the devportal rest of the product should work fine.

**Version Info**
8.x-1.6 Apigee Edge module.

arlina-espinoza:
Should fix #339.
It updates the listing method of the controller, so that if a call with `expand=true` fails (the default way to load all entities), it would retry fetching that page with just it's IDs (`expand=false`), and then loading each entity individually.
As mentioned in the issue comments, it is definitively not as performant, but it a solution to an edge case, and should allow showing all of the other entities (API products in the case of issue #339), while logging the errors in watchdog for both the `expand=true` call and the failed entity load call.
It should work, assuming the call with `expand=false` works.

### DIFF
--- a/src/Entity/Controller/CachedPaginatedEntityListingControllerTrait.php
+++ b/src/Entity/Controller/CachedPaginatedEntityListingControllerTrait.php
@@ -20,6 +20,7 @@
 
 namespace Drupal\apigee_edge\Entity\Controller;
 
+use Apigee\Edge\Exception\ApiException;
 use Apigee\Edge\Structure\PagerInterface;
 
 /**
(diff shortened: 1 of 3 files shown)


cnovak:
Closing since this fix did not help customer.

PR #342, opened 2020-02-27
[#332] Apigee Edge Debug: Display responses as default setting

### LINKED ISSUE #332: Apigee Edge Debug: Display responses as default setting
The formatting when first enabled does not show responses out of the box, so you always need to change the log message format at `/admin/config/apigee-edge/debug` before getting any useful info. Change the module to display responses by default.

arlina-espinoza:
Closes #332 .


cnovak (changes requested):
Changes look good, but message on screen should be updated:

Screen Shot 2020-03-02 at 9 11 25 AM

How about:
> Note: Adding API responses to log messages by using {response_formatted} token may contain sensitive data (ex.: app credentials, etc.).

arlina-espinoza:
> Changes look good, but message on screen should be updated:
> 
> Screen Shot 2020-03-02 at 9 11 25 AM
> 
> How about:
> 
> > Note: Adding API responses to log messages by using {response_formatted} token may contain sensitive data (ex.: app credentials, etc.).

I've pushed an update fixing that warning! :sweat_smile: Ready for another review.

cnovak (approved):
LGTM

mxr576:
@cnovak @arlina-espinoza is this PR intentionally does not contain an update hook that would activate this change on existing sites that are using this module?

cnovak:
Good point @mxr576, but I am not sure if we want to mess with the settings if someone changed it already.  What do you think @arlina-espinoza ?

mxr576:
I would only update it if it has not been changed yet.

PR #358, opened 2020-03-05
[#341] Enable kernel, functional, and functional JS tests

### LINKED ISSUE #341: Enable kernel, functional and JS tests in CircleCI
Enable CircleCI to run kernel, functional and JS tests - need to fix any tests that aren't passing for this to happen.

arlina-espinoza:
Enable CircleCI to run kernel, functional and JS tests.


arlina-espinoza:
Ready for review. This PR enables the kernel, functional and JS functional tests to run on each PR in CircleCI. Tests that have been refactored to use the mock client will run with mocks, while the other tests would be integration tests (pinging the real Edge API).

Some things that were done:

- Cleaned up composer.json and composer dependencies for CircleCI - removing duplicates and things that are brought in by "drupal/core-recommended" (the recommended way to ensure core dependencies are locked down to the required versions).
- Fixed some tests that were failing. In particular, functional tests were erroing out for the apigee_edge_apiproduct_rbac submodule, but switching them over to be functional JS tests made them work. I believe that is because D 8.8 made some part of the API product config form dependent on a JS (maybe a collapsible field or something), and so they worked when switched.

.circleci/RoboFile.php
```diff
@@ -410,7 +408,7 @@ public function configureModuleDependencies()
     $config = json_decode(file_get_contents('composer.json'));
 
     // The Drupal core image might need updating. Request the newest stable.
-    $config->require->{"drupal/core"} = "~8.8";
+    unset($config->require->{"drupal/core"});
```
cnovak, line 411: Can you comment on why this `unset` is being called?

composer.json
```diff
@@ -14,12 +14,9 @@
     },
     "require-dev": {
         "apigee/apigee-mock-client-php": "^1.0",
-        "behat/mink" : "dev-master#9ea1cebe",
```
cnovak, line 17: Why are `mink` packages being removed, but not `mink-extension`?

composer.json
```diff
@@ -14,12 +14,9 @@
...
         "behat/mink-extension": "v2.2",
-        "behat/mink-selenium2-driver": "1.3.x-dev",
         "bex/behat-screenshot": "^1.2",
         "cweagans/composer-patches": "^1.6",
-        "drupal/coder": "^8.3",
```
cnovak, line 22: why is `coder` being removed?

cnovak (commented):
Left some comments

.circleci/RoboFile.php
arlina-espinoza: The next line is `$config->require->{"drupal/core-recommended"} = "~8.8";`. We require `drupal/core-recommended` instead of `drupal/core` because _core-recommended_ locks down core dependencies correctly, and makes sure we are testing against the same library  versions core has been tested against (see https://www.drupal.org/docs/8/update/update-core-via-composer#migrate-webflo-drupal-core-strict-to-core-recommended)

composer.json
arlina-espinoza: We are already requiring `drupal/core-dev` in the module's `require-dev` section, which includes `behat/mink`, but not `mink-extension`. The added bonus is that requiring it through _drupal/core-dev_ makes sure it's brought in a version that is compatible with other libraries (ie, less code to maintain).

composer.json
arlina-espinoza: Same as the comment above (https://github.com/apigee/apigee-edge-drupal/pull/358#discussion_r393205915).

cnovak (approved):
LGTM

shadcn:
@cnovak @arlina-espinoza Since this is LGTM, I'm merging it so that I can properly debug #360

PR #359, opened 2020-03-11
Stop changing username to “firstName LastName”

cnovak:
Fixes #307 by removing `apigee_edge_user_format_name_alter` function. Username is now properly displayed.

PR #378, opened 2020-03-19
Fix apigee_edge_teams\Functional\UiTest::testUi()

arlina-espinoza:
Related to [#359] - fixes a test that relied on username displayed as firstname lastname.


cnovak (approved):
LGTM

PR #379, opened 2020-03-20
#346 Switch to mocks for kernel EntityControllerCacheTest

### LINKED ISSUE #346: Switch to mocks for kernel EntityControllerCacheTest

`/tests/src/Kernel/EntityControllerCacheTest.php`

arlina-espinoza:


shadcn (approved):
LGTM. Thanks @arlina-espinoza 👍

PR #380, opened 2020-03-20
#347 Switch to mocks for Kernel MockIntegrationToggleKernelTest

### LINKED ISSUE #347: Switch to mocks for Kernel MockIntegrationToggleKernelTest

`tests/src/Kernel/MockIntegrationToggleKernelTest.php`

arlina-espinoza:


shadcn (approved):

PR #381, opened 2020-03-20
#348 Switch to mocks for kernel OauthTokenFileStorageTest 

### LINKED ISSUE #348: Switch to mocks for kernel OauthTokenFileStorageTest

`/tests/src/Kernel/OauthTokenFileStorageTest.php`

arlina-espinoza:


shadcn (approved):

PR #382, opened 2020-03-20
#349 Switch to mocks for kernel TestFrameworkKernelTest

### LINKED ISSUE #349: Switch to mocks for kernel TestFrameworkKernelTest

`/tests/src/Kernel/TestFrameworkKernelTest.php`

arlina-espinoza:
This test didn't need refactoring, as it was already compatible. Just marking it as compatible/ready with the mock client.


shadcn (approved):

PR #386, opened 2020-03-24
[#385] Add an update hook to set the last installed storage definition

### LINKED ISSUE #385: Mismatched entity and/or field definitions after First name, Last name field changes
**Describe the bug**
After an `8.x-1.5` to `8.x-1.7` update it seems everything is working fine but I noticed the following message on the status page:

> Mismatched entity and/or field definitions
The following changes were detected in the entity type and field definitions.
User
The First name field needs to be updated.
The Last name field needs to be updated.

I have tried to view and edit user profiles and looks like everything is fine but this message appears.

**Screenshots**

entity-field-definitions.

**Additional context**

There is no custom code in the project which plays with the first name and last name fields or somehow modifies them.

shadcn:
Fixes #385

### DIFF
--- a/apigee_edge.install
+++ b/apigee_edge.install
@@ -222,3 +222,27 @@ function apigee_edge_update_8101() {
     }
   }
 }
+
+/**
+ * Update the field storage definitions for first_name and last_name fields.
+ */
+function apigee_edge_update_8102() {
+  $entity_type_id = 'user';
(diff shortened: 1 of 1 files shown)


arlina-espinoza (approved):
Tested and it LGTM. Thanks @arshad :)

PR #360, opened 2020-03-17
[#267] Add the apigee_edge_actions module

shadcn:
Fixes #267 

This PR adds the `apigee_edge_actions` (experimental) module.

The Apigee Edge Actions module provides rules integration for Apigee Edge. It makes it easy to automate tasks and react on events such as:

- Send an email when an App is created.
- Notify a developer when added to a Team.
- Notify admin when an API product is added to an App.

See the README for more.

See tests: https://app.circleci.com/pipelines/github/arshad/apigee-actions-drupal

PR #388, opened 2020-03-27
[#365] Switch functional DeveloperAppAnalyticsTest: use mock client.

### LINKED ISSUE #365: Switch to mocks for Functional DeveloperAppAnalyticsTest
`tests/src/Functional/DeveloperAppAnalyticsTest.php`

arlina-espinoza:
Refactors the test to use the sdk connector's client, and apigee_mock_api_client.
It reuses some of the mock templates that would be added by #360 , so this PR will need a touch up after that one gets merged.

### DIFF
--- a/tests/src/Functional/DeveloperAppAnalyticsTest.php
+++ b/tests/src/Functional/DeveloperAppAnalyticsTest.php
@@ -22,6 +22,7 @@ namespace Drupal\Tests\apigee_edge\Functional;
 use Apigee\Edge\Api\Management\Entity\App;
 use Drupal\apigee_edge\Entity\Developer;
 use Drupal\apigee_edge\Entity\DeveloperApp;
+use Drupal\apigee_edge\Entity\DeveloperAppInterface;
 use Drupal\Component\Utility\Crypt;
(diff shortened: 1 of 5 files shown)


cnovak (approved):
LGTM

tests/src/Functional/DeveloperAppAnalyticsTest.php
```diff
@@ -228,4 +248,36 @@ protected function exportAnalyticsTest() {
...
+
+  /**
+   * Add an app analytics mock response to the stack.
+   *
+   * @param \Drupal\apigee_edge\Entity\DeveloperAppInterface $app
+   *   The app.
+   * @param int $response_code
+   *   Response code, defaults to 200.
+   */
+  protected function queueDeveloperAppResponse(DeveloperAppInterface $app, $response_code = 200) {
```
cnovak: Should this function be in `ApigeeMockApiClientHelperTrait` instead for other tests to use?

shadcn: @arlina-espinoza Let's move it to the trait. I have a few tests that could use this.

shadcn:
@googlebot I consent.

PR #389, opened 2020-03-30
[#375] Switch to mocks for Functional StatusReportTest

### LINKED ISSUE #375: Switch to mocks for Functional StatusReportTest
`tests/src/Functional/StatusReportTest.php`

shadcn:
Fixes #375


tests/src/Functional/StatusReportTest.php
```diff
@@ -41,6 +41,19 @@ class StatusReportTest extends ApigeeEdgeFunctionalTestBase {
...
+   * {@inheritdoc}
+   */
+  protected static $mock_api_client_ready = TRUE;
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp() {
+    parent::setUp();
+    $this->installExtraModules(['apigee_edge_debug']);
```
shadcn: This test had a dependency on `apigee_edge_test` which in turn depends on `apigee_edge_debug`. 

@arlina-espinoza I couldn't really figure out why removing `apigee_edge_debug` makes this test fail.

arlina-espinoza:
@googlebot I consent.

tests/src/Functional/StatusReportTest.php
arlina-espinoza: @arshad `apigee_edge_debug` decorates the SDK connector service, and our mock client module alters the service definition by switching an argument (the http client for our mock client), so as it is they are not currently compatible. If you enable `apigee_edge_debug` alongside `apigee_mock_api_client` what happens is that the debug module overwrites the changed parameter and it makes real API calls.
I added a note to the mock client module README that explains why the modules aren't compatible and that refactored tests shouldn't use either `apigee_edge_debug` or `apigee_edge_test`.

arlina-espinoza:
@cnovak Please look at [this comment](https://github.com/apigee/apigee-edge-drupal/pull/389#discussion_r403183673) and the note added to the README. Thanks :)

arlina-espinoza (commented):
As I've pushed multiple changes, I'm letting someone else review. @cnovak / @arshad

shadcn:
This looks good @arlina-espinoza. Thank you. I don't think I could've figured this out 👍

cnovak (approved):
+1 for updated README

PR #390, opened 2020-03-30
[#373] Switch to mocks for Functional ErrorHandlerTest

### LINKED ISSUE #373: Switch to mocks for Functional ErrorHandlerTest
`tests/src/Functional/ErrorHandlerTest.php`

shadcn:
Fixes #373

### DIFF
--- a/tests/src/Functional/ErrorHandlerTest.php
+++ b/tests/src/Functional/ErrorHandlerTest.php
@@ -30,6 +30,11 @@ use Symfony\Component\HttpFoundation\Response;
 class ErrorHandlerTest extends ApigeeEdgeFunctionalTestBase {
 
   /**
+   * {@inheritdoc}
+   */
+  protected static $mock_api_client_ready = TRUE;
+
+  /**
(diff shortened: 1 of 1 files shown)


tests/src/Functional/ErrorHandlerTest.php
```diff
@@ -78,8 +83,6 @@ public function testErrorPages() {
     ];
 
     $routes = [
-      'apigee_edge_test.entity_storage_exception',
```
shadcn, line 81: @arlina-espinoza Should we still test these routes? Or are we completely replacing `apigee_edge_test`?

arlina-espinoza: We can remove them, as the `apigee_edge_test` isn't compatible with the mock client as it is (see https://github.com/apigee/apigee-edge-drupal/pull/389#discussion_r403183673).

arlina-espinoza (approved):
LGTM. Thanks @arshad .

PR #394, opened 2020-03-31
With m10n module enabled, non monetized API products do not follow RBAC

### LINKED ISSUE #392: "Visibility by Access" - Visibility attribute for API product not enforced correctly
I'm using the "visibility by access" capability in `Edge > API products` to control API product visibility through the internal/private/public attribute in Edge. This applies to an API which isn't part of a Product Bundle/Rate Plan (as using w/ monetization modules enabled), yet it's still showing up even though set to internal. Any ideas what could be missing / is it a bug?

Steps to reproduce: Create an API product with visibility set to Internal or Private, and leave those unchecked in the Apigee Edge > API products section. Then log in as an authenticated user, and create new app, and check if API product is visible from the list.

image (14)

cnovak:
The Apigee M10n module’s hook_ENTITY_TYPE_access uses the Apigee Monetization API to determine an API Product ‘assign’ access by calling `/developers/{developer}/eligible-products`, which returns AccessResult::allowed for all API products.

This API returns all API Products
1. That are able to be assigned to an App because the developer purchased a Rate plan that contains the API product
2. All API products that are not monetized.

Due to #2, the Apigee Edge module hook_ENTITY_TYPE_access needs to return AccessResult::forbidden when the operation is ‘assign’ and the user does not have the correct role to assign an API product to an app.

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -365,8 +365,9 @@ function apigee_edge_entity_view(array &$build, EntityInterface $entity, EntityV
  * for the API Product (entity) if the API product's access attribute value is
  * either one of the selected access attribute values OR if a developer
  * app is in association with the selected API product.
- * - If operation is "assign" then second part of the previous sentence does
- * not apply only access attribute's value matters.
+ * - If operation is "assign" then disallow access if the role is configured
+ * in the "Access by visibility" settings at the route
+ * apigee_edge.settings.developer.api_product_access.
  */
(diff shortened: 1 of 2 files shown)


cnovak:
Validated that API Product access tests pass

arlina-espinoza (approved):
Tested locally and LGTM. Thanks @cnovak :)

PR #399, opened 2020-04-20
Sync: prevent exception if username is too long

mxr576:
Developer sync does not handle properly if the username coming from Apigee Edge is longer than the default accepted length by Drupal (60 charachters as per [UserNameConstraintValidator](https://github.com/drupal/core/blob/8.8.x/modules/user/src/Plugin/Validation/Constraint/UserNameConstraintValidator.php#L56) and [UserInterface](https://github.com/drupal/core/blob/8.8.x/modules/user/src/UserInterface.php#L21)).

Before:
before

After:
after

### DIFF
--- a/src/Job/UserCreate.php
+++ b/src/Job/UserCreate.php
@@ -19,35 +19,12 @@
 
 namespace Drupal\apigee_edge\Job;
 
-use Drupal\apigee_edge\Exception\DeveloperToUserConversationInvalidValueException;
 use Drupal\apigee_edge\Structure\DeveloperToUserConversionResult;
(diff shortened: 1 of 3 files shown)


mxr576:
@arshad @arlina-espinoza @cnovak Hi, may I get a review on this?

cnovak:
Reviewing it now, thanks for screenshots for clarification

cnovak (approved):
LGTM, confirmed exception in logs

mxr576:
Thanks for the approval, I do not have permission to merge anymore, pls merge it.

shadcn:
Merged. Thank you 👍

PR #400, opened 2020-04-21
Validate field level constraints on form save

mxr576:
Field level constraints are not validated when an instance of a FieldableEdgeEntityForm gets submitted. As FieldableEdgeEntityForm if a fine-tuned copy of ContentEntityForm without being tied to an SQL entity storage, I copied the missing logic from the ContentEntityForm.

Also added a test coverage to the Developer App UI test because by default, this is the only entity with CU UI in the Apigee Edge module (submodules not included) and until this time all "fieldable edge entity" base classes were valided as part of its subclasses' tests, therefore there are no dedicated tests for "fieldable edge entity" base classes.

### DIFF
--- a/src/Entity/Form/FieldableEdgeEntityForm.php
+++ b/src/Entity/Form/FieldableEdgeEntityForm.php
@@ -120,7 +120,7 @@ abstract class FieldableEdgeEntityForm extends EntityForm implements FieldableEd
    * @return string[]
    *   An array of field names.
    *
-   * TODO Add missing return type-hint in 2.x.
+   * @todo Add missing return type-hint in 2.x.
    */
   protected function getEditedFieldNames(FormStateInterface $form_state) {
     return array_keys($this->getFormDisplay($form_state)->getComponents());
@@ -147,7 +147,7 @@ abstract class FieldableEdgeEntityForm extends EntityForm implements FieldableEd
    * @param \Drupal\Core\Form\FormStateInterface $form_state
    *   The current state of the form.
    *
-   * TODO Add missing return type-hint in 2.x.
+   * @todo Add missing return type-hint in 2.x.
    */
   protected function flagViolations(EntityConstraintViolationListInterface $violations, array $form, FormStateInterface $form_state) {
     // Flag entity level violations.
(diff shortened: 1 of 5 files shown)


mxr576:
> Drupal\Tests\apigee_edge\Functional\DeveloperSyncTest::testDeveloperSync
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'1'
+'0'

Based on the error message I think it was caused by a MGMT API communication error so I am considering this as a false alarm, but I cannot restart tests anymore.

mxr576:
@arshad @arlina-espinoza @cnovak Hi, may I get a review on this?

mxr576:
Bump, this is quite a blocker on some projects :)

arlina-espinoza (changes requested):
Thanks @mxr576 ! Tested it out locally and works as expected now. The only issue is that the following methods have signatures that are a breaking change for the M10n module. Could they be updated so they don't contain a return value? This is what the m10n module expects:

- public function validateForm(array &$form, FormStateInterface $form_state);
- protected function getEditedFieldNames(FormStateInterface $form_state);
- protected function flagViolations(EntityConstraintViolationListInterface $violations, array $form, FormStateInterface $form_state);

We'd need a follow up task on [apigee-m10n-drupal](https://github.com/apigee/apigee-m10n-drupal) to deprecate the `FieldableMonetizationEntityForm`, as it was previously filling in this functionality, but I don't think that can be done until a 2.x.

arlina-espinoza (approved):
Thanks for the changes @mxr576 :) It all looks good for me, just a note that the code sniffer validation is failing due to an unrelated Drupal bug that I'm trying to fix here: https://github.com/apigee/apigee-edge-drupal/pull/411

mxr576:
@arlina-espinoza Thanks, I saw it. Just let me know when I can rebase this PR.

arlina-espinoza:
@mxr576 #411 has been merged, so tests should be passing once you rebase. :+1:

arlina-espinoza:
@googlebot I consent.

PR #402, opened 2020-04-24
[#336] Add Apigee entities display settings

### LINKED ISSUE #336: Create a template for listing of the Apps Listing page
**Is your feature request related to a problem? Please describe.**
Today if someone wants to change the structure of App listing pages they have to extend the Drupal\apigee_edge\Entity\ListBuilder\DeveloperAppListBuilderForDeveloper class
which means they have to write a custom module. 

**Describe the solution you'd like**
A template that someone can override in the theme would be ideal which gives them more control to make changes.

Additionally can we ensure this change is also made applied to the apigee kickstart profile

shadcn:
Fixes #336 

This PR adds `apigee_edge.display_settings.*` config that any Apigee entity type can extend.

It provides a config form `apigee_edge.display_settings.*` that lets an admin configure the display type for an Apigee edge display. 

Example: the display type for the apps on the `/apps` page.

Display_settings___Apigee

This can be extended by any entity type that implements `EntityListBuilder`, i.e all the entity displays such as App, Team Apps, Teams can use this. This also eliminates custom implementation for Rate plans displays in m10n.

## Implementation

```yml
apigee_edge.display_settings.developer_app:
  path: '/admin/config/apigee-edge/app-settings/developer-apps/display-settings'
  defaults:
    _form: '\Drupal\apigee_edge\Form\EdgeEntityDisplaySettingsForm'
    _title: 'Display settings'
    entity_type_id: 'developer_app'
  requirements:
    _permission: 'administer apigee edge'
```

## Theming

When using the display settings, the following templates are available for theming.

`apigee-entity-list.html.twig`
`apigee-entity-list--ENTITY-TYPE.html.twig`
`apigee-entity-list--ENTITY-TYPE--VIEW-MODE.html.twig`

Example:

`apigee-entity-list.html.twig`
`apigee-entity-list--developer-app.html.twig`
`apigee-entity-list--developer-app--collapsible-card.html.twig`

~~This is WIP and still needs tests.~~ ✅ Tests added

### DIFF
--- a/modules/apigee_edge_teams/src/Entity/ListBuilder/TeamAppListByTeam.php
+++ b/modules/apigee_edge_teams/src/Entity/ListBuilder/TeamAppListByTeam.php
@@ -52,8 +52,6 @@ class TeamAppListByTeam extends AppListBuilder implements ContainerInjectionInte
    *   The entity type.
    * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
    *   The entity type manager.
-   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
-   *   The config factory.
    * @param \Drupal\Core\Render\RendererInterface $render
    *   The render.
    * @param \Symfony\Component\HttpFoundation\RequestStack $request_stack
@@ -62,9 +60,15 @@ class TeamAppListByTeam extends AppListBuilder implements ContainerInjectionInte
    *   The time service.
    * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
    *   The route match object.
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
+   *   The config factory.
    */
-  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager, ConfigFactoryInterface $config_factory, RendererInterface $render, RequestStack $request_stack, TimeInterface $time, RouteMatchInterface $route_match) {
-    parent::__construct($entity_type, $entity_type_manager, $config_factory, $render, $request_stack, $time);
+  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager, RendererInterface $render, RequestStack $request_stack, TimeInterface $time, RouteMatchInterface $route_match, ConfigFactoryInterface $config_factory = NULL) {
+    if (!$config_factory) {
+      $config_factory = \Drupal::service('config.factory');
+    }
+
+    parent::__construct($entity_type, $entity_type_manager, $render, $request_stack, $time, $config_factory);
     $this->routeMatch = $route_match;
   }
 
@@ -75,11 +79,11 @@ class TeamAppListByTeam extends AppListBuilder implements ContainerInjectionInte
     return new static(
       $entity_type,
       $container->get('entity_type.manager'),
-      $container->get('config.factory'),
       $container->get('renderer'),
       $container->get('request_stack'),
       $container->get('datetime.time'),
-      $container->get('current_route_match')
+      $container->get('current_route_match'),
+      $container->get('config.factory')
     );
   }
 
--- a/modules/apigee_edge_teams/src/Entity/ListBuilder/TeamAppListByTeam.php
+++ b/modules/apigee_edge_teams/src/Entity/ListBuilder/TeamAppListByTeam.php
@@ -103,6 +103,22 @@ class TeamAppListByTeam extends AppListBuilder implements ContainerInjectionInte
   }
 
   /**
+   * {@inheritdoc}
+   */
+  public function render() {
+    $build = parent::render();
+
+    // Update the empty text for table views.
+    if (!empty($build['table'])) {
+      $build['table']['#empty'] = t('There are no @label yet.', [
+        '@label' => $this->entityType->getPluralLabel(),
+      ]);
+    }
+
+    return $build;
+  }
+
+  /**
    * Returns the title of the "team app list by team" page.
    *
    * @return \Drupal\Core\StringTranslation\TranslatableMarkup
--- a/modules/apigee_edge_teams/src/Entity/ListBuilder/TeamAppListByTeam.php
+++ b/modules/apigee_edge_teams/src/Entity/ListBuilder/TeamAppListByTeam.php
@@ -22,6 +22,7 @@ namespace Drupal\apigee_edge_teams\Entity\ListBuilder;
 
 use Drupal\apigee_edge\Entity\ListBuilder\AppListBuilder;
 use Drupal\Component\Datetime\TimeInterface;
+use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
@@ -51,6 +52,8 @@ class TeamAppListByTeam extends AppListBuilder implements ContainerInjectionInte
    *   The entity type.
    * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
    *   The entity type manager.
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
+   *   The config factory.
    * @param \Drupal\Core\Render\RendererInterface $render
    *   The render.
    * @param \Symfony\Component\HttpFoundation\RequestStack $request_stack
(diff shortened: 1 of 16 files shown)


mxr576:
:+1: for the feature, but isn't is possible to use the same UIs as Drupal provides for other entity types: "Manage display"? With the field management UIs for fieldable entities we tried to follow the same pattern so users with Drupal experience can find what they are looking for easily.

arlina-espinoza:
Thanks @arshad , this is looking really good! I tested extending the templaates, as well as modifying the variables in a hook preprocess (`hook_preprocess_apigee_entity_list()`), and it all works nicely.
Some notes:
- For this module, I think we'd just need to add the display settings route to the teams listing page. You already covered adding it to the developer apps and team apps.
- @mxr576 Not sure I follow on your comment from above. The Edge entities still have the "manage form display" and "manage display" tabs to configure the display of the entity itself, what is being added is a way to configure the display of the listing of the entities. Drupal core does not provide that for other entities because, well, they use views for that, which we can't :sweat_smile: Perhaps an improvement would be to rename the tab to _"List display settings"_?
image

shadcn:
Thanks for the review @arlina-espinoza. I'm going to add same for teams listing.

@mxr576 Thanks. So we're still using core field ui and manage display. This is for the `EdgeEntityListBuilder`. It lets you configure which view mode to use for the entity lists.

shadcn:
> For this module, I think we'd just need to add the display settings route to the teams listing page. You already covered adding it to the developer apps and team apps.

@arlina-espinoza This is done. I've also added tests. Ready for review.

modules/apigee_edge_teams/src/Entity/ListBuilder/TeamAppListByTeam.php
```diff
@@ -60,8 +63,8 @@ class TeamAppListByTeam extends AppListBuilder implements ContainerInjectionInte
    * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
    *   The route match object.
    */
-  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager, RendererInterface $render, RequestStack $request_stack, TimeInterface $time, RouteMatchInterface $route_match) {
-    parent::__construct($entity_type, $entity_type_manager, $render, $request_stack, $time);
+  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager, ConfigFactoryInterface $config_factory, RendererInterface $render, RequestStack $request_stack, TimeInterface $time, RouteMatchInterface $route_match) {
+    parent::__construct($entity_type, $entity_type_manager, $config_factory, $render, $request_stack, $time);
```
arlina-espinoza: This is a breaking change. We should add the `$config_factory` as the last and optional parameter, and instantiate the service if missing during the construct call. See `UserController::_construct()` as an example: https://git.drupalcode.org/project/drupal/-/blob/8.8.x/core/modules/user/src/Controller/UserController.php#L60-97

modules/apigee_edge_teams/src/Entity/ListBuilder/TeamAppListByTeam.php
```diff
@@ -98,6 +102,22 @@ protected function buildEntityIdQuery(): QueryInterface {
...
+  /**
+   * {@inheritdoc}
+   */
+  public function render() {
+    $build = parent::render();
+
+    // Update the empty text for table views.
+    if (!empty($build['table'])) {
+      $build['table']['#empty'] = t('There are no @label yet.', [
+        '@label' => $this->entityType->getPluralLabel(),
```
arlina-espinoza, line 118:
```suggestion
        '@label' => strtolower($this->entityType->getPluralLabel()),
```

src/Entity/ListBuilder/AppListBuilder.php
```diff
@@ -61,22 +62,32 @@ class AppListBuilder extends EdgeEntityListBuilder {
...
    *   The renderer service.
    * @param \Symfony\Component\HttpFoundation\RequestStack $request_stack
    *   The request stack object.
    * @param \Drupal\Component\Datetime\TimeInterface $time
    *   The time service.
    */
-  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager, RendererInterface $renderer, RequestStack $request_stack, TimeInterface $time) {
-    parent::__construct($entity_type, $entity_type_manager);
+  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager, ConfigFactoryInterface $config_factory, RendererInterface $renderer, RequestStack $request_stack, TimeInterface $time) {
+    parent::__construct($entity_type, $entity_type_manager, $config_factory);
```
arlina-espinoza: See my comment on `TeamAppListByTeam::_construct()` about BC.

src/Entity/ListBuilder/DeveloperAppListBuilderForDeveloper.php
```diff
@@ -83,8 +86,8 @@ class DeveloperAppListBuilderForDeveloper extends AppListBuilder implements Cont
    * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
    *   The route match object.
    */
-  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager, RendererInterface $render, RequestStack $request_stack, TimeInterface $time, AccountInterface $current_user, RouteMatchInterface $route_match) {
-    parent::__construct($entity_type, $entity_type_manager, $render, $request_stack, $time);
+  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager, ConfigFactoryInterface $config_factory, RendererInterface $render, RequestStack $request_stack, TimeInterface $time, AccountInterface $current_user, RouteMatchInterface $route_match) {
+    parent::__construct($entity_type, $entity_type_manager, $config_factory, $render, $request_stack, $time);
```
arlina-espinoza: See my comment on `TeamAppListByTeam::_construct()` about BC.

src/Entity/ListBuilder/EdgeEntityListBuilder.php
```diff
@@ -31,24 +32,47 @@
...
    * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
    *   The entity type manager.
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
+   *   The config factory.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
    */
-  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager) {
+  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager, ConfigFactoryInterface $config_factory) {
```
arlina-espinoza: See my comment on `TeamAppListByTeam::_construct()` about BC.

src/Form/EdgeEntityDisplaySettingsForm.php
```diff
@@ -0,0 +1,223 @@
...
+      '#type' => 'select',
+      '#title' => $this->t('Type'),
+      '#default_value' => $config->get('display_type'),
+      '#required' => TRUE,
+      '#description' => $this->t('Select <em>Display mode</em> to configure a custom display.'),
+      '#options' => [
+        EdgeEntityListBuilder::DEFAULT_DISPLAY_TYPE => $this->t('Default'),
+        EdgeEntityListBuilder::VIEW_MODE_DISPLAY_TYPE => $this->t('Display mode'),
+      ],
+    ];
```
arlina-espinoza: Should we specify the "Default" is a table listing of entities, with links to entity operations?

arlina-espinoza (changes requested):
Thanks @arshad , this is looking pretty good. My main feedback is to fix the ListBuilder constructors to avoid a breaking change in the module (posted a link to an example in core).
The other thing I'm not sure is if we should call the tab "Display settings" or something else that makes it clearer that it's to configure the listing of entities, not the individual entities. Any thoughts on this @cnovak ?

arlina-espinoza:
@googlebot I consent.

arlina-espinoza:
@arshad I updated a test class and added a `tearDown()` to delete any data that was created if tests were run against a real API connection.

shadcn:
I've implemented the suggestions.

Thanks for the review @arlina-espinoza

arlina-espinoza (approved):
Thanks @arshad . Tested out locally and LGTM. :wink:

PR #407, opened 2020-05-01
[#396] Add a target bundle to first and last name user fields.

### LINKED ISSUE #396: First and Last Name fields can't be populated by cas_attributes module
**Describe the bug**
I'd like to be able to populate the "First Name" and "Last Name" fields for a user using CAS attribute tokens provided by the cas_attributes module. This is possible for "normal" fields like username/email, and "normal" custom fields -- I tested this out by adding a "Nickname" field that I was able to configure.

I think this is an issue related to how those fields are added to the user entity, but I have no idea how to approach this problem otherwise.

If this is the wrong repository for this issue, let me know and I can move it somewhere else.

**To Reproduce**
1. Create a new apigee kickstart site (this is how I encountered the issue)
2. Install and configure the [drupal/cas](https://www.drupal.org/project/cas) module (I don't think you need a working CAS server to test this)
3. Enable the `cas_attributes` module
4. Navigate to /admin/config/people/cas/attributes -- the "User Field Mappings" > "Fields" section should only show the email and username fields, and not first name and last name

**Expected behavior**
I would expect the "Fields" section in step four to also show "First Name" and "Last Name".

**Screenshots**
Here's what the user fields mapping section looks like:

Screenshot from 2020-04-07 16-25-08

**Additional context**
This also causes an issue where users whose accounts are created via a CAS sign-in don't have a first name and last name configured by default, which can then cause issues on subsequent sign-ins. Previously, in our D7 portal, we were able to use cas_attributes to automatically populate these values, which made the process transparent to users.

**Version Info**
1.8.0

arlina-espinoza:
At least one module seems to think that user fields that should be exposed for other configuration should have a target bundle set, as a way to distinguish the fields from other internal fields (such as uuid, roles, etc). This is at least the case with the_CAS Attributes_ module that allows setting remote attributes mapping to user fields. 
This PR adds the bundle to the first/last name fields.
As a note, fields added through the UI and saved into configuration do have the bundle set to "user", however, fields added through `hook_entity_base_field_info_alter()` do not (even if using `->setTargetBundle('user')`), I think that because by definition base fields apply to all bundles of an entity type.

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -280,6 +280,13 @@ function apigee_edge_entity_base_field_info_alter(&$fields, EntityTypeInterface 
     $mail = $fields['mail'];
     $mail->setRequired(TRUE);
     $mail->addConstraint('DeveloperMailUnique');
+
+    // Add a bundle to these fields to allow other modules to display them
+    // as configurable (fields added through the UI or configuration do have a
+    // target bundle set).
+    // @see https://github.com/apigee/apigee-edge-drupal/issues/396
+    $fields['first_name']->setTargetBundle('user');
+    $fields['last_name']->setTargetBundle('user');
   }
 }
 


cnovak (approved):
LGTM

PR #419, opened 2020-06-05
[#408] Fix team lists cache errors.

### LINKED ISSUE #408: API Consumers (Authenticated Users) are getting incorrect Teams(Companies)
**Incorrect Teams(List of Teams which user does not belong to) are displaying for API Consumers(Authenticated Users) on Team(Company) List** 
   
 
  | API Consumers | API Providers | Administrators
-- | -- | -- | --
Issue   occurring for | Yes | No | No
  
**To Reproduce**

To reproduce the issue, we need two users.

Let's Consider **ConsumerOne** and **ConsumerTwo** as two API Consumers(Authenticated user) in the portal.

_API Consumers have permission to create teams on the portal._

When **ConsumerOne** created a new team; the team list for the user **ConsumerOne** can see valid teams (list of the teams which belong to **ConsumerOne**). But other API Consumers are getting the same list of **ConsumerOne**. For Other API Consumers(including **ConsumerTwo**), those are invalid teams(Companies). 

**Notes:** 

1. This issue is happening only when API Consumers(Authenticated users) creating a new team.
2. When we are clearing the Drupal Portal cache the issue is disappearing.
3. Apigee Edge Team Caching set as 0/-1, still, the issue persists. 

**Expected behavior**
API Consumers(Authenticated users) should get relevant teams which he/she created or he/she belongs to.

**Version Info**
Drupal           : 8.7.12
Apigee Edge : 8.x-1.8

arlina-espinoza:
Fixes the caching issue reported on #408 by adding the current user to the render array context and cache tags.


modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
```diff
@@ -110,4 +110,18 @@ public function buildRow(EntityInterface $entity) {
...
 
+  /**
+   * {@inheritdoc}
+   */
+  public function render() {
+    $build = parent::render();
+
+    // Team lists vary for each user.
+    $build['#cache']['contexts'][] = 'user';
+    $build['#cache']['tags'][] = 'user:' . \Drupal::currentUser()->id();
```
mxr576: Adding the tags means the the page should be invalidated when the user with the given entity changes. I think you only need user context, which means the context of this page vary per user, so it should be cached that way.

modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
```diff
@@ -110,4 +110,18 @@ public function buildRow(EntityInterface $entity) {
...
+    $build['#cache']['max-age'] = 0;
```
mxr576: This kills the cache, so the above settings does nothing...

mxr576 (changes requested):

mxr576:
This also needs test coverage.

modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
arlina-espinoza: I think it should be kept, as the team list should be invalidated if the user's permissions change.

modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
arlina-espinoza: This brings up an interesting point. If we only cache per user context, then the cache will not be invalidated if a user's role is edited to have different permissions (for example, given the "administer teams" permission), or if team roles are edited. I just pushed an update that takes that into account, as well as the max-age from the team caching configuration.

modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
mxr576: I think the "user" context already covers that.

> Second, cache contexts are hierarchical in nature. The clearest example: when varying something per user, it's pointless to also vary that per permissions (i.e., the set of permissions that a user has), because per-user is already more granular. A user has a set of permissions, so per-user caching implies per-permissions caching.
Now for the most interesting aspect: if one part of the page is varied per user and another per permissions, then Drupal needs to be smart enough to make the combination of the two: only vary per user. That is where Drupal can exploit the hierarchy information to not create unnecessary variations.

https://www.drupal.org/docs/drupal-apis/cache-api/cache-contexts

modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
```diff
@@ -110,4 +110,33 @@ public function buildRow(EntityInterface $entity) {
...
+    $account = $this->entityTypeManager->getStorage('user')->load(\Drupal::currentUser()->id());
+
+    // Team lists vary for each user.
+    $build['#cache']['contexts'][] = 'user';
+
+    // This team list should be invalidated if the user's permissions change.
+    $build['#cache']['tags'][] = 'user:' . $account->id();
+
+    // This team list should be invalidated if the user roles permissions change.
+    foreach ($account->getRoles() as $rid) {
```
mxr576: Isn't it covered by the user:$id tag or the user context already? What happens if a user gets a new role and that should cause the invalidation of this page? The new role is of course not available in the list of the _current_ roles of the user.

mxr576 (changes requested):

modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
```diff
@@ -110,4 +110,33 @@ public function buildRow(EntityInterface $entity) {
     return $row + parent::buildRow($entity);
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function render() {
```
mxr576, line 117: Missing return typehint.

modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
mxr576: Based on the documentation is should cover it, based on the code it should not... :D

arlina-espinoza: Yeah, I don't want to assume something core does, and it doesn't hurt if the code is explicit about it.

mxr576: Your code does exactly the same thing as:

https://github.com/drupal/core/blob/8.8.x/lib/Drupal/Core/Cache/Context/AccountPermissionsCacheContext.php

So I would rather enable cache tag debuggin in my local and check the actual tags and context on the page whether they are sufficient by using the `user` and/org `user.permissions` cache contexts. Also, I would add a test coverage to warrant the UI works as it should even if something changes in core.

modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
arlina-espinoza: This class is overwriting a method from `Drupal\Core\Entity\EntityListBuilder` which doesn't provide a return typehint, so it just follows the same signature.

mxr576: Drupal core will not provide return type hints for years, but I thought that this contrib module start adding/using those and parameter typehints consistenly because it makes the code type-safe and easier to be debugged. I agree the current codebase already contains 70% return typehint only, this is what I try to improve with this comment.

modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
arlina-espinoza: Adding the `user` and `user.permissions` context was enough. :)

modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
arlina-espinoza: Yep, it turned out this was covered by the cache context.

arlina-espinoza:
This is ready for another review - failing tests will be fixed once #426 is in.

minnur (approved):
LGTM

PR #420, opened 2020-06-05
Adding Patch Firstname and Lastname must be of the type string null g…

### LINKED ISSUE #417: First Name and Last Name TypeError: Argument 1 passed to Drupal\apigee_edge\Entity\Developer::setFirstName() must be of the type string, null given
## Description
We are still seeing the First Name Last Name error come up. This time it happened on prod with the user 
## Steps to Reproduce
1. Go to 'Simplesaml Login'
2. Click on 'Enter to login'

## Actual Behavior
If login with the simple saml if null string in the first name and Last name is showing the below error.
The website encountered an unexpected error. Please try again later.
TypeError: Argument 1 passed to Drupal\apigee_edge\Entity\Developer::setFirstName() must be of the type string, null given, called in /home/site/wwwroot/modules/contrib/apigee_edge/src/UserDeveloperConverter.php on line 107 in Drupal\apigee_edge\Entity\Developer->setFirstName() (line 375 of modules/contrib/apigee_edge/src/Entity/Developer.php).
## Expected Behavior
If null string in the first name last name need to add default first name and Last name in the base field of first name and Last name. so check the module and added the condition of the base field Value is not NULL.

## Notes
https://www.drupal.org/project/apigee_edge
Added patch in the below reference module to fix the issues and need to add the patch in the upcoming version.
userdevelopercoverter php_file

## Version Info
https://www.drupal.org/project/apigee_edge
8.x-1.10 Stable release covered by the Drupal Security Team released 13 May 2020

ganesh-k4:
…iven

Fixes #417

### DIFF
--- a/src/UserDeveloperConverter.php
+++ b/src/UserDeveloperConverter.php
@@ -104,7 +104,10 @@ class UserDeveloperConverter implements UserDeveloperConverterInterface {
       $setter = 'set' . ucfirst($developer_prop);
       $getter = 'get' . ucfirst($developer_prop);
       if ($user->get($base_field)->value !== $developer->{$getter}()) {
+      // Checking the base field Value is not NULL.
+      if ($user->get($base_field)->value != NULL) {
         $developer->{$setter}($user->get($base_field)->value);
+      }
         $successful_changes++;
       }
     }


mxr576:
In which circumstances these fields could be null? There are required, they could be only unset for uid1 or if a custom code allowed creation of a user without calling $user->validate() before save that would return a list of violations, including that the required constraint has failed for these fields.

cnovak:
I just linked PR to issue #417 which explains the situation this happens

mxr576:
Thanks @cnovak , I explained in the issue why this PR cannot be accepted as a solution.

mxr576 (changes requested):
See my comment in the related issue.

arlina-espinoza:
Thanks @ganesh-k4, but I'm closing this PR per the comments on: https://github.com/apigee/apigee-edge-drupal/issues/417#issuecomment-643942367 Let us know if the workaround suggested works.

Issue #415, opened 2020-06-04
Apigee Edge admin page shows password in plaintext within source


akac:
## Description
As a security concern, the Apigee Edge plugin config should not have the password in plain text within the source. 

## Steps to Reproduce
1. Login as an admin
2. Goto the Apigee Edge config
3. View source

## Actual Behavior
Password is shown in plain text within the source

## Expected Behavior
No password is ever shown. It may be saved but never retrieved. Or if its retrieved its done via a GET call.

## Version Info
Machine name: apigee_edge
Version: 8.x-1.6

cnovak:
The Apigee Edge module uses the [Drupal Key module](https://www.drupal.org/project/key) to store the Edge credentials. If you install using Kickstart it will use the **Configuration Key Provider** since no other **Key Providers** are available. 

Check out [our FAQ on making the Edge connection more secure](https://www.drupal.org/docs/8/modules/apigee-developer-portal-kickstart/apigee-kickstart-faqs#s-during-installation-a-warning-is-displayed-that-the-apigee-edge-connection-key-provider-is-not-considered-secure-what-should-i-do) for instructions on how to update your site.

This issue should be more clear when installing and reading documentation.

# Tasks
- We have an [issue to add a warning to the Kickstart installer](https://github.com/apigee/apigee-devportal-kickstart-drupal/issues/305). It should be a higher priority to fix, will bump it up so that a message will be shown during installation process.
- Added [a warning message in the installation documentation](https://www.drupal.org/docs/contributed-modules/apigee-developer-portal-kickstart/get-started-with-kickstart#s-installation-instructions)
- Added [a warning message to project page](https://www.drupal.org/project/apigee_devportal_kickstart)

akac:
We are using the Apigee Edge: Private File - but that’s not the problem. The problem is not the key provider. The problem is that when you go to that page, if you view source of the web page - you can see password in plain text. 

```
<div class="js-form-item form-item js-form-type-password form-type-password js-form-item-key-input-settings-password form-item-key-input-settings-password">
      <label for="edit-key-input-settings-password">Password</label>
[...]

PR #434, opened 2020-07-06
[#416] Myapps endpoint toggle password.

### LINKED ISSUE #416: My Apps area should not retreive the key/secret at the same time
## Description
As a security concern, the My Apps area should not retreive the key/secret at the same time. When the user clicks the view icon, it should do a get of the key.

## Steps to Reproduce
1. Go to My Apps
2. View source

## Actual Behavior
You can see both the key/secret combo

## Expected Behavior
No security info seen. 

## Notes
We did a penetration test and these bugs were found as medium/high for their purposes.

## Version Info
Machine name: apigee_edge
Version: 8.x-1.6

minnur:
Related to https://github.com/apigee/apigee-edge-drupal/issues/415

I implemented new endpoint that returns JSON response with credentials for a given app. Credentials are getting pulled by app name and current user.


src/Routing/EdgeDynamicRouting.php
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Build routes dynamically.
+   */
+  public function routes() {
+    $route_collection = new RouteCollection();
+    $route = new Route('/admin/config/apigee-edge/app/{developer_app_by_name}/credentials',
+      [
+        '_controller' => '\Drupal\apigee_edge\Entity\ListBuilder\DeveloperAppListBuilderForDeveloper::appCredentials',
+        '_title' => 'App credentials',
+        'user' => \Drupal::currentUser()->id(),
```
mxr576: 1. Use DI.
2. Should not admins also have access to credentials of _all_ users? Since they have administer/bypass permission... 
3. This should also support team apps.

mxr576 (changes requested):

src/Routing/EdgeDynamicRouting.php
minnur: This is how it is currently pulling apps for current user. See Drupal\apigee_edge\ParamConverter\DeveloperAppNameConverter

src/Entity/ListBuilder/DeveloperAppListBuilderForDeveloper.php
```diff
@@ -215,4 +216,21 @@ public function pageTitle(): TranslatableMarkup {
...
 
+  /**
+   * Returns app credentials.
+   *
+   * @return \Symfony\Component\HttpFoundation\JsonResponse
+   *   The app credentials.
+   */
+  public function appCredentials($developer_app_by_name): JsonResponse {
+    $payload = [];
+    if ($credentials = array_shift($developer_app_by_name->getCredentials())) {
```
mxr576: The Apigee Edge module supports app with multiple credentials, this feature should not be removed with this change.

src/Routing/EdgeDynamicRouting.php
```diff
@@ -0,0 +1,58 @@
...
+ * Dynamically built routes.
+ */
+class EdgeDynamicRouting {
+
+  /**
+   * Build routes dynamically.
+   */
+  public function routes() {
+    $route_collection = new RouteCollection();
+    $route = new Route('/admin/config/apigee-edge/app/{developer_app_by_name}/credentials',
```
mxr576: Why should is an admin path?

src/Routing/EdgeDynamicRouting.php
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Dynamically built routes.
+ */
+class EdgeDynamicRouting {
+
+  /**
+   * Build routes dynamically.
+   */
+  public function routes() {
```
mxr576: Missing return typehint.

mxr576 (changes requested):

src/Entity/ListBuilder/DeveloperAppListBuilderForDeveloper.php
minnur: @arlina-espinoza do you have an example of the app with multiple credentials ?

src/Routing/EdgeDynamicRouting.php
mxr576: I know,I wrote it :) , but you can still use dependency injection here I think, you just need to implement the right interface. Also, probably it would be better introducing a route like the following if you would like to solve the 2nd point, user/[uid]/apps/[app]/api-keys/[api-key].
For the 3rd point, you need a team app spec endpoint.

src/Entity/ListBuilder/DeveloperAppListBuilderForDeveloper.php
mxr576: Go to Apigee Edge MGMT UI, can you can add more than one certificate to an app on the UI, those should be visible in the devportal too even if they cannot be created on the devportal UI by default.

src/Routing/EdgeDynamicRouting.php
minnur: good point. I made the change and added two routes:
/api-keys/developer/{uid}/{app_name}
/api-keys/team/{team}/{app_name}

src/Entity/ListBuilder/DeveloperAppListBuilderForDeveloper.php
minnur: Yes, found it and made the change to support multiple credentials.

src/Routing/EdgeDynamicRouting.php
minnur: Changed it and got rid of the class. Instead I put the route into the route file.

src/Routing/EdgeDynamicRouting.php
minnur: Class removed, no longer relevant.

minnur:
@arlina-espinoza @arshad do we have some code that would check user permission in teams ? I would like to add some routine that would make sure users cannot pull credentials for different developers or teams.

src/Controller/DeveloperAppKeysController.php
```diff
@@ -0,0 +1,98 @@
+<?php
+
+/**
+ * Copyright 2018 Google Inc.
```
arlina-espinoza:
```suggestion
 * Copyright 2020 Google Inc.
```

arlina-espinoza (changes requested):
> @arlina-espinoza @arshad do we have some code that would check user permission in teams ? I would like to add some routine that would make sure users cannot pull credentials for different developers or teams.

@minnur If you switch the paths to something like `/user/{user}/apps/{app}/api-keys` and `/teams/{team}/apps/{app}/api-keys`, you could take advantage of the route requirement `_app_access_check_by_app_name`: `view` to enforce a _view_ access permission. See [`DeveloperAppRouteProvider::getCanonicalRouteByDeveloper()`](https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/src/Entity/DeveloperAppRouteProvider.php#L172-L182)

src/Controller/DeveloperAppKeysController.php
minnur: thanks Arlina, I am working on this.

minnur: Resolved in b4cd0f2099578240972d594490f9c43e0e22bf89

arlina-espinoza:
@googlebot I consent.

arlina-espinoza (approved):
Thanks @minnur , I tested locally and fixed the tests. LGTM :+1:

PR #435, opened 2020-07-06
[#421] Show member role on team page

### LINKED ISSUE #421: Member role does not show up on team members page
On a team members page, the "member" role is assumed but unlisted. See example:

image

It would be better if it clearly called out the "member" role.

shadcn:
Fixes https://github.com/apigee/apigee-edge-drupal/issues/421

@arlina-espinoza I wonder if we need to load the default role into the storage or can we do something as simple as this?

### DIFF
--- a/modules/apigee_edge_teams/src/Controller/TeamMembersList.php
+++ b/modules/apigee_edge_teams/src/Controller/TeamMembersList.php
@@ -48,7 +48,7 @@ class TeamMembersList extends ControllerBase {
    *
    * @var array
    */
-  protected $defaultRoles;
+  protected $defaultRoles = [];
 
   /**
    * TeamMembersList constructor.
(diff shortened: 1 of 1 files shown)


modules/apigee_edge_teams/src/Controller/TeamMembersList.php
```diff
@@ -160,11 +160,14 @@ protected function buildRow(string $member, array $users_by_mail, array $team_me
       $row['data']['member'] = $member;
     }
 
+    // Default role assigned to all users.
+    $default_roles = [TeamRoleInterface::TEAM_MEMBER_ROLE => $this->t('Member')];
```
arlina-espinoza: I like this approach. However we do need to fetch the `TeamRoleInterface::TEAM_MEMBER_ROLE` label instead of hard-coding it to "member", because the role name might have been edited by an administrator to be something else.

arlina-espinoza (changes requested):

arlina-espinoza (approved):
LGTM :+1: I'll just wait for the tests to pass.

shadcn:
Some unrelated failing tests. I created a separate issue to track it: https://github.com/apigee/apigee-edge-drupal/issues/436

I'm going to merge this one.

PR #437, opened 2020-07-07
[#430] Add configurable cache max age for edge entity list builder

### LINKED ISSUE #430: Status of the App is not updated immediately on the App list
## Description
When someone logs in as an authenticated user (not admin) into a site, and changes the status of an App's product in Apigee Edge's UI, the status will not change on the App list as well. Meanwhile, it's refreshed on the App's page itself.
This might be confusing for a user.
For a user who is logged in to the site as an administrator, this works as intended.
None of the cache settings (API Product caching or Developer App settings cache) affects the outcome. 

## Steps to Reproduce
1. Create an App in Drupal, append a Product and a User to it.
2. Log in to the site with the corresponding user and go to the App list.
3. In Apigee Edge's UI, change any Product's status of the App (from Approved to Revoked or vice versa).
4. Refresh the App list page.

## Actual Behavior
The status in the App list remains unchanged, whilst if you click on the App's name and go to the App page itself, the status is changed. The status on the App list will only change if a global cache rebuild is applied.

## Expected Behavior
Change the status of the App on the App list as well.

## Screenshots
On the App list: https://i.imgur.com/ndKIA69.png (status is not changed)
On the App's page: https://i.imgur.com/mmpWq5A.png (status is changed)

## Version Info
Apigee Edge Drupal integration Module 8.x-1.7, Drupal core 8.8.6

shadcn:
Fixes #430

### DIFF
--- a/src/Entity/ListBuilder/EdgeEntityListBuilder.php
+++ b/src/Entity/ListBuilder/EdgeEntityListBuilder.php
@@ -124,12 +124,21 @@ class EdgeEntityListBuilder extends EntityListBuilder {
    * {@inheritdoc}
    */
   public function render() {
+    $build = parent::render();
+
     $settings = $this->getDisplaySettings();
     if ($this->usingDisplayType(static::VIEW_MODE_DISPLAY_TYPE)) {
-      return $this->renderUsingViewMode($settings['view_mode']);
+      $build = $this->renderUsingViewMode($settings['view_mode']);
     }
 
-    return parent::render();
+    // Add cache contexts.
+    $build['#cache'] = [
+      'contexts' => $this->entityType->getListCacheContexts(),
+      'tags' => $this->entityType->getListCacheTags(),
+      'max-age' => $this->getCacheMaxAge(),
+    ];
+
+    return $build;
   }
 
   /**
@@ -147,10 +156,6 @@ class EdgeEntityListBuilder extends EntityListBuilder {
       '#entities' => $this->load(),
       '#entity_type' => $this->entityType,
       '#view_mode' => $view_mode,
-      '#cache' => [
-        'contexts' => $this->entityType->getListCacheContexts(),
-        'tags' => $this->entityType->getListCacheTags(),
-      ]
     ];
   }
 
@@ -183,6 +188,16 @@ class EdgeEntityListBuilder extends EntityListBuilder {
     return $this->configFactory
       ->get("apigee_edge.display_settings.{$this->entityTypeId}")
       ->getRawData();
+  }
+
+  /**
+   * Returns the cache max age.
+   *
+   * @return int
+   *   The cache max age.
+   */
+  protected function getCacheMaxAge() {
+    return 0;
   }
 
 }
--- a/src/Entity/ListBuilder/EdgeEntityListBuilder.php
+++ b/src/Entity/ListBuilder/EdgeEntityListBuilder.php
@@ -124,11 +124,12 @@ class EdgeEntityListBuilder extends EntityListBuilder {
    * {@inheritdoc}
    */
   public function render() {
-    $build = parent::render();
-
     $settings = $this->getDisplaySettings();
     if ($this->usingDisplayType(static::VIEW_MODE_DISPLAY_TYPE)) {
       $build = $this->renderUsingViewMode($settings['view_mode']);
+    }
+    else {
+      $build = parent::render();
     }
 
     // Add cache contexts.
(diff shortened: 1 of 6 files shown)


src/Entity/ListBuilder/EdgeEntityListBuilder.php
```diff
@@ -185,4 +191,14 @@ protected function getDisplaySettings(): array {
...
   }
 
+  /**
+   * Returns the cache max age.
+   *
+   * @return int
+   *   The cache max age.
+   */
+  protected function getCacheMaxAge() {
+    return 0;
```
mxr576, line 201: How about providing a sane default?

```php
<?php
    $config = $this->configFactory->get("apigee_edge.{$this->entityTypeId}_app_settings");
    if (!$config->isNew()) {
      return  $config ->get('cache_expiration');
    }
   return 0;
```

src/Entity/ListBuilder/EdgeEntityListBuilder.php
```diff
@@ -126,10 +126,20 @@ protected function buildEntityIdQuery(): QueryInterface {
...
     }
+    else {
+      $build = parent::render();
+    }
+
+    // Add cache contexts.
+    $build['#cache'] = [
+      'contexts' => $this->entityType->getListCacheContexts(),
+      'tags' => $this->entityType->getListCacheTags(),
+      'max-age' => $this->getCacheMaxAge(),
```
mxr576, line 139: Isn't the entity type's default cache should be considered here or it should be the same as what is stored in the entity type specific config (by default)?

mxr576 (changes requested):

src/Entity/ListBuilder/EdgeEntityListBuilder.php
shadcn: I see but `cache_expiration` is not available for all entity types. It's up to the Edge entity list to implement the config for this and the list builder. So `0` makes sense here since we don't want to cache by default.

src/Entity/ListBuilder/EdgeEntityListBuilder.php
shadcn: Not sure what you mean here. Can you point me to `"what is stored in the entity type specific config (by default)"`? Thanks.

mxr576: Ohh, so there is no `getListCacheMaxAge()` on Drupal entity types by default, probably because of the same reason: https://github.com/apigee/apigee-edge-drupal/issues/430#issuecomment-649539244

so probably on Apigee Edge entity types, we should have a `getListCacheMaxAge()` method that returns the applicable max-age for an entity type specific list builders and it should retrieve the value from the related config.

src/Entity/ListBuilder/EdgeEntityListBuilder.php
mxr576: > I see but cache_expiration is not available for all entity types

let me challenge that... if this is true, why do we have a 'getCacheMaxAge()' in the base class if it not leverage by 100% of the sub-classes?

mxr576 (changes requested):

src/Entity/ListBuilder/EdgeEntityListBuilder.php
shadcn: Yeah. Only `getListCacheContexts` and `getListCacheTags` on `EntityInterface`.

We could implement a `getListCacheMaxAge` in `EdgeEntityInterface` but not all entity types defines a `list_builder`.

I think we can leave it on the base `EdgeEntityListBuilder` class.

src/Entity/ListBuilder/EdgeEntityListBuilder.php
```diff
@@ -126,10 +126,20 @@ protected function buildEntityIdQuery(): QueryInterface {
...
+      $build = $this->renderUsingViewMode($settings['view_mode']);
     }
+    else {
+      $build = parent::render();
+    }
+
+    // Add cache contexts.
+    $build['#cache'] = [
+      'contexts' => $this->entityType->getListCacheContexts(),
+      'tags' => $this->entityType->getListCacheTags(),
```
arlina-espinoza, line 138: The entity type list cache tags and contexts are not user aware, which means the render array will get cached the same for all of the users. We need to add the `user` and maybe also `user.permissions` to  the contexts to make the render array aware that the response changes per user.

arlina-espinoza (changes requested):
Looking good! Just check my comment about adding the user context. Thanks @arshad :)

src/Entity/ListBuilder/EdgeEntityListBuilder.php
shadcn: Added. Thanks.

arlina-espinoza (approved):
LGTM, thanks @arshad !

PR #446, opened 2020-07-16
[#432] Implement credentials deletion and generation

### LINKED ISSUE #432: API Key Creation/Rotation within one App
Similarly to what has been implemented for Drupal 7 here: https://github.com/apigeecs/apigee_apikey_regeneration, it would be valuable to have the ability to manage multiple API keys within one app. I have seen in many cases that the "App" is being renamed to "Keys" straight away due to the missing capability, so it does not confuse API consumers.

It is important for enterprises as keys need to be rotated (esp. for security reasons), and would be less of a disruption when this can be done within an existing app (with existing configuration tying it to a specific API Product) as opposed to creating a new app from scratch.

shadcn:
Fixes #432 

This PR implements credential generation, status update (revoke/approve) and deletion for key rotation.

[Update: Deleted outdated screenshots]


mxr576:
What about regenerating a customer secret on an existing credential?

modules/apigee_edge_teams/src/Entity/TeamAppRouteProvider.php
```diff
@@ -118,6 +138,87 @@ protected function getCollectionRouteByTeam(EntityTypeInterface $entity_type) {
...
+   * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
+   *   The entity type.
+   *
+   * @return \Symfony\Component\Routing\Route|null
+   *   The generated route, if available.
+   */
+  protected function getGenerateCredentialRoute(EntityTypeInterface $entity_type) {
+    if ($entity_type->hasLinkTemplate('generate-credential-form')) {
+      $route = new Route($entity_type->getLinkTemplate('generate-credential-form'));
+      $route->setDefault('_form', TeamAppCredentialGenerateForm::class);
```
mxr576: I think new permissions should be introduced for these new operations, like not every developer or team member should be able to destroy an existing/request new a API key.

shadcn: This is already being discussed. Thanks @mxr576

cnovak:
# Approve App

I am not sure, but I believe you may be adding "approve" into this PR which are used for API Providers to approve  credentials from developers. This should not be done since the API provider may set up manual approvals on an API Product, and with this change the developer could self approve the app.  

Check out what the internal portal is doing. They do not let users delete or approve credentials (they use "key" instead of "credentials" for this entity):

Screen Shot 2020-07-29 at 12 44 26 PM

# Generate Credentials

It seems odd to have a tab to create a new credential versus  an add button.

shadcn:
@cnovak OK so we remove **approve** and keep **revoke** or do we make revoke the delete action?

cnovak:
> What about regenerating a customer secret on an existing credential?

I created #459 to look into this @mxr576

cnovak:
@arshad here are details for you:
 
# Add Credentials
* The "Add Credentials" should be a button on the credentials listing page. 
* Ideally, a modal would come up to allow user to add details. 
* The add credentials form should not allow user to change the API Products for the app by default. However, add an admin setting to allow API products to be changed per credential. 

# Approve
* The approve functionality should be removed.

# Revoke
When a credential is revoked, it should follow the integrated portal concept of hiding the revoked credentials when you open the page. There should be a way to see revoked keys:

expand-collapse

* Currently the system will show a warning for apps that have revoked credentials. This should not be considered a warning state since revoking credentials is part of the maintenance of  an app.

# Delete
* The delete should be either removed or made as an optional admin setting to allow deletes on credentials.
* On an install, the delete functionality should default to not being shown.

shadcn:
@cnovak Thanks. I'll make the changes.

Question about the following:

> The add credentials form should not allow user to change the API Products for the app by default. However, add an admin setting to allow API products to be changed per credential.

OK so when a user adds new credentials, we use the API products from the app? What if we have two credentials with different products on the same app?

On the portal, API products are listed separately from API keys (which makes it look like API products are tied to the app and not keys).

But on both the Drupal portal and Edge UI, each key lists its own products.

## Integrated Portal

Test_postal_app___Arshad_Test

## Drupal
Hello_app___Apigee

## Edge
Apps_-_Apigee

arlina-espinoza:
@arshad I would use the API products of the first cred as the default, but let me circle back with the Apigee team to check.

arlina-espinoza:
@arshad We should take the API products from the most recently created credential that is active (not revoked or expired), and have those as the default products for new creds. There is a `getIssuedAt()` method on the credential entity that would help with that.

shadcn:
## Updates

✅ 1. Added the ability to create new API keys for an app.
✅ 2. The Add Key button opens form in a modal.
✅ 3. You can select the expiry date for the new API key.
✅ 4. When a new key is added, it is assigned API products from the most recently (issuedAt) active key.
✅ 5. Added revoke and delete key features.
✅ 6. Everything has permissions for access control.
✅ 7. Team-level permissions have been added too.
✅ 8. Revoked keys are hidden in a collapsible section.
✅ 9. Warnings for revoked keys removed.
✅ 10. If you have only one key on an app, it cannot be revoked or deleted.
✅ 11. Added permission to allow admin to edit API products for app. User cannot edit API products once created.

## Add API key form

apigee-add-key

## Developer App permissions

apigee-permissions

## Team App permissions

apigee-team-permissions

## Revoked keys

apigee-revoked-keys

css/apigee_edge.components.css
```diff
@@ -45,6 +45,11 @@
   vertical-align: super;
 }
 
+/* This fixes the ui datepicker position when displayed in a modal. */
+.ui-datepicker {
```
arlina-espinoza, line 49: Does this need to be targeted more specifically, something like `#drupal-modal  .ui-datepicker`?

arlina-espinoza (commented):
This is looking really good @arshad ! I feel we also need to add default permissions for new module installations or when updating to this version, ie, `Add API key to own Apps` and `Revoke API key from own Apps` should be given to the authenticated user by default.

src/Entity/Form/AppEditForm.php
```diff
@@ -145,6 +146,7 @@ public function form(array $form, FormStateInterface $form_state) {
           '#title' => $api_product_def->getPluralLabel(),
           '#required' => TRUE,
           '#options' => $credential_product_options,
+          '#disabled' => $this->canEditApiProducts(),
```
arlina-espinoza:
```suggestion
          '#disabled' => !$this->canEditApiProducts(),
```
This needs to be negated :stuck_out_tongue:

arlina-espinoza (commented):
Another issue I found was that when showing the key/secret, the revoked keys is using the key/secret from the active keys - like an index error somewhere related to this PR #434 . Attached example:
image

modules/apigee_edge_teams/src/DefaultTeamPermissionsProvider.php
```diff
@@ -74,6 +74,10 @@ public function permissions(): array {
           'update' => $this->t('Edit any Team Apps'),
           'delete' => $this->t('Delete any Team Apps'),
           'analytics' => $this->t('View analytics of any Team Apps'),
+          'add_api_key' => $this->t('Add API key to any Team Apps'),
+          'revoke_api_key' => $this->t('Revoke API key from any Team Apps'),
+          'delete_api_key' => $this->t('Delete API key from any Team Apps'),
+          'edit_api_products' => $this->t('Edit API products for any Team Apps'),
```
arlina-espinoza:
```suggestion
          'add_api_key' => $this->t('Add API key to Team Apps'),
          'revoke_api_key' => $this->t('Revoke API key from Team Apps'),
          'delete_api_key' => $this->t('Delete API key from Team Apps'),
          'edit_api_products' => $this->t('Edit API products for Team Apps'),
```
I feel the "any" word changes the meaning as if to say user with permission can manage the api keys regardless of team membership.

arlina-espinoza: Also think we need to add default permissions: `add_api_key` and `revoke_api_key` should be given to team administrators by default.

arlina-espinoza (changes requested):

modules/apigee_edge_teams/src/DefaultTeamPermissionsProvider.php
shadcn: Done. Added an update hook as well.

src/Entity/Form/AppEditForm.php
shadcn: Woah. Thanks @arlina-espinoza 👍

css/apigee_edge.components.css
shadcn: The `.ui-datepicker` is outside the modal.

shadcn:
> I feel we also need to add default permissions for new module installations or when updating to this version, ie, `Add API key to own Apps` and `Revoke API key from own Apps` should be given to the authenticated user by default.

@arlina-espinoza I assigned the `Add API key to own Apps` permissions to authenticated users. Do we need the revoke one as well?

minnur:
@googlebot I consent

minnur:
@arshad @arlina-espinoza I also made related fix in the Kickstarter https://github.com/apigee/apigee-devportal-kickstart-drupal/pull/408

arlina-espinoza:
@googlebot I consent.

arlina-espinoza (approved):
Thanks @arshad . I've manually tested this PR and it all looks really good :100: 
I only pushed an additional change after the latest conversation with the Apigee team, about the default permissions upon installing: developers should be able to add/revoke keys and edit api products in their own apps, and the same for team admins for their team apps. So the change was to add "edit_api_products developer_app" to the hook install/update.

arlina-espinoza:
@arshad I've fixed the new functional tests so that they are able to run both with mocks and as integration tests. As part of it, I removed the static app name and consumer key, so that they are able to run (possible concurrently) against the remote API.

PR #450, opened 2020-07-22
Replace autocomplete with textfield on "add team member" form.

arlina-espinoza:
Replaces the email autocomplete field on the "add team member" form with a textfield that accepts comma separated emails, to avoid displaying email addresses of users of the system.


shadcn (approved):
Approve. LGTM. Thanks @arlina-espinoza

PR #451, opened 2020-07-24
Adding http_client_proxy setting field

ngonendiaye:
image


cnovak:
Thanks for the submission @ngonendiaye. Can you add more details on this feature request and what problem it is solving? It looks like this PR only has the UI, are you planning on adding the implementation also?

ngonendiaye:
You're welcome. This feature request is for the UI only and allows easy configuration of the http_client_proxy variable for connections to Apigee edge.

ngonendiaye:
@googlebot I signed it!

arlina-espinoza (approved):
Thanks @ngonendiaye , it looks like we already had all the code and were just missing exposing the field for configuration. I just fixed some codesniffer validations, and updated the proxy field message to make it clear that not everyone will need to set this field.

cnovak:
@googlebot recheck CLA

arlina-espinoza:
Closing in favor of #475 because it has the signed CLA.

Issue #453, opened 2020-07-29
Optimize AppStorage::loadUnchanged


vitalie-cracan:
Apigee can be quite slow. Reducing the number of requests would help. I noticed that AppStorage::loadUnchanged($id) is making two identical requests. It does it because it needs to cover the possibility that $id can sometimes be an app name and not an app uuid.

I am new to the module and I have actually never seen a case when $id is an app name. Is this still a possible scenario?

Could we maybe pattern match the id and if it fits a uuid pattern then simply remove that id from cache? And if it does not fit the uuid pattern, only then have it load the app first to get the app id?

https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/src/Entity/Storage/AppStorage.php#L92

mxr576:
> Apigee can be quite slow. 

This is the reason why this multi-layered, extensive cache system have been invented to minimize the communication with the API backend.

> I have actually never seen a case when $id is an app name.

An app could be loaded several different ways:
* by [UUID](https://apidocs.apigee.com/management/apis/get/organizations/%7Borg_name%7D/apps/%7Bapp_id%7D)
* by owner id ([developer email](https://apidocs.apigee.com/management/apis/get/organizations/%7Borg_name%7D/developers/%7Bdeveloper_email_or_id%7D/apps/%7Bapp_name%7D-0), [company/team name](https://apidocs.apigee.com/management/apis/get/organizations/%7Borg_name%7D/companies/%7Bcompany_name%7D/apps/%7Bapp_name%7D)) + app name
* by owner uuid (developer uuid) + app name

These are the scenarios that we have to support, so technically yes, it is possible that an app is loaded by its name if the owner is in the context. 

Uncached load - in general - should be minimized, even if the storage is not Apigee. Is it a custom code where are you facing with performance issues or an OOTB UI provided by this module?

vitalie-cracan:
Thanks for your comment Dezső. The scenario I was referring to is reproducible on a clean standard Drupal install with only apigee_edge module enabled. One can see it when when one goes for the second, third, etc time to the edit page of an App, /user/{id}/apps/{app_name}/edit.

I've been a bit more careful now that you asked to describe how this happens and I see that after a Drupal cache clear, the callstack is slightly different, but the bottom line is that:

DeveloperAppNameConverter::convert() **always queries Apigee twice for a given app**. When cache was already warmed by:

[...]

PR #456, opened 2020-07-30
[#453]: pattern match the app id in loadUnchanged (#453)

vitalie-cracan:
Match the id against a uuid pattern and invalidate cache based on that id if matches.


mxr576:
Can you describe the issue that this change fixes? If there is a cache invalidation bug extending the test coverage would be also beneficial.

vitalie-cracan:
Based on feedback, I have done another pull request that solves the issue in a different way.

PR #457, opened 2020-07-30
[#415] Make sure no password is visible in the rendered configuration form

### LINKED ISSUE #415: Apigee Edge admin page shows password in plaintext within source

minnur:


cnovak:
I would rather follow the way SMTP module does the password where if you don't put it in it is left unchanged. This way if someone submits a change to the form they do not loose the current password. This could cause an outage on a production site.

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -259,6 +260,21 @@ public function validateConfigurationForm(array &$form, FormStateInterface $form
...
+      $key_input_plugin_form_state = clone $form_state;
+      $key_input_plugin_form_state->setValues($form_state->getUserInput()['key_input_settings']);
+      // @see \Drupal\key\Form\KeyFormBase::validateForm()
+      $key_input_processed_values = $form_state->getFormObject()->getEntity()->getKeyInput()->processSubmittedKeyValue($key_input_plugin_form_state);
+      $key_value = $key_input_processed_values['processed_submitted'];
+    }
+
+    // Could be an empty array.
+    $values = Json::decode($key_value);
+
```
arlina-espinoza: Thanks @minnur , this is looking good. However something around this section is giving me a white screen when checking the scenario of the module just installed (no credentials have ever been set). Could you try to replicate?

arlina-espinoza (changes requested):
See comment about the scenario of module just installed.

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
minnur: @arlina-espinoza looks like the module is generating WSOD even without my changes :(

minnur: @arlina-espinoza I think I got this working. Would you mind reviewing the changes?

arlina-espinoza:
@googlebot I consent.

arlina-espinoza (approved):
Thanks @minnur, LGTM. I've also fixed the tests.

PR #460, opened 2020-08-03
[#453]: added AppStorage::loadUnchangedById (#453)

### LINKED ISSUE #453: Optimize AppStorage::loadUnchanged

vitalie-cracan:
Fixes #453

### DIFF
--- a/src/Entity/Storage/AppStorage.php
+++ b/src/Entity/Storage/AppStorage.php
@@ -95,6 +95,30 @@ abstract class AppStorage extends AttributesAwareFieldableEdgeEntityStorageBase 
   }
 
   /**
+   * Load app by UUID.
+   *
+   * This function is more efficient than loadUnchanged(), because it does not
+   * need to cover the case when loading is done by App name.
+   *
+   * @param string $uuid
+   *   App UUID.
+   *
+   * @return \Drupal\apigee_edge\Entity\AppInterface|null
+   *   The unchanged entity, or NULL if the entity cannot be loaded.
+   *
+   * @TODO: this method should be also available in the AppStorageInterface, but
+   *   that would be a breaking change, so we can only add that in the next
+   *   major version of the module.
+   */
+  public function loadUnchangedByUuid(string $uuid): ?AppInterface {
+    // Clear the app controller's cache if it has one.
+    if ($this->appController instanceof EntityCacheAwareControllerInterface) {
+      $this->appController->entityCache()->removeEntities([$uuid]);
+    }
+    return parent::loadUnchanged($uuid);
+  }
+
+  /**
    * {@inheritdoc}
    */
   protected function initFieldValues(FieldableEdgeEntityInterface $entity, array $values = [], array $field_names = []) {
--- a/src/ParamConverter/DeveloperAppNameConverter.php
+++ b/src/ParamConverter/DeveloperAppNameConverter.php
@@ -20,6 +20,7 @@
 
 namespace Drupal\apigee_edge\ParamConverter;
 
+use Drupal\apigee_edge\Entity\Storage\AppStorage;
 use Drupal\apigee_edge\Exception\DeveloperDoesNotExistException;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Logger\LoggerChannelInterface;
@@ -87,7 +88,12 @@ final class DeveloperAppNameConverter implements ParamConverterInterface {
           // Load the entity directly from Apigee Edge if needed.
           // @see \Drupal\apigee_edge\ParamConverter\ApigeeEdgeLoadUnchangedEntity
           if (!empty($defaults['_route_object']->getOption('apigee_edge_load_unchanged_entity'))) {
-            $entity = $app_storage->loadUnchanged($app_id);
+            if ($app_storage instanceof AppStorage) {
+              $entity = $app_storage->loadUnchangedByUuid($app_id);
+            }
+            else {
+              $entity = $app_storage->loadUnchanged($app_id);
+            }
           }
           else {
             $entity = $app_storage->load($app_id);
--- a/modules/apigee_edge_teams/src/ParamConverter/TeamAppNameConverter.php
+++ b/modules/apigee_edge_teams/src/ParamConverter/TeamAppNameConverter.php
@@ -20,6 +20,7 @@
 
 namespace Drupal\apigee_edge_teams\ParamConverter;
 
+use Drupal\apigee_edge\Entity\Storage\AppStorage;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Logger\LoggerChannelInterface;
 use Drupal\Core\ParamConverter\ParamConverterInterface;
@@ -84,7 +85,12 @@ final class TeamAppNameConverter implements ParamConverterInterface {
         // Load the entity directly from Apigee Edge if needed.
         // @see \Drupal\apigee_edge\ParamConverter\ApigeeEdgeLoadUnchangedEntity
         if (!empty($defaults['_route_object']->getOption('apigee_edge_load_unchanged_entity'))) {
-          $entity = $app_storage->loadUnchanged($app_id);
+          if ($app_storage instanceof AppStorage) {
+            $entity = $app_storage->loadUnchangedByUuid($app_id);
+          }
+          else {
+            $entity = $app_storage->loadUnchanged($app_id);
+          }
         }
         else {
           $entity = $app_storage->load($app_id);


cnovak:
@en-vcracan thank you for your contribution. Could you sign a Contributor License Agreement (CLA), see above. Thanks

src/Entity/Storage/AppStorage.php
```diff
@@ -94,6 +94,27 @@ public function loadUnchanged($id) {
...
+   * because it does not need to cover the case
+   * when loading is done by App name.
+   *
+   * @param string $id
+   *   App id.
+   *
+   * @return \Drupal\Core\Entity\EntityInterface|mixed|null
+   *   The unchanged entity, or NULL if the entity cannot be loaded.
+   */
+  public function loadUnchangedById($id) {
```
mxr576: Missing parameter typehint (string) and return typehint: this is either null or AppInterface.

src/Entity/Storage/AppStorage.php
```diff
@@ -94,6 +94,27 @@ public function loadUnchanged($id) {
...
+   * Load unchanged for when $id is an App id.
+   *
+   * This function is more efficient than loadUnchanged,
+   * because it does not need to cover the case
+   * when loading is done by App name.
+   *
+   * @param string $id
+   *   App id.
+   *
+   * @return \Drupal\Core\Entity\EntityInterface|mixed|null
```
mxr576: 1. This is either null or AppInterface.
2. We need a TODO here because this method should be also available in the AppStorageInterface but that would be a breaking change so we can only add that in the next major version of the module.

src/Entity/Storage/AppStorage.php
```diff
@@ -94,6 +94,27 @@ public function loadUnchanged($id) {
     return parent::loadUnchanged($id);
   }
 
+  /**
+   * Load unchanged for when $id is an App id.
```
mxr576: I would rephase this: "Load app by UUID.".

src/Entity/Storage/AppStorage.php
```diff
@@ -94,6 +94,27 @@ public function loadUnchanged($id) {
...
+   *
+   * This function is more efficient than loadUnchanged,
```
mxr576: Nit: loadUnchanged()

src/Entity/Storage/AppStorage.php
```diff
@@ -94,6 +94,27 @@ public function loadUnchanged($id) {
...
+   * because it does not need to cover the case
+   * when loading is done by App name.
+   *
+   * @param string $id
```
mxr576: I would call the variable `$uuid`

src/Entity/Storage/AppStorage.php
```diff
@@ -94,6 +94,27 @@ public function loadUnchanged($id) {
...
+   *   App id.
```
mxr576: id => UUID.

mxr576 (changes requested):

src/Entity/Storage/AppStorage.php
vitalie-cracan: added typehints; now the only function that does it in AppStorage.

src/Entity/Storage/AppStorage.php
vitalie-cracan: Corrected. Technicall loadUnchanged() then should not simply say @inheritdoc, because it as well returns an AppInterface then.

src/Entity/Storage/AppStorage.php
vitalie-cracan: done

src/Entity/Storage/AppStorage.php
vitalie-cracan: np

src/Entity/Storage/AppStorage.php
vitalie-cracan: right

src/Entity/Storage/AppStorage.php
vitalie-cracan: changed

vitalie-cracan:
@cnovak thx, I am working on it. I am having to decide if I want to sign it as part of an organisation or not and it takes some time.

src/Entity/Storage/AppStorage.php
mxr576: That is the pattern the Drupal core also follows, I agree that an entity type-specific storage always narrows down the possible return values. I think, at least on custom methods, we should be as specific about the possible return values as we can, on inherited methods we should follow the pattern that we see in Drupal core, from the context it becomes quite clear that possible return value and it could be explicitly taught to the IDE too `/** @var ..../AppInterface $entity`

src/Entity/Storage/AppStorage.php
```diff
@@ -94,6 +94,31 @@ public function loadUnchanged($id) {
...
+   *   App UUID.
+   *
+   * @return \Drupal\apigee_edge\Entity\AppInterface|null
+   *   The unchanged entity, or NULL if the entity cannot be loaded.
+   *
+   * @TODO: this method should be also available in the AppStorageInterface,
+   *   but that would be a breaking change,
+   *   so we can only add that in the next major version of the module.
+   */
+  public function loadUnchangedById(string $uuid): ?AppInterface {
```
mxr576: If everything is "UUID" now, should not this be also `loadUnchangedByUuid`?

src/Entity/Storage/AppStorage.php
```diff
@@ -94,6 +94,31 @@ public function loadUnchanged($id) {
...
+   * when loading is done by App name.
+   *
+   * @param string $uuid
+   *   App UUID.
+   *
+   * @return \Drupal\apigee_edge\Entity\AppInterface|null
+   *   The unchanged entity, or NULL if the entity cannot be loaded.
+   *
+   * @TODO: this method should be also available in the AppStorageInterface,
+   *   but that would be a breaking change,
```
mxr576: Every line should be 80 characters long max, you can still add some text to this one. Please fix it.

src/Entity/Storage/AppStorage.php
```diff
@@ -94,6 +94,31 @@ public function loadUnchanged($id) {
     return parent::loadUnchanged($id);
   }
 
+  /**
+   * Load app by UUID.
+   *
+   * This function is more efficient than loadUnchanged(),
+   * because it does not need to cover the case
```
mxr576: Every line should be 80 characters long max, you can still add some text to this one. Please fix it.

mxr576 (changes requested):

src/Entity/Storage/AppStorage.php
vitalie-cracan: I am not sure it should be, it would definitely be inconsistent with rest of the code. This module uses id instead of uuid all over the place.

This line: https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/src/ParamConverter/DeveloperAppNameConverter.php#L90

would have to become:            
$entity = $app_storage->loadUnchangedByUuid($app_id);

The module has variables like $app_id, when in fact, according to this thinking, it should be $app_uuid. Similarly, you have $developer_id, etc

https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/src/ParamConverter/DeveloperAppNameConverter.php#L78

src/Entity/Storage/AppStorage.php
vitalie-cracan: fixed

src/Entity/Storage/AppStorage.php
vitalie-cracan: fixed

src/Entity/Storage/AppStorage.php
shadcn: @en-vcracan I see what you mean. Though having a `loadUnchanged($id)` and a `loadUnchangedById` which accepts a `uuid` can be misleading.

shadcn (commented):
@en-vcracan This is looking good. I added a comment about the method naming. 

Could you also sign the CLA please?

Thank you.

src/Entity/Storage/AppStorage.php
vitcra: @arshad we can go back and rename $uuid to $id, that's no problem. I myself think that would be best. But if you confirm you want this to be uuid, then I will change the function name. Let me know.

vitalie-cracan: @arshad we can go back and rename $uuid to $id, that's no problem. I myself think that would be best. But if you confirm you want this to be uuid, then I will change the function name. Let me know.

shadcn: @en-vcracan Yes. Let's rename it to `loadUnchangedByUuid`. Thank you.

src/ParamConverter/DeveloperAppNameConverter.php
```diff
@@ -87,7 +88,12 @@ public function convert($value, $definition, $name, array $defaults) {
           // Load the entity directly from Apigee Edge if needed.
           // @see \Drupal\apigee_edge\ParamConverter\ApigeeEdgeLoadUnchangedEntity
           if (!empty($defaults['_route_object']->getOption('apigee_edge_load_unchanged_entity'))) {
-            $entity = $app_storage->loadUnchanged($app_id);
+            if ($app_storage instanceof AppStorage) {
+              $entity = $app_storage->loadUnchangedById($app_id);
+            }
+            else {
+              $entity = $app_storage->loadUnchanged($app_id);
```
shadcn, line 95: I think this bit could use some work. What is `$app_storage` going to be in the alternate? `$app_storage->loadUnchanged`is going to fail if `$app_storage` is not an instance of `AppStorage` right?

vitalie-cracan: It is not going to fail, that alternative is the current only alternative. It should be replaced with loadUnchangedByUuid to make it more efficient, but this new method is available only on AppStorage. There is not hard guarantee that $app_storage will always be an AppStorage, thus the check.

See this note by @mxr576 : "We need a TODO here because this method should be also available in the AppStorageInterface but that would be a breaking change so we can only add that in the next major version of the module."  https://github.com/apigee/apigee-edge-drupal/pull/460#discussion_r482836296

shadcn: I see. Thank you.

shadcn: A `if (method_exists($app_storage, 'loadUnchangedById'))` might work as well.

vitalie-cracan: Could you be more specific, yes it might work, but current also works. Does your comment has a particular purpose? Why is that better?

src/Entity/Storage/AppStorage.php
vitalie-cracan: Renamed

vitalie-cracan:
@googlebot I signed it!

shadcn:
@en-vcracan Looks like code sniffer is failing. We fixed it upstream `8.x-1.x`. Can you rebase please? Thank you

vitalie-cracan:
@arshad, I have rebased and have rebased my own commits to clean up the history, but the lint is still failing.

shadcn:
@en-vcracan This is looking great. Thank you.

The failing tests are unrelated "Reflection error: Drupal\Tests\PhpunitCompatibilityTrait not found.". We have other PRs failing as well. We are going to fix then merge this PR.

This is a nice addition to the module. Thank you.

arlina-espinoza:
@googlebot I consent.

arlina-espinoza:
We have fixed the failing tests in 8.x-1.x, merging here so tests can pass before approving.

arlina-espinoza (approved):
LGTM, and all tests are passing. Thanks!

PR #462, opened 2020-08-04
[#461] Quickedit compatibility.

### LINKED ISSUE #461: Problem integrating Stripe Commerce
## Description
I try to implement Stripe Commerce on Apigee on Drupal 8. Whenever I try to add some credit to my account I get the error "We have accountered an unexpected error while analyzing your payment method. Please try later"

## Steps to Reproduce
1. Go to Pricing & Plans
2. Click on a Product, then purchase it.
3. Funds are insufficient so click on Add credit
4. Then add the needed funds
5. Continue
6. Purchase

## Actual Behavior
I get the error : "We have accountered an unexpected error while analyzing your payment method. Please try later" and no fund is added.
I get the following error on the portal reports : You cannot create a charge with a PaymentMethod. Use the Payment Intents API instead.

And the following errors on the error.log of the nginx : 
[error] 12427#0: *1371 FastCGI sent in stderr: "PHP message: Error: Call to undefined method Drupal\apigee_m10n\Entity\RatePlan::hasTranslation() in /opt/portal/web/core/modules/quickedit/src/QuickEditController.php on line 150 #0 [internal function]: Drupal\quickedit\QuickEditController->metadata(Object(Symfony\Component\HttpFoundation\Request))
apigee/apigee-m10n-drupal#1 /opt/portal/web/core/lib/Drupal/Core/EventSubscriber/EarlyRenderingControllerWrapperSubscriber.php(123): call_user_func_array(Array, Array)
apigee/apigee-m10n-drupal#2 /opt/portal/web/core/lib/Drupal/Core/Render/Renderer.php(573): Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscriber->Drupal\Core\EventSubscriber\{closure}()
apigee/apigee-m10n-drupal#3 /opt/portal/web/core/lib/Drupal/Core/EventSubscriber/EarlyRenderingControllerWrapperSubscriber.php(124): Drupal\Core\Render\Renderer->executeInRenderContext(Object(Drupal\Core\Render\RenderContext), Object(Closure))
apigee/apigee-m10n-drupal#4 /opt/portal/web/core/lib/Drupal/Core/EventSubscriber/EarlyRenderingControllerWrapperSubscriber.php(97): Drupal\Core\EventSubscriber\EarlyRenderingControllerWrapperSubscr" while reading response header from upstream, client: 10.23.77.31, server: 10.23.75.222, request: "POST /quickedit/metadata HTTP/1.1", upstream: "fastcgi://unix:/run/php-fpm/php-fpm.sock:", host: "10.23.75.222", referrer: "http://10.23.75.222/user/1/plans"

## Expected Behavior
Money should be added to my account

## Version Info
Drupal core : 8.9.0
Apigee Edge 8.x-1.10
Drupal Commerce 8.x-2.19
Commerce Stripe : I have tried 3 versions  : 8.x-1.0-rc2, 8.x-1.0-rc3 and 8.x-1.0-rc4

arlina-espinoza:
Fixes #461 . The issue is that core's quickedit module assumes all content extends `ContentEntityBase`, which is not the case with Apigee Edge entities. This PR adds a couple of methods that are called in the quickedit module to make it compatible. 

Note: The best fix would be for the quick edit module to check if an entity is translatable or not before calling these methods (there is an [issue with a patch for that](https://www.drupal.org/project/drupal/issues/3082529)), but given that there is not much interest in core to account for edge use cases, it makes sense to fix it in the module.

### DIFF
--- a/src/Entity/EdgeEntityBase.php
+++ b/src/Entity/EdgeEntityBase.php
@@ -161,4 +161,22 @@ abstract class EdgeEntityBase extends EntityBase implements EdgeEntityInterface 
     return FALSE;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getTranslation($langcode) {
+    if ($langcode == $this->language()->getId()) {
+      return $this;
+    }
+
+    throw new \InvalidArgumentException("Invalid translation language ($langcode) specified.");
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function hasTranslation($langcode) {
+    return $langcode == $this->language()->getId();
+  }
+
 }


shadcn (approved):
Ah thanks. Finally we can get rid of this notice. Thanks @arlina-espinoza LGTM

PR #477, opened 2020-09-07
[#223] Fixed apps title.

### LINKED ISSUE #223: "Apps" menu links does not respect the plural label of the developer apps entity anymore
Before #183 the "My Apps" menu link title always reflected the plural label (alias) of the developer app entity. For example, if the alias was "Foo" (singular) and "Foos" (plural) then the title of the "My Apps" menu item was "My Foos". Since #183 it is hard-coded to "Apps".
Even if the "My" prefix has been removed, it should respect and display the plural label of the developer app entity.

VladimirAus:
Changes title functionality inline with `apigee_edge_teams` submodule.

### DIFF
--- a/apigee_edge.api.php
+++ b/apigee_edge.api.php
@@ -1,0 +1,23 @@
+<?php
+
+/**
+ * @file
+ * Hooks for apigee_edge module.
+ */
+/**
+ * @addtogroup hooks
+ * @{
+ */
+
+/**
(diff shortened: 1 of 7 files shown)


arlina-espinoza:
@googlebot I consent.

arlina-espinoza (approved):
Looks good to me, thanks @VladimirAus for submitting this! I just fixed a couple of codesniffer validations and added a test to avoid regressions.

PR #488, opened 2020-09-15
[#455] Check for config id for key provider

### LINKED ISSUE #455: Incorrect warning message on the status report page about Edge connection credentials
## Description

After a `8.x-1.10` to `8.x-1.12` update I noticed that a new warning message appeared on the status report page which says:

> Edge connection settings are stored in Drupal’s configuration system, which is not designed to store sensitive information. When installing Kickstart for uses other than local development, we highly recommend changing the Apigee Edge connection key provider to a more secure storage location. Learn more.

The problem is that in my case the credentials are coming from environment variables defined in `settings.local.php` by `putenv()`. I double checked it, no credentials are stored elsewhere.

## Steps to Reproduce

1. Define the environment variables in `settings.local.php` using `putenv()`.
2. Visit the Edge credentials settings page and check the connection status: it should be ok (screenshot 1).
3. Check the status report page to see the incorrect warning message (screenshot 2).

## Actual Behavior

On the status report page the module incorrectly shows a misleading warning message that the Edge credentials are stored in Drupal's configuration system, meanwhile they are stored in environment variables.

## Expected Behavior

The above mentioned warning message shouldn't appear on the status report page if the Edge credentials are set correctly.

## Screenshots

apigee_edge_8-x-1-12-edge-credentials
apigee_edge_8-x-1-12-status-report-warning

## Version Info

- Drupal core 8.8.8
- Apigee Edge 1.12

shadcn:
Fixes #455

### DIFF
--- a/apigee_edge.install
+++ b/apigee_edge.install
@@ -71,7 +71,7 @@ function apigee_edge_requirements($phase) {
     $auth_config = \Drupal::config('apigee_edge.auth');
     if ($key = $auth_config->get('active_key')) {
       $key_repository = \Drupal::service('key.repository')->getKey($key);
-      if ($key_repository && $key_repository->getKeyProvider()->getPluginId()) {
+      if ($key_repository && $key_repository->getKeyProvider()->getPluginId() === "config") {
         $requirements['apigee_edge_insecure_config_key_provider'] = [
           'title' => t('Apigee Edge'),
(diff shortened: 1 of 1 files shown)


arlina-espinoza (approved):
LGTM, thanks @arshad. And thanks @balazswmann for reporting it.

PR #490, opened 2020-09-15
#478 Enable CI testing for D9 and fix deprecations

### LINKED ISSUE #478: Enable CI testing matrix against both D8 and D9
Enable CI testing matrix against both D8 and D9, fixing any incompatibilities found. 
Note: groundwork on the CircleCI config file was already made in: #404 .

arlina-espinoza:
Closes #478. This PR:
- Enables the testing matrix against both D8 and D9.
- Fixes some tests and deprecations, so the `apigee_edge` module and submodules can run in both D8 and D9 (with the exception of `apigee_edge_actions`, which is only for D8 because the `rules` module still doesn't have a D9 release).

Example of the test matrix in CircleCI:
image


shadcn (approved):
LGTM. This is a big win! Thanks @arlina-espinoza

PR #491, opened 2020-09-23
Use phpstan/phpstan 0.12.42 due to bug after phpstan release

arlina-espinoza:
Use phpstan/phpstan 0.12.42 due to bug that got revealed after the latest release of phpstan/phpstan 0.12.43. 
Full report here: https://github.com/mglaman/drupal-check/issues/186


shadcn (approved):
Thanks

PR #502, opened 2020-10-01
[#494] Add multiple_products field to config form

### LINKED ISSUE #494: Add ability to limit an app to one API Product in UI
There is an ability to limit an app to one API Product through configuration, but it is not displayed in the UI: https://www.drupal.org/docs/8/modules/apigee-edge/configure-app-registration#allowing-app-developers-to-select-a-single-API-product-only

This task is to add this to the UI so that users understand it is available.

shadcn:
Fixes #494 

General_app_settings___Apigee


arlina-espinoza (approved):
LGTM, thanks @arshad :)

PR #503, opened 2020-10-01
[#486] Add labels to filters on Apps -> Analytics for accessibility.

### LINKED ISSUE #486: A11Y: Analytics form is missing labels on form controls
## Description
Accessibility check fails App/Analytics form:

WCAG 1.3.1, WCAG 4.1.2
path: #edit-metrics, #edit-quick-date-picker

Fix ONE of the following:
- aria-label attribute does not exist or is empty
- aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
- Form element does not have an implicit (wrapped) <label>
- Form element does not have an explicit <label>
- Element has no title attribute or the title attribute is empty

image

zakiya:
Fixes missing labels on filters at user/*/apps/[appname]/analytics.

### DIFF
--- a/src/Form/AppAnalyticsFormBase.php
+++ b/src/Form/AppAnalyticsFormBase.php
@@ -147,6 +147,8 @@ abstract class AppAnalyticsFormBase extends FormBase {
         'error_count' => $this->t('Error count'),
       ],
       '#default_value' => 'total_response_time',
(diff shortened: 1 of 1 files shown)


arlina-espinoza (approved):
LGTM :+1:

PR #504, opened 2020-10-02
487 Remove attributes that fail validation on required fieldsets. 

### LINKED ISSUE #487: A11Y: API Edge: invalid aria attribute
## Description
WCAG 4.1.1, WCAG 4.1.2

Path: --wrapper
Fix the following:
-ARIA attribute is not allowed: aria-required="true"

When creating a new App, API products checkboxes' aria attribute is invalid

zakiya:
This problem comes from core. I also opened an issue here: https://www.drupal.org/project/drupal/issues/3174459

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -1782,3 +1782,19 @@ function apigee_edge_app_listing_page_title(): TranslatableMarkup {
 
   return $title;
 }
+
+/**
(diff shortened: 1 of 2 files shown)


arlina-espinoza (approved):
LGTM, thanks @zakiya :)

PR #505, opened 2020-10-06
Add the ability to use Google Compute Engine Service Account for Authentication

giteshk:
For Apigee Hybrid, portals hosted on the Google compute engines should be able to use the default service account associated with the instance. 
Apigee-Hybrid-GCE-SericeAccount


src/Plugin/EdgeKeyTypeBase.php
```diff
@@ -166,4 +170,11 @@ public function getAccountKey(KeyInterface $key): array {
     return $json;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function useGcpDefaultServiceAccount(KeyInterface $key): bool {
+    return $key->getKeyValues()['gcp_hosted'] ? TRUE : FALSE;
```
arlina-espinoza: Using `empty` prevents throwing a notice if the value is not set on the key.
```suggestion
    return $key->getKeyValues()['gcp_hosted'] ? TRUE : FALSE;
    return !empty($key->getKeyValues()['gcp_hosted']);
```

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -119,15 +123,33 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
...
       $form['password']['#states']['required'] = [$state_for_public, $state_for_private];
     }
+
+    $state_for_not_gcp_hosted = [];
+
+    if (GceServiceAccountAuthentication::isAvailable()) {
+      $form['gcp_hosted'] = [
+        '#type' => 'checkbox',
+        '#title' => $this->t('Use the default service account if this portal is hosted on GCP'),
+        '#description' => $this->t("Please ensure you have added 'Apigee Developer Administrator' role to the default compute engine service account hosting this portal."),
```
arlina-espinoza, line 136: I'm not familiar enough with this, but is "Apigee Developer Administrator" the only way the role can be named?

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -266,6 +288,9 @@ public function processSubmittedKeyValue(FormStateInterface $form_state) {
         $input_values['authorization_server'] = '';
         $input_values['client_id'] = '';
         $input_values['client_secret'] = '';
+        if ($input_values['gcp_hosted'] === '1') {
+          $input_values['account_json_key'] = '';
+        }
```
arlina-espinoza: This will never be true, because it's within an _if_ that only evaluates if `$input_values['gcp_hosted']` is unset or false. We may need to move this logic outside this scope.

arlina-espinoza (changes requested):
Thanks @giteshk, I have not tested the PR yet, but made some initial observations.

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
giteshk: This is what they named it. 

Documented here => https://www.drupal.org/docs/contributed-modules/apigee-edge/configure-the-connection-to-apigee-edge#s-set-the-service-account-roles

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
giteshk: I have addressed this in the recent code push

src/Plugin/EdgeKeyTypeBase.php
giteshk: Thank you I fixed this.

src/Connector/GceServiceAccountAuthentication.php
```diff
@@ -0,0 +1,89 @@
+<?php
+
+/**
+ * Copyright 2019 Google Inc.
```
mxr576: Since this is a new file the copyright header should be 2020.

src/Connector/GceServiceAccountAuthentication.php
```diff
@@ -0,0 +1,89 @@
...
+      throw new HybridOauth2AuthenticationException($e->getMessage(), $e->getCode(), $e);
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function authClient(): ClientInterface {
+    /** @var \Drupal\apigee_edge\SDKConnectorInterface $sdk_connector */
+    $sdk_connector = \Drupal::service('apigee_edge.sdk_connector');
```
mxr576, line 36: Can we use dependency injection here instead somehow?

src/Connector/GceServiceAccountAuthentication.php
```diff
@@ -0,0 +1,89 @@
...
+   * Validate if the GCE Auth Server URL is reachable.
+   *
+   * This is only available on GCP.
+   *
+   * @return bool
+   *   If GCE Service Account authentication is available.
+   */
+  public static function isAvailable(): bool {
+    try {
+      $client = \Drupal::httpClient();
```
mxr576: Can we use dependency injection here instead somehow?

mxr576 (changes requested):
This PR already contains several "Fix Code Sniffer Errors" commits. Please do not forget to clean the commit history before this PR gets merged or merge with "squash".

src/Connector/GceServiceAccountAuthentication.php
```diff
@@ -0,0 +1,89 @@
...
+use Apigee\Edge\Exception\HybridOauth2AuthenticationException;
+use Apigee\Edge\HttpClient\Plugin\Authentication\AbstractOauth;
+use Apigee\Edge\HttpClient\Plugin\Authentication\OauthTokenStorageInterface;
+use GuzzleHttp\Exception\ConnectException;
+use Http\Message\Authentication\Header;
+
+/**
+ * Decorator for Hybrid authentication plugin.
+ */
+class GceServiceAccountAuthentication extends AbstractOauth {
```
mxr576: Hm, my gut says that this authentication method should be added to the [PHP API client ](https://github.com/apigee/apigee-client-php)as the Drupal module is just one of the possible PHP consumers.

(Did we forget to add support for the authentication available on Hybrid to the PHP API client? :thinking: )

giteshk: Good catch I added moved over a lot of the code to the PHP API Client.

mxr576: Just for the reference :) https://github.com/apigee/apigee-client-php/pull/107

src/Connector/GceServiceAccountAuthentication.php
giteshk: Moved this to sdk

src/Connector/GceServiceAccountAuthentication.php
giteshk: Will add a code cleanup ticket for this. Following what was done for HybridAuthentication 

Don't want to add additional modifications to this.

arlina-espinoza (approved):
LGTM. Pushed an update to require the (to-be-released) ` apigee/apigee-client-php` 2.0.6, because it needs the `\Apigee\Edge\HttpClient\Plugin\Authentication\GceServiceAccount` class from it.

arlina-espinoza:
@googlebot I consent.

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -119,15 +122,33 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
     if (empty($values['organization'])) {
       $form['password']['#states']['required'] = [$state_for_public, $state_for_private];
     }
+
+    $state_for_not_gcp_hosted = [];
+    $gceServiceAccountAuth = new GceServiceAccountAuthentication(\Drupal::service('apigee_edge.authentication.oauth_token_storage'));
```
mxr576, line 131: Service should be DI-d

src/Plugin/KeyType/ApigeeAuthKeyType.php
```diff
@@ -130,12 +135,15 @@ public function validateKeyValue(array $form, FormStateInterface $form_state, $k
    */
   public function getAuthenticationMethod(KeyInterface $key): Authentication {
     $values = $key->getKeyValues();
-
     if ($this->getInstanceType($key) === EdgeKeyTypeInterface::INSTANCE_TYPE_HYBRID) {
-      $account_key = $this->getAccountKey($key);
-      return new HybridAuthentication($account_key['client_email'], $account_key['private_key'], \Drupal::service('apigee_edge.authentication.oauth_token_storage'));
+      if ($this->useGcpDefaultServiceAccount($key)) {
+        return new GceServiceAccountAuthentication(\Drupal::service('apigee_edge.authentication.oauth_token_storage'));
```
mxr576, line 140: Service should be DI-d.

src/Plugin/KeyType/ApigeeAuthKeyType.php
```diff
@@ -130,12 +135,15 @@ public function validateKeyValue(array $form, FormStateInterface $form_state, $k
...
+      }
+      else {
+        $account_key = $this->getAccountKey($key);
+        return new HybridAuthentication($account_key['client_email'], $account_key['private_key'], \Drupal::service('apigee_edge.authentication.oauth_token_storage'));
```
mxr576, line 144: Service should be Di-d.

mxr576: Idea: Create a HybridAuthenticationFactory for example to handle dependencies that should be constructor inject and those that vary per context, so they should be injected via method parameters.

arlina-espinoza:
CI failing because of incompatible dependencies. We need https://github.com/apigee/apigee-edge-drupal/pull/525, but that one in turns needs https://github.com/apigee/apigee-mock-client-php/pull/4.

PR #506, opened 2020-10-13
[#499] Use ManageTeamMembersAccess::access for team invitations

### LINKED ISSUE #499: Pending invitations block should be visible to a user with Drupal "Manage team members" permission (not only team permission)
## Description
Pending invitations block is displayed to users with team permission "Manage team members", but if a user is not a team admin but has "Manage team members" Drupal permissions - this block is not displayed.

Users with Drupal "Manage team members" are able to send invitation to users even if they are not a member of the team.

## Steps to Reproduce
1. Create a user who can manage Teams but not a Team admin (eg Dev portal manager)
2. Invite a user to a team

## Actual Behavior
Pending invitations block is not displayed.
Dev portal manager user can try to send invitation to same user again, but they'll get "this user already has invitation" error

## Expected Behavior
pending invitations should be displayed to users with **team permissions** "Team Admin" OR **Drupal** permissions "Manage team members" permissions 

**Users with Drupal "Manage team members" should be able to revoke/resend invitations as well**

## Note
I changed permissions to on that view to Drupal permission "Manage team members"
then Dev portal manger is able to view pending invitations, but this user cannot resend/revoke them.
If this user is able to create invitations, they should be able to revoke invitations.

## Screenshots
View's permissions:
image

## Version Info
apigee_edge 8.x-1.16

shadcn:
Fixes #499 

This PR removes the `team_manage_members` access config from the `Team invitations` view.

This view is embedded as part of `\Drupal\apigee_edge_teams\Controller\TeamMembersList::overview` and access for this is already handled in `\Drupal\apigee_edge_teams\Access\ManageTeamMembersAccess::access`.

https://github.com/apigee/apigee-edge-drupal/blob/c4ccce70d6b09b3d855446a06d83d1f83432d990/modules/apigee_edge_teams/src/Access/ManageTeamMembersAccess.php#L77-L101

Removing the access control on the view allows one source of truth for handling access for both members and team invitations.

Another possible fix would be a new view access plugin that does the same thing as `\Drupal\apigee_edge_teams\Access\ManageTeamMembersAccess::access`.

### DIFF
--- a/apigee_edge.install
+++ b/apigee_edge.install
@@ -27,6 +27,7 @@ use Drupal\apigee_edge\OauthTokenFileStorage;
 use Drupal\Core\Installer\InstallerKernel;
 use Drupal\Core\Url;
 use Drupal\user\RoleInterface;
+use Drupal\views\Views;
 
 /**
  * Implements hook_requirements().
@@ -274,3 +275,23 @@ function apigee_edge_update_8103() {
     user_role_grant_permissions(RoleInterface::AUTHENTICATED_ID, $authenticated_user_permissions);
   }
 }
+
+/**
+ * Remove the "Manage team members and invitations" access for the Team invitations view.
+ */
+function apigee_edge_update_8104() {
+  /** @var \Drupal\views\ViewExecutable $view */
+  $view = Views::getView('team_invitations');
+  $view->setDisplay('team');
+  $access = $view->getDisplay()->getOption('access');
+  if (empty($access['type']) || $access['type'] !== "team_permission") {
+    return;
+  }
+
+  $view->getDisplay()->setOption('access', [
+    'type' => 'none',
+    'options' => [],
+  ]);
+  $view->save();
+}
+
--- a/modules/apigee_edge_teams/config/optional/views.view.team_invitations.yml
+++ b/modules/apigee_edge_teams/config/optional/views.view.team_invitations.yml
@@ -721,9 +721,8 @@ display:
         empty: false
         access: false
       access:
-        type: team_permission
-        options:
-          permission: team_manage_members
+        type: none
+        options: {  }
     cache_metadata:
       max-age: -1
       contexts:
--- a/modules/apigee_edge_teams/src/Entity/TeamInvitation.php
+++ b/modules/apigee_edge_teams/src/Entity/TeamInvitation.php
@@ -60,7 +60,7 @@ use Drupal\user\UserInterface;
  *   },
  *   base_table = "team_invitation",
  *   data_table = "team_invitation_field_data",
- *   admin_permission = "administer team invitations",
+ *   admin_permission = "manage team members",
  *   entity_keys = {
  *     "id" = "uuid",
  *     "label" = "label",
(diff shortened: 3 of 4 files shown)


arlina-espinoza (changes requested):
Thanks @arshad, I noticed a couple of things while testing:
- The change to `views.view.team_invitations.yml` needs an update hook.
- If you grant a Drupal role the global "Manage team members" (`manage team members`) permission, they do get access to all the teams `/teams/{{team}}/members` page, where they can send new invitations, and see the team invitations block, but they cannot resend or revoke an existing invitation (no access to any invitation operations).
- I see in `TeamInvitationAccessControlHandler` that access to the "delete" or "resend" operations should be available to a user with the global permission `administer team invitations`, but I don't see that permission declared in the `TeamPermissionProvider`. It might have been carryover code from when we decided to unify the permission to manage team members and team invitations back in the original PR with this feature. In any case, access to these operations should be granted to a user with the global `manage team members` permission.
- The "admin permission on the `TeamInvitation` entity declaration should also be updated to be `manage team members`.

Thanks and let me know your comments/thoughts.

shadcn:
Thanks for the review @arlina-espinoza 

> If you grant a Drupal role the global "Manage team members" (manage team members) permission, they do get access to all the teams /teams/{{team}}/members page, where they can send new invitations, and see the team invitations block, but they cannot resend or revoke an existing invitation (no access to any invitation operations).

I noticed this bug too yesterday. I have a fix on the way.

> I see in TeamInvitationAccessControlHandler that access to the "delete" or "resend" operations should be available to a user with the global permission administer team invitations, but I don't see that permission declared in the TeamPermissionProvider. It might have been carryover code from when we decided to unify the permission to manage team members and team invitations back in the original PR with this feature. In any case, access to these operations should be granted to a user with the global manage team members permission.

I'll look into this and add a test for this.

shadcn:
~~@arlina-espinoza Since we are already using `\Drupal\apigee_edge_teams\Access\ManageTeamMembersAccess::access` for team invitations and we want to keep one permission, should we use same for the entity operations (resend/revoke) as well or do we want granular control here?~~

~~Note: member administration does NOT have granular control. `\Drupal\apigee_edge_teams\Access\ManageTeamMembersAccess::access` is used for everything.~~

~~I'm thinking we keep the granular permissions and add something like `ManageTeamMembersAccess:: access ` but for team invitation. So you'll have access if you either have "Delete/Remove own/any team invitation" or "Manage team members".~~

Update: Fixed using `TeamInvitationAccessControlHandler`.

shadcn:
@arlina-espinoza This is ready for another review.

1. I added an update hook to remove the `team_permission` access from the view.
2. Updated `\Drupal\apigee_edge_teams\Entity\TeamInvitationAccessControlHandler::checkAccess` to use `manage team members` or `administer team` for `revoke` and `resend` entity operations.
3. Switch the `admin permission` to `manage team members` for the `TeamInvitation` entity type.

arlina-espinoza:
@googlebot I consent.

arlina-espinoza (approved):
LGTM!

PR #507, opened 2020-10-13
[#492] Refactor app warnings

### LINKED ISSUE #492: Application API warnings are not present in apigee edge module
## Description

Warnings only present when profile is present making module is not fully functional.

Screen Shot 2020-09-29 at 6 30 15 pm

## Steps to Reproduce

1. Invalidate one of the APIs used by the application.
2. Warning only visible when installation profile is enable

## Proposal

Move the code from kickstart into the apigee edge module:

* https://github.com/apigee/apigee-devportal-kickstart-drupal/blob/f1672a9822c6294f02cd87a4974aaa322e582fec/modules/custom/apigee_kickstart_enhancement/apigee_kickstart_enhancement.module#L110-L139
* https://github.com/apigee/apigee-devportal-kickstart-drupal/blob/f1672a9822c6294f02cd87a4974aaa322e582fec/modules/custom/apigee_kickstart_enhancement/apigee_kickstart_enhancement.module#L258-L313

shadcn:
Fixes #492 

This PR

1. Adds a warning for appe with all credentials revoked.
2. Moves the warning checks from the AppListBuilder into a separate service: `\Drupal\apigee_edge\Entity\AppWarningsChecker` so that it can be re-used (example in Apigee Kickstart).
3. Adds a *Warnings* extra field to app entities for showing warnings in entity view displays.

### DIFF
--- a/src/Entity/AppWarningsChecker.php
+++ b/src/Entity/AppWarningsChecker.php
@@ -108,7 +108,7 @@ class AppWarningsChecker implements AppWarningsCheckerInterface {
 
     // If all credentials are revoked, show a warning.
     if (count($app->getCredentials()) === count($revoked_credentials)) {
-      $warnings['revokedCred'] = $this->t('All credentials associated with this @app are in revoked status.', $args);
+      $warnings['revokedCred'] = $this->t('No valid credentials associated with this @app.', $args);
     }
 
     return $warnings;
--- a/src/Entity/AppWarningsChecker.php
+++ b/src/Entity/AppWarningsChecker.php
@@ -1,0 +1,117 @@
+<?php
+
+/**
+ * Copyright 2020 Google Inc.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * version 2 as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+namespace Drupal\apigee_edge\Entity;
+
+use Apigee\Edge\Api\Management\Entity\AppCredentialInterface;
+use Apigee\Edge\Structure\CredentialProduct;
+use Drupal\Component\Datetime\TimeInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+
+/**
+ * A service for handling app warnings.
+ */
+class AppWarningsChecker implements AppWarningsCheckerInterface {
+
+  use StringTranslationTrait;
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * The time service.
+   *
+   * @var \Drupal\Component\Datetime\TimeInterface
+   */
+  protected $time;
+
+  /**
+   * AppWarningsChecker constructor.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
+   * @param \Drupal\Component\Datetime\TimeInterface $time
+   *   The time service.
+   */
+  public function __construct(EntityTypeManagerInterface $entity_type_manager, TimeInterface $time) {
+    $this->entityTypeManager = $entity_type_manager;
+    $this->time = $time;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getWarnings(AppInterface $app): array {
+    $warnings = [];
+    $warnings['revokedCred'] = FALSE;
+    $warnings['revokedOrPendingCredProduct'] = FALSE;
+    $warnings['expiredCred'] = FALSE;
+    $revoked_credentials = [];
+    $args = [
+      '@app' => mb_strtolower($app->getEntityType()->getSingularLabel()),
+    ];
+
+    foreach ($app->getCredentials() as $credential) {
+      // Check for revoked credentials.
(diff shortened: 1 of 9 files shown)


src/Entity/AppWarningsChecker.php
```diff
@@ -0,0 +1,117 @@
...
+   */
+  public function __construct(EntityTypeManagerInterface $entity_type_manager, TimeInterface $time) {
+    $this->entityTypeManager = $entity_type_manager;
+    $this->time = $time;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getWarnings(AppInterface $app): array {
```
mxr576, line 66: As this becomes decoupled finally, adding unit test coverage would be beneficial for how default warnings get calculated.

src/Entity/ListBuilder/AppListBuilder.php
```diff
@@ -84,19 +92,25 @@ class AppListBuilder extends EdgeEntityListBuilder {
...
+   * @param \Drupal\apigee_edge\Entity\AppWarningsCheckerInterface $app_warnings_checker
+   *   The app warnings checker service.
    */
-  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager, RendererInterface $renderer, RequestStack $request_stack, TimeInterface $time, ConfigFactoryInterface $config_factory = NULL) {
+  public function __construct(EntityTypeInterface $entity_type, EntityTypeManagerInterface $entity_type_manager, RendererInterface $renderer, RequestStack $request_stack, TimeInterface $time, ConfigFactoryInterface $config_factory = NULL, AppWarningsCheckerInterface $app_warnings_checker = NULL) {
     if (!$config_factory) {
       $config_factory = \Drupal::service('config.factory');
     }
+    if (!$app_warnings_checker) {
+      $app_warnings_checker = \Drupal::service('apigee_edge.entity.app_warnings_checker');
```
mxr576, line 103: An error should be also triggered if a required parameter was missing and a fallback solution had to be applied
See: https://github.com/drupal/core/blob/8.8.x/lib/Drupal/Core/Entity/ContentEntityStorageBase.php#L81

shadcn:
I'm still working on this (see draft status). Not ready for review. Thanks @mxr576

src/Entity/AppWarningsChecker.php
shadcn: Yeah I'm going to refactor the AppListBuilderTest (Kernel) to cover this new class.

src/Entity/AppWarningsCheckerInterface.php
```diff
@@ -0,0 +1,40 @@
...
+   * Checks credentials of an app and returns warnings about them.
+   *
+   * @param \Drupal\apigee_edge\Entity\AppInterface $app
+   *   The app entity to be checked.
+   *
+   * @return array
+   *   An associative array that contains information about the revoked
+   *   credentials and revoked or pending API products in a credential.
+   */
+  public function getWarnings(AppInterface $app): array;
```
mxr576, line 38: Probably it would be better calling it `getMessages()` and allowing downstream projects to decide the severity of the message that is going to be displayed.

src/Entity/AppWarningsChecker.php
mxr576: Well, I would rather suggest a dedicated unit test for this small unit instead of covering it one location where it is used. Having this as a service makes it possible to render app "warnings" basically anywhere where an app is available in the context.

src/Entity/AppWarningsCheckerInterface.php
shadcn: Makes sense. TBH I'm not completely sold on the `AppWarningsChecker` name either.

src/Entity/AppWarningsChecker.php
shadcn: Yeah this is what I meant. We're going to re-use this service outside of AppListBuilder.

mxr576:
> I'm still working on this (see draft status). Not ready for review. Thanks @mxr576

Yeah I know, I am just wanted to share my ideas to help to design this.

src/Entity/AppWarningsChecker.php
```diff
@@ -0,0 +1,117 @@
...
+
+      // Check for expired credentials.
+      if (($expired_date = $credential->getExpiresAt()) && $this->time->getRequestTime() - $expired_date->getTimestamp() > 0) {
+        $warnings['expiredCred'] = $this->t('At least one of the credentials associated with this @app is expired.', $args);
+      }
+
+      // Check status of API products for credential.
+      foreach ($credential->getApiProducts() as $cred_product) {
+        if ($cred_product->getStatus() == CredentialProduct::STATUS_REVOKED || $cred_product->getStatus() == CredentialProduct::STATUS_PENDING) {
+          $args['@api_product'] = mb_strtolower($this->entityTypeManager->getDefinition('api_product')
```
arlina-espinoza: The default singular label is "API", and I think it would be a common name. Lower-casing it means the warning looks unprofessional ("_Bla Bla_ api associated with this app is in pending status."), should we remove `mb_strtolower()`?

arlina-espinoza (changes requested):

arlina-espinoza:
@googlebot I consent.

arlina-espinoza:
@arshad I made some updates to the `AppWarningsCheckerTest` and `AppListBuilderTest` to make sure they also pass when integration is enabled during tests (ie, real API connection).

src/Entity/AppWarningsChecker.php
shadcn: Yeah. We don't really need `mb_strtolower` in most cases anyway. `->getSingularLabel()` should already return the expected format.

```php
/**
   * Gets the indefinite singular form of the name of the entity type.
   *
   * This should return the human-readable name for a single instance of
   * the entity type. For example: "opportunity" (with the plural as
   * "opportunities"), "child" (with the plural as "children"), or "content
   * item" (with the plural as "content items").
   *
   * Think of it as an "in a full sentence, this is what we call this" label. As
   * a consequence, the English version is lowercase.
   *
   * @return string|\Drupal\Core\StringTranslation\TranslatableMarkup
   *   The singular label.
   */
  public function getSingularLabel();
```

CR: https://www.drupal.org/node/3075567

shadcn: ~~On the other hand, can we assume it will be API in most cases (since this can be renamed via the UI)?~~

It looks like yes API is common. So I went ahead and removed the `mb_strtolower`. Thanks @arlina-espinoza

arlina-espinoza (approved):
LGTM :+1:

mxr576:
Raised a new issue as a follow up of a conversion in this PR ^

PR #509, opened 2020-10-19
[#483] Normalize module paths.

### LINKED ISSUE #483: Application path is mixed with Drupal path breaking breadcrumbs and active menus.
## Description
Application path is mixed with Drupal path breaking breadcrumbs and active menus. Examples:

* `/user/1/create-app`
* `/user/1/apps/adminapp/edit`
* `/user/1/apps/adminapp`
* `/user/1/apps`
* `/add-team`

## Expected Behavior
Proposition is to do align with other Drupal entities and redefine paths as:

* `/app/add`
* `/app/adminapp/edit`
* `/app/adminapp`
* `/apps`
* `/team/add`

zakiya:
Addresses #483 

Release notes:
- Will require `drush cr` after update. 
- The following paths have changed:

Old path | New path | Route
-- | -- | --
/add-team | /teams/add | entity.team.add_form
/teams/{team}/add-members | /teams/{team}/members/add | entity.team.add_members
/team-apps | /teams/apps | entity.team_app.collection
/team-apps/add | /teams/apps/add | entity.team_app.add_form
/teams/{team}/create-app | /teams/{team}/apps/add | entity.team_app.add_form_for_team
/user/{user}/create-app | /user/{user}/apps/add | entity.developer_app.add_form_for_developer


arlina-espinoza:
@cnovak Wondering if we also need to add a redirect to avoid breaking users that have hardcoded these paths.

cnovak:
> @cnovak Wondering if we also need to add a redirect to avoid breaking users that have hardcoded these paths.

I can see that being an issue.

arlina-espinoza:
@zakiya Let's add a redirect from the old paths to the new paths to avoid breaking existing sites.

modules/apigee_edge_teams/apigee_edge_teams.routing.yml
```diff
@@ -71,3 +71,51 @@ apigee_edge_teams.settings.team_app.cache:
...
+    parameters:
+      entity_slug:
+        type: entity:team
+  requirements:
+    _permission: 'access content'
+
+apigee_edge_teams.redirect.team_app.add_form_for_team:
+  path: '/teams/{entity_slug}/create-app'
+  defaults:
+    _controller: '\Drupal\apigee_edge\Controller\redirectController::manageRedirect'
```
arlina-espinoza: I think the issue the tests are failing is because the redirect controller should be uppercase on the class name.

zakiya: That was it. Thank you.

apigee_edge.routing.yml
```diff
@@ -190,3 +190,15 @@ apigee_edge.export_analytics.csv:
...
+  path: '/user/{entity_slug}/create-app'
+  defaults:
+    _controller: '\Drupal\apigee_edge\Controller\RedirectController::manageRedirect'
+    redirect_route: 'entity.developer_app.add_form_for_developer'
+  options:
+    parameters:
+      entity_slug:
+        type: entity:user
+  requirements:
+    _permission: 'access content'
```
arlina-espinoza: I just noticed that if users have hardcoded any links in menus (using tokens or any other method) then the legacy links will not respect the original route access permissions because of this "access content" requirement (it will lead them to perhaps seeing a link that ultimately will give a 403 error if they don't have the necessary permissions).
I've added a suggestion of how we could rely on the original access controller instead:

```suggestion
    entity_slug: '\d+'
    _custom_access: '\Drupal\apigee_edge\Controller\RedirectController::createDeveloperAppAccess'
```
 Note: suggestion continues on the `RedirectController` class, and I've only added it for the `/user/{entity_slug}/create-app` route, it would need to be implemented on the other legacy routes as well, if you think this is a good approach.

src/Controller/RedirectController.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+namespace Drupal\apigee_edge\Controller;
+
+use Drupal\Core\Entity\EntityBase;
+use Drupal\Core\Controller\ControllerBase;
```
arlina-espinoza, line 7:
```suggestion
use Drupal\Core\Access\AccessResult;
use Drupal\Core\Entity\EntityBase;
use Drupal\Core\Controller\ControllerBase;
use Drupal\Core\Routing\RouteMatchInterface;
use Drupal\Core\Session\AccountInterface;
use Drupal\Core\Url;
```

src/Controller/RedirectController.php
```diff
@@ -0,0 +1,83 @@
...
+      $redirect = $this->redirect($redirect_route, [$this->entity_type_id => $this->entity_id]);
+    }
+
+    else {
+      $redirect = $this->redirect($redirect_route);
+    }
+
+    return $redirect;
+
+  }
```
arlina-espinoza, line 85:
```suggestion
  }

  /**
   * Grant access to "apigee_edge.redirect.developer_app.add_form_for_developer".
   *
   * @param mixed $entity_slug
   *   Entity slug from input.
   * @param string $redirect_route
   *   Route parameter defined as defined in routing.yml.
   * @param \Drupal\Core\Routing\RouteMatchInterface $route_match
   *   The route match.
   * @param \Drupal\Core\Session\AccountInterface $account
   *   The currently logged in account.
   *
   * @return \Drupal\Core\Access\AccessResultInterface
   *   The access result.
   */
  public function createDeveloperAppAccess($entity_slug, string $redirect_route, RouteMatchInterface $route_match, AccountInterface $account) {
    if ($entity_slug != NULL) {
      $this->setEntityId($entity_slug);
      $url = Url::fromRoute($redirect_route, ['user' => $this->entity_id]);
      return $url->access($account) ? AccessResult::allowed() : AccessResult::forbidden();
    }

    return AccessResult::forbidden('User is missing from route.');
  }
```

arlina-espinoza (changes requested):
@zakiya This looks good, just I added a final comment about the legacy routes permissions.

src/Controller/RedirectController.php
zakiya: I added another parameter because I thought I might need it for the teams redirects but I think those all can be handled by existing permissions. Let me know if you think I should change it back.

arlina-espinoza: I like the idea of making this more generic with the `$access_type`, and I think it will be needed, because the issue with having the legacy routes just check for permissions is that it ignores the team context and team permissions (they are different from drupal permissions). 
For example, `apigee_edge_teams.redirect.team.add_members` should be accessible to a user with the drupal permission `manage team members` or if they belong to the given team and have the team permission of `team_manage_members`. All of that logic is already taken care of in the access check of the redirect_route, so I think it would be easier to just construct the Url object of the redirect route and check access on it. LMK if it makes sense.

zakiya: Ah! That makes sense. Changes on the way.

arlina-espinoza (approved):
LGTM, thanks @zakiya :)
I merged with upstream/8.x-1.x so tests will pass, and also renamed the access control check method to a more generic name.

arlina-espinoza:
@googlebot I consent.

PR #513, opened 2020-10-28
[#480] Add warnings to form and status page if using basic auth.

### LINKED ISSUE #480: Deprecate/Remove Basic Auth from Edge Connection Settings
[Basic Auth is going to be deprecated](https://docs.apigee.com/api-platform/system-administration/basic-auth#warning). We should:

- Remove the ability to configure Basic Auth on the Apigee connection settings, or for a first release display a warning to the user on the connection settings page and in the status report that Basic Auth is being deprecated and that they should use OAuth instead.
- Migrate anyone using Basic Auth to use OAuth. We can do this because there is a hardcoded OAuth [Client ID  and secret ](https://github.com/apigee/apigee-client-php/blob/52e4713d8ade32ba09ed3606f5df8ea314c03ba7/src/HttpClient/Plugin/Authentication/Oauth.php#L37) for all orgs that are not SAML enabled, and SAML enabled orgs cannot use Basic Auth. 
- When migrating a site to use OAuth, if the OAuth connection does not work, we should warn the user, add info to log, and keep the site on Basic Auth.

arlina-espinoza:
Closes #480 . As a note, we discussed that we could not safely migrate everyone to using oauth because they might be using a custom authorization server, so for now we will only show warning messages.
 
This PR:
- Adds a warning message to the connection settings form if using basic auth.
- Marks the "basic auth" option on the dropdown as deprecated.
- Adds a warning message to Drupal's status report page if using basic auth.

Screenshots:

Connection settings form:
image

Status report page:
image

### DIFF
--- a/apigee_edge.install
+++ b/apigee_edge.install
@@ -84,15 +84,20 @@ function apigee_edge_requirements($phase) {
       }
 
       // Warning message in status report if using basic auth.
-      if ($key && $key->getKeyType() instanceof EdgeKeyTypeInterface &&
-        $key->getKeyType()->getAuthenticationType($key) === EdgeKeyTypeInterface::EDGE_AUTH_TYPE_BASIC) {
-        $requirements['apigee_edge_http_basic_auth'] = [
-          'title' => t('Apigee Edge'),
-          'description' => t('Apigee Edge HTTP basic authentication will be deprecated. Please choose another authentication method. Visit the <a href=":url">Apigee Edge general settings</a> page to get more information.', [
-            ':url' => Url::fromRoute('apigee_edge.settings', ['destination' => 'admin/reports/status'])->toString(),
-          ]),
-          'severity' => REQUIREMENT_WARNING,
-        ];
+      try {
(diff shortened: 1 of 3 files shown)


apigee_edge.install
```diff
@@ -68,17 +69,36 @@ function apigee_edge_requirements($phase) {
...
+            'title' => t('Apigee Edge'),
+            'description' => t('Apigee Edge HTTP basic authentication will be deprecated. Please choose another authentication method. Visit the <a href=":url">Apigee Edge general settings</a> page to get more information.', [
+              ':url' => Url::fromRoute('apigee_edge.settings', ['destination' => 'admin/reports/status'])->toString(),
+            ]),
+            'severity' => REQUIREMENT_WARNING,
+          ];
+        }
+      }
+      catch (Exception $e) {
+        // Do nothing.
```
cnovak, line 100: Should this log an error to watchdog if exception happens?

arlina-espinoza: An exception is thrown when the connection has not been configured, and we already log a warning in the status detail page, so there is no need to further capture that here.

cnovak (approved):
LGTM

PR #514, opened 2020-10-29
[#512] Masks consumerKey and consumerSecret in all debug logs

### LINKED ISSUE #512: Apigee Edge Debug: Hide key/secret in the response.
When user updates their app and apigee_edge_debug is enabled, user's app's consumerKey and consumerSecret is visible in plain text in response body (message of the log message).

Parse and replace consumerKey/consumerSecret with *** as that might lead to security issues.

shadcn:
Fixes #512


arlina-espinoza (approved):
LGTM, thanks @arshad .

PR #518, opened 2020-11-10
[#508] Disable team actions if team is inactive

### LINKED ISSUE #508: able to modify app inside inactive company
Hi, 

We are using Apigee OPDK + Drupal 8.x + Monetization (Companies\Teams).

We hit a situation, which is super strange to us, while the company is inactive status, the company owner able to do anything inside the company apps (add apps, modify, delete etc.) - our understanding is that if company is inactive, all the other things are kind of open from UI\UX POV. 

We are aware that they cannot use the apps, and this is fine, but from UI\UX customer would like to also make the inside of company read-only. 

Is this a bug or feature, little explanation?

Thanks,

Denis

shadcn:
Fixes #508 

Since Drupal does not render links with `\Drupal\Core\Access\AccessResultForbidden`, we can't alter access here.

This PR implements an `event_subscriber` to alter the response when a team is inactive. An error message is displayed.

As for the UI part, we are replacing local actions and tasks with url `<none>` and showing a disabled state using CSS.

### DIFF
--- a/modules/apigee_edge_teams/src/Controller/TeamMembersList.php
+++ b/modules/apigee_edge_teams/src/Controller/TeamMembersList.php
@@ -91,7 +91,8 @@ class TeamMembersList extends ControllerBase {
   public static function create(ContainerInterface $container) {
     return new static(
       $container->get('apigee_edge_teams.team_membership_manager'),
-      $container->get('entity_type.manager')
+      $container->get('entity_type.manager'),
+      $container->get('module_handler')
     );
   }
 
--- a/modules/apigee_edge_teams/src/Controller/TeamMembersList.php
+++ b/modules/apigee_edge_teams/src/Controller/TeamMembersList.php
@@ -71,7 +71,7 @@ class TeamMembersList extends ControllerBase {
    */
   public function __construct(TeamMembershipManagerInterface $team_membership_manager, EntityTypeManagerInterface $entity_type_manager, ModuleHandlerInterface $module_handler = NULL) {
     if (!$module_handler) {
-      @trigger_error('Calling ' . __METHOD__ . ' without the $module_handler is deprecated in apigee_edge:8-x-1.19 and is required before apigee_edge:8.x-2.0.', E_USER_DEPRECATED);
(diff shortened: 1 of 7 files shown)


modules/apigee_edge_teams/src/Controller/TeamMembersList.php
```diff
@@ -51,21 +52,37 @@ class TeamMembersList extends ControllerBase {
...
+   * @param \Drupal\Core\Extension\ModuleHandlerInterface|null $module_handler
+   *   The module handler.
    */
-  public function __construct(TeamMembershipManagerInterface $team_membership_manager, EntityTypeManagerInterface $entity_type_manager) {
+  public function __construct(TeamMembershipManagerInterface $team_membership_manager, EntityTypeManagerInterface $entity_type_manager, ModuleHandlerInterface $module_handler = NULL) {
+    if (!$module_handler) {
+      @trigger_error('Calling ' . __METHOD__ . ' without the $module_handler is deprecated in apigee_edge:8-x-1.19 and is required before apigee_edge:8.x-2.0. See https://github.com/apigee/apigee-edge-drupal/pull/518.', E_USER_DEPRECATED);
+      $module_handler = \Drupal::moduleHandler();
+    }
+
```
arlina-espinoza, line 77: Let's add the new `$module_handler` parameter to the `create()` function too.

arlina-espinoza (changes requested):
This looks good to me, I'll just check with the Apigee team before approving. I also added a comment requiring a small change.

modules/apigee_edge_teams/src/Controller/TeamMembersList.php
arlina-espinoza: @arshad Talked this over with the Apigee team and it looks good. Lets just get that `$module_handler` param in the `create()` function and mark it as ready for review.

shadcn: Done. How did I miss this! 🤪

arlina-espinoza (approved):
LGTM, thanks @arshad !

PR #519, opened 2020-11-14
Move teams related update function to apigee_edge_teams

arlina-espinoza:
The `apigee_edge` `function apigee_edge_update_8105()` update function fails if the `apigee_edge_teams` is not enabled. This PR moves that update function to the `apigee_edge_teams` module.


shadcn (approved):
LGTM. Thanks @arlina-espinoza

PR #522, opened 2020-11-20
[#516-517] Fix team invitations for multiple and deleted teams

### LINKED ISSUE #516: Team member invitation limit
## Description
Once a member is invited to a team, until he accepts or rejects the invitation, he can't be added to a second team

## Steps to Reproduce
1. Create a team
2. Add a member to the team. (Do not accept/reject the invitation yet)
3. Create a second team
4. Add the same member to the second team

## Actual Behavior
This throws an error saying "The following developer has already been invited to the team: 'developer_email@abc.com'."

## Expected Behavior
A member should be allowed to receive multiple invitations, irrespective of he accepts/rejects the invitation.

## Version Info
1.18

### LINKED ISSUE #517: Unable to invite members to another team, if teams get deleted with pending member invitation
## Description
Unable to invite members to another team, if teams get deleted with pending member invitation

## Steps to Reproduce
1. Create a team
2. Add a member to the team. (Do not accept/reject the invitation yet)
3. Delete the team.
3. Create a second team
4. Add the same member to the second team

## Actual Behavior
This throws an error saying "The following developer has already been invited to the team: 'developer_email@abc.com'."

## Expected Behavior
All invitation of the team(pending), should be deleted once a team is deleted, and a member should be allowed to add in second team.

## Version Info
1.18

shadcn:
Fixes #516 and #517 

This PR:

1. Adds the team id when validation existing team_invitations for user.
2. Implements a query access subscriber to add a condition to check for existing teams.

### DIFF
--- a/modules/apigee_edge_teams/config/optional/views.view.team_invitations.yml
+++ b/modules/apigee_edge_teams/config/optional/views.view.team_invitations.yml
@@ -1150,7 +1150,7 @@ display:
       query:
         type: views_query
         options:
-          disable_sql_rewrite: true
+          disable_sql_rewrite: false
           distinct: false
           replica: false
           query_comment: ''
--- a/modules/apigee_edge_teams/src/Entity/TeamInvitation.php
+++ b/modules/apigee_edge_teams/src/Entity/TeamInvitation.php
@@ -289,4 +289,14 @@ class TeamInvitation extends ContentEntityBase implements TeamInvitationInterfac
     }
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTags() {
+    if ($this->getTeam()) {
+      $this->cacheTags = array_merge($this->cacheTags, $this->getTeam()->getCacheTags());
+    }
+    return parent::getCacheTags();
+  }
+
 }
--- a/modules/apigee_edge_teams/src/EventSubscriber/TeamInvitationQueryAccessSubscriber.php
+++ b/modules/apigee_edge_teams/src/EventSubscriber/TeamInvitationQueryAccessSubscriber.php
@@ -1,0 +1,72 @@
+<?php
+
+/**
+ * Copyright 2020 Google Inc.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * version 2 as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+namespace Drupal\apigee_edge_teams\EventSubscriber;
+
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\entity\QueryAccess\QueryAccessEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Subscribes to query access events for team_invitation.
+ */
+class TeamInvitationQueryAccessSubscriber implements EventSubscriberInterface {
+
+  /**
+   * The entity type manager service.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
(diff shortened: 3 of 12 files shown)


modules/apigee_edge_teams/config/optional/views.view.team_invitations.yml
```diff
@@ -1150,7 +1150,7 @@ display:
       query:
         type: views_query
         options:
-          disable_sql_rewrite: true
+          disable_sql_rewrite: false
```
arlina-espinoza, line 1153: This needs to be added as well to a module update hook.

modules/apigee_edge_teams/src/Entity/TeamInvitation.php
```diff
@@ -289,4 +289,14 @@ public function preSave(EntityStorageInterface $storage) {
...
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTags() {
+    if ($this->getTeam()) {
+      $this->cacheTags = array_merge($this->cacheTags, $this->getTeam()->getCacheTags());
+    }
+    return parent::getCacheTags();
+  }
```
arlina-espinoza, line 300: Should we also add the user account cache tags?

modules/apigee_edge_teams/src/EventSubscriber/TeamInvitationQueryAccessSubscriber.php
```diff
@@ -0,0 +1,72 @@
...
+   * Modifies the access conditions for team_invitation.
+   *
+   * @param \Drupal\entity\QueryAccess\QueryAccessEvent $event
+   *   The event.
+   */
+  public function onQueryAccess(QueryAccessEvent $event) {
+    // Add a condition to check for a valid team.
+    // We query team from storage instead of check for a null team field because
+    // the team might have been deleted on the remote server.
+    $team_ids = array_keys($this->entityTypeManager->getStorage('team')->loadMultiple());
```
arlina-espinoza, line 68: During my testing, I had a pending invitation a user on a team, and deleted the team (from another browser session). The user's "team invitations" tab is showing a white-screen error, and I don't see this function firing at any point.

arlina-espinoza (changes requested):
Thanks @arshad ! Left some comments. I found that the PR does fix #516. :+1: 
Also, about the team deletion issue:
-  I think that if the team deletion does happen from the Drupal side, we should immediately delete all invitations for the team. 
- `TeamInvitationAccessControlHandler::checkAccess()` should return an access denied if the team cannot be loaded, before any other checks (that would solve the white-screen issue I experienced).
- A cron job would still be useful for periodic purging of Apigee-initiated team deletions, and at least the `checkAccess()` above would prevent issues if the invitation has not been deleted yet.

modules/apigee_edge_teams/config/optional/views.view.team_invitations.yml
shadcn: Done. Added `apigee_edge_teams_update_8705`.

modules/apigee_edge_teams/src/Entity/TeamInvitation.php
shadcn: The user account is actually not tied to a `team_invitation`. We're using a `recipient` string field.

modules/apigee_edge_teams/src/EventSubscriber/TeamInvitationQueryAccessSubscriber.php
shadcn: I think this is probably because of the missing update hook. Can you test again please? This should set `disable_sql_rewrite` to false and the query access should fire.

arlina-espinoza: Tried it and still getting the whitepage error. I also tried it on a clean install and same result. Still think it would be useful to check if the team can be loaded in `TeamInvitationAccessControlHandler::checkAccess() `.

shadcn: I actually did this before the query access :)

I'll add it back.

shadcn:
> I think that if the team deletion does happen from the Drupal side, we should immediately delete all invitations for the team.

> A cron job would still be useful for periodic purging of Apigee-initiated team deletions, and at least the checkAccess() above would prevent issues if the invitation has not been deleted yet.

@arlina-espinoza Since this issue is limiting team invitations, can we merge this PR and handle purging team invitations in a separate ticket?

arlina-espinoza:
@googlebot I consent.

arlina-espinoza (approved):
@arshad I agree we can handle the purging on cron of remotely deleted teams in a follow up ticket.
I did add to this PR a hook to delete related invitations when a team gets deleted, and also a testing class.

PR #525, opened 2020-11-29
Updates php-http/guzzle6-adapter to 2.x

arlina-espinoza:
Supports https://github.com/apigee/apigee-client-php/pull/106, to drop support of HTTPlug 1.x. Will need to re-run tests once a new version of `apigee/apigee-client-php` has been released, until then, CI will fail.


mxr576 (approved):

zakiya (approved):
Looks good!

PR #528, opened 2020-12-04
Revert "[#483] Normalize module paths. (#509)"

### LINKED ISSUE #527: [urgent] Rollback "Normalize module paths" #509
## Description
PR #509 was merged and it possibly introduces major regressions in handling "special" entity names. 

> Old path	New path	Route
> /add-team	/teams/add	entity.team.add_form
> /teams/{team}/add-members	/teams/{team}/members/add	entity.team.add_members
> /team-apps	/teams/apps	entity.team_app.collection
> /team-apps/add	/teams/apps/add	entity.team_app.add_form
> /teams/{team}/create-app	/teams/{team}/apps/add	entity.team_app.add_form_for_team
> /user/{user}/create-app	/user/{user}/apps/add	entity.developer_app.add_form_for_developer

The reason behind those olds paths was to be able to handle special entity names.
Like the path of the add team form was `add-team` instead of `/teams/add` because the canonical URL of a team entity is also `teams/[id]` and if the id of an entity is `add` then there is a collision between paths.
The same principle applies to all paths since the id of the entity is usually a machine name of Apigee entity which depends on a user's input. If it would be a UUID then this issue would not exist, but for the sake of pretty URLs (and following the design from Drupal 7), it was decided to use the machine name of the entity instead.

Also changing entity paths without providing a BC layer could be an issue and a concern was raised about that in the original PR but it has not been addressed before the PR was merged. https://github.com/apigee/apigee-edge-drupal/pull/509#issuecomment-720712827

## Steps to Reproduce
1. Go to Apigee Edge MGMT UI and create a team which machine is "add" or use the API for that.

image

2. Go to a developer portal that is connected to the same Apigee org.
3. Open the `/team/add` path. 

Peek 2020-12-04 10-08

## Notes

The requested behavior can be achieved by using path aliases and changing entity ids, BUT first, let's rollback this PR before it causes problems on downstream projects.

mxr576:
This closes #527


arlina-espinoza (approved):
Good catch, I'll reopen original ticket and PR to discuss further.

PR #530, opened 2020-12-09
[#524] Ensure team view builder uses cache_expiration config

### LINKED ISSUE #524: caching on team level
Hi, we have found a bug (as we think): 

To reproduce: 

1. Create company in Apigee
image

2. Observe the status of the company in the list in the portal: 

image

image

Now, change the status in Apigee to Inactive:
image

Observe the portal: 

image

image

As we can see all works as expected, however, this works only for Global Portal admin, 

For team member\team admin we can see that change only reflected in the list of companies: 

image

While in the company we can see status remains 

image

(flushing routing and links cache) solves this

But we want the behavior be same between admin and enduser

Thanks,

Denis

shadcn:
Fixes #524 

This adds the `apigee_edge_teams.team_settings.cache_expiration` config to the `Team` max-age.

We implemented this for `TeamListBuilder` but not for `ViewBuilder`.


arlina-espinoza (approved):
Good catch :+1:

PR #531, opened 2020-12-09
[#472] Make analytics environments filter configurable

### LINKED ISSUE #472: Support environment selection on App analytics page
**Is your feature request related to a problem? Please describe.**

Currently, the user specific App analytics page (**/user/[UID]/apps/[APP ID]/analytics**) fetches the analytics data from one environment which can be changed only by admins through the **/admin/config/apigee-edge/app-settings/general/analytics** page.

It would be nice if end users would be also able to switch among environments on the App analytics page. Otherwise, they only see analytics data for one pre-defined environment.

**Describe the solution you'd like**

A new filter option could be added to the App analytics page which would list the available environments.

**Describe alternatives you've considered**

Perhaps the analytics related admin page (**/admin/config/apigee-edge/app-settings/general/analytics**) could be also modified or extended with additional settings to restrict or control which environments can be selected on the App analytics page.

shadcn:
Fixes #472

### DIFF
--- a/apigee_edge.install
+++ b/apigee_edge.install
@@ -324,3 +324,15 @@ function apigee_edge_update_8104() {
  */
 function apigee_edge_update_8105() {
 }
+
+/**
+ * Update the default environment config for app analytics.
+ */
+function apigee_edge_update_8106() {
+  $config_factory = \Drupal::configFactory();
+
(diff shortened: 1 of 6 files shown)


arlina-espinoza (approved):
LGTM, thanks @arshad :100:

PR #539, opened 2020-12-17
Added the ability to autodetect Apigee organization for Hybrid

giteshk:
For hybrid we can figure out the organization that the current gcp project is connected to.

This is to improve the install experience on GCP

### DIFF
--- a/src/Connector/GceServiceAccountAuthentication.php
+++ b/src/Connector/GceServiceAccountAuthentication.php
@@ -20,9 +20,9 @@
 
 namespace Drupal\apigee_edge\Connector;
 
+use Apigee\Edge\Api\Management\Controller\OrganizationController;
 use Apigee\Edge\ClientInterface;
 use Apigee\Edge\HttpClient\Plugin\Authentication\GceServiceAccount;
-use Drupal\apigee_edge\Entity\Controller\OrganizationController;
 
 /**
  * Decorator for Hybrid authentication plugin.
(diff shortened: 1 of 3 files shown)


src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * Copyright 2018 Google Inc.
+ * Copyright 2020 Google Inc.
```
mxr576, line 4: As far as I remember, the data here is when the file was created and not when the file was last modified. So it should not change #fixme.

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -134,14 +134,24 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
...
         ':input[name="key_input_settings[gcp_hosted]"]' => ['checked' => FALSE],
       ];
+      if (empty($form['organization']['#default_value'])) {
+        $organizations = $gceServiceAccountAuth->getAuthorizedOrganizations();
+        $organization = array_pop($organizations);
+        if ($organization) {
+          $form['gcp_hosted']['#default_value'] = TRUE;
+          $form['instance_type']['#default_value'] = EdgeKeyTypeInterface::INSTANCE_TYPE_HYBRID;
+          $form['organization']['#default_value'] = $organization->getName();
+          \Drupal::messenger()->addWarning("Apigee configuration has been auto detected. Please save this configuration. \n You may change this configuration to connect to a different Apigee organization.");
```
mxr576, line 152: Use HTML tags instead of `\n` to break text,

src/Plugin/KeyInput/ApigeeAuthKeyInput.php
```diff
@@ -134,14 +134,24 @@ public function buildConfigurationForm(array $form, FormStateInterface $form_sta
...
         '#states' => [
           'visible' => $state_for_hybrid,
         ],
       ];
       $state_for_not_gcp_hosted = [
         ':input[name="key_input_settings[gcp_hosted]"]' => ['checked' => FALSE],
       ];
+      if (empty($form['organization']['#default_value'])) {
+        $organizations = $gceServiceAccountAuth->getAuthorizedOrganizations();
+        $organization = array_pop($organizations);
```
mxr576, line 147: What about the other orgs? What about the other available? (The method name indicates to me that there could be more). Why not just `$organization = reset($organizations);`?

src/Connector/GceServiceAccountAuthentication.php
```diff
@@ -37,4 +38,18 @@ protected function authClient(): ClientInterface {
...
 
+  /**
+   * Returns the list of all Organizations available to the service account.
+   *
+   * @return \Apigee\Edge\Api\Management\Entity\OrganizationInterface[]
+   *   Returns the list of organizations the service account can access.
+   */
+  public function getAuthorizedOrganizations(): array {
+    /** @var \Drupal\apigee_edge\SDKConnectorInterface $sdk_connector */
+    $sdk_connector = \Drupal::service('apigee_edge.sdk_connector');
```
mxr576, line 49: Still, using dependency injection here would be useful, not just because the code could unit tested :)

mxr576 (changes requested):

arlina-espinoza (commented):
Nice feature to add :) 
I agree with @mxr576 comments. Additionally, if this PR requires a fix on the PHP client (https://github.com/apigee/apigee-client-php/pull/114), it should also update the minimum required version in this module's composer file.

raakesh-blokhra:
@giteshk There is no issue associated with this pull request, and also the checks have failed. How do you plan to proceed with this? Do we need to close this?

PR #544, opened 2020-12-29
[#343] Hides portal environment in app analytics config

### LINKED ISSUE #343: Portal environment should be hidden in app analytics config
The "portal" environment is for internal use with integrated portals and is not an actual environment customers should use. It should be hidden from configuration to reduce confusion:

portal-environment

divya-intelli:
The portal environment is for internal use with integrated portals and is not an actual environment for customer's use. To reduce confusion portal environment is hidden from the app analytics configuration.
This fixes #343.

### DIFF
--- a/src/Form/AppAnalyticsSettingsForm.php
+++ b/src/Form/AppAnalyticsSettingsForm.php
@@ -84,7 +84,9 @@ class AppAnalyticsSettingsForm extends ConfigFormBase {
     // Portal environment is for internal use with integrated portals and
     // is not an actual environment for customers use.
     // To reduce confusion portal environment is hidden from configuration.
-    $environments = array_diff($this->environmentController->getEntityIds(), ['portal']);
+    $environments = $this->environmentController->getEntityIds();
+    $environments = array_combine($environments, $environments);
+    unset($environments['portal']);
 
     $form['label'] = [
       '#type' => 'fieldset',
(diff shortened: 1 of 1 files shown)


src/Form/AppAnalyticsSettingsForm.php
```diff
@@ -81,7 +81,8 @@ public function getFormId() {
    * {@inheritdoc}
    */
   public function buildForm(array $form, FormStateInterface $form_state) {
-    $environments = $this->environmentController->getEntityIds();
+    // Remove portal environment from environment list.
+    $environments = array_diff($this->environmentController->getEntityIds(), ['portal']);
```
mxr576: Since  `'array_combine($environments, $environments),` is being called below (multiple times 🙄 ), it is cheaper if you do the `array_combine()` here and then call `unset($environments['portal']);`.

Also please leave a more detailed description about WHY the portal environment is removed (something like we can see in the original issue.)

mxr576 (changes requested):

raakesh-blokhra (commented):
Looks good.

raakesh-blokhra (approved):
Looks good to me.

mxr576 (changes requested):
I still do not see my previous suggestion resolved which could simplify this new feature and in general, the logic in this code:

> Since 'array_combine($environments, $environments), is being called below (multiple times 🙄 ), it is cheaper if you do the array_combine() here and then call unset($environments['portal']);.

raakesh-blokhra:
> I still do not see my previous suggestion resolved which could simplify this new feature and in general, the logic in this code:
> 
> > Since 'array_combine($environments, $environments), is being called below (multiple times 🙄 ), it is cheaper if you do the array_combine() here and then call unset($environments['portal']);.

Thanks for the suggestion @mxr576. I remember this very well. Let me get back, and if we can optimize it, we definitely will. Even if that comes as a separate PR.

mxr576:
I do not see a reason not changing that ~3 lines in this PR before it gets (where 1 affected line is introduced by this PR) :) but thanks for keeping eye on this @raakesh-blokhra

PR #545, opened 2021-01-05
[#235] Add translation to debug exception errors

### LINKED ISSUE #235: Add translation to debug exception errors
PR #227 adds in the ability to show exception messages on screen, but are not translatable.  

Displaying translatable error messages from exceptions is not easy, but we already have something in the module: KeyProviderRequirementsException.

divya-intelli:
Fixes #235.
This PR adds the translation for the exception added in #227.
translation2
translation1

### DIFF
--- a/src/Exception/DeveloperDoesNotExistException.php
+++ b/src/Exception/DeveloperDoesNotExistException.php
@@ -47,7 +47,7 @@ class DeveloperDoesNotExistException extends ApiException implements ApigeeEdgeE
    */
   public function __construct(string $email, string $message = 'Developer with @email email address not found. Running the developer sync could fix this.', int $code = 0, \Throwable $previous = NULL) {
     $this->email = $email;
-    $message = strtr($message, ['@email' => $email]);
+    $message = \Drupal::translation()->translate($message, ['@email' => $email]);
     parent::__construct($message, $code, $previous);
   }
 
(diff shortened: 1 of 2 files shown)


src/Exception/DeveloperDoesNotExistException.php
```diff
@@ -47,7 +47,7 @@ class DeveloperDoesNotExistException extends ApiException implements ApigeeEdgeE
    */
   public function __construct(string $email, string $message = 'Developer with @email email address not found. Running the developer sync could fix this.', int $code = 0, \Throwable $previous = NULL) {
     $this->email = $email;
-    $message = strtr($message, ['@email' => $email]);
+    $message = \Drupal::translation()->translate($message, ['@email' => $email]);
```
mxr576, line 50: https://www.drupal.org/docs/develop/coding-standards/php-exceptions

> All Exceptions should include an appropriate message and should not be translated. Only messages from the install and update system are currently translated as they are user facing.

This is not how you make an exception translated for users _on the UI_. If an exception can reach the front-end you should capture that with an exception subscriber and based on the exception you should render a translatable message there.

See https://github.com/apigee/apigee-edge-drupal/blob/8.x-1.x/src/EventSubscriber/EdgeExceptionSubscriber.php

Some related issues that I found with Google:
https://www.drupal.org/project/drupal/issues/2055851
https://www.drupal.org/project/search_api_solr/issues/3001220

mxr576 (changes requested):

phdhiren:
Thanks @mxr576 for pointing out https://www.drupal.org/docs/develop/coding-standards/php-exceptions

Will not fix this as :
All Exceptions should include an appropriate message and should not be translated.

PR #542, opened 2020-12-23
Apigee edge debug module - debug logs and recent logs formatting enhancement #333 & #331

shishir-intelli:
Closes #333 & #331 
Enhanced formatting on Apigee edge debug module recent logs and request/response logs to Improve Readability.

1. Removed limitation of showing 10000 characters on detail logs.
2. Request and response differentiated by showing arrows (<<<< / >>>>).
3. In recent logs truncated  "/v1/organizations/***organization***" for better readability of debug logs.

**_Screenshot - Updated recent log page which was earlier showing "/v1/organizations/***organization***"_**
Resolves #331 
recent-log

**_Screenshot - Updated debug logs request and response._**
Resolves #333
debug-log

PR #547, opened 2021-01-07
Removing multiple occurances of @file annotations. Issue# 541.

### LINKED ISSUE #541: Only one @file annotation should be in each file
One thing that I have missed - and I see it is also incorrect in apigee_edge.install and also in module - there should be only one `@file` annotation in the file.

Since it has to be fixed in several places, I am not considering this a blocker.

_Originally posted by @mxr576 in https://github.com/apigee/apigee-edge-drupal/pull/538#r547211849_

raakesh-blokhra:
@file annotations can only be used once in a file. Edited all the .module and .install files that had @file repeated multiple times. 
closes #541

### DIFF
--- a/apigee_edge.api.php
+++ b/apigee_edge.api.php
@@ -20,7 +20,6 @@
  */
 
 /**
- * @file
  * Hooks for apigee_edge module.
  */
 /**
--- a/apigee_edge.install
+++ b/apigee_edge.install
@@ -19,7 +19,6 @@
  */
 
 /**
- * @file
  * Install, update and uninstall functions for Apigee Edge.
  */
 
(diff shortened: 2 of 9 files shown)


phdhiren (approved):
Looks good to me

cnovak (commented):
@raakesh-blokhra do you know why the linter and other test are not running for this PR? I am not sure if this is the valid solution without running the linter myself.  It should run along with all tests when PRs are submitted.

raakesh-blokhra:
> @raakesh-blokhra do you know why the linter and other test are not running for this PR? I am not sure if this is the valid solution without running the linter myself. It should run along with all tests when PRs are submitted.

Interesting, I don't see it for some others as well including [542](https://github.com/apigee/apigee-edge-drupal/pull/542).

PR #579, opened 2021-04-26
[#558] Disable cache while cache_expiration is set to 0 in Apigee Edge Teams

### LINKED ISSUE #558: Changing team status in the APIGEE is not reflected in the portal
In case of updating the team's status- the status is not reflected in the developers portal.
After creating a company and updating the status to Inactive and from Inactive to active.
The cache is set to 0 (zero).

Ver 8.X- 1.20

phdhiren:
Fixes #558

### DIFF
--- a/modules/apigee_edge_teams/src/CompanyMembershipObjectCache.php
+++ b/modules/apigee_edge_teams/src/CompanyMembershipObjectCache.php
@@ -98,6 +98,13 @@ final class CompanyMembershipObjectCache implements CompanyMembershipObjectCache
    * {@inheritdoc}
    */
   public function saveMembership(string $company, CompanyMembership $membership): void {
+    $expiration = $this->persistentCacheExpiration;
+
+    // Do not save in cache, if expiration is set to 0.
+    if (0 === $expiration) {
+      return;
+    }
+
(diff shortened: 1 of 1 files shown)


raakesh-blokhra (approved):
Looks Good to me

modules/apigee_edge_teams/src/CompanyMembershipObjectCache.php
```diff
@@ -98,14 +98,20 @@ public function __construct(CacheFactoryInterface $cache_factory, MemoryCacheFac
    * {@inheritdoc}
    */
   public function saveMembership(string $company, CompanyMembership $membership): void {
+    $expiration = $this->persistentCacheExpiration;
+
+    // Do not save in cache, if expiration is set to 0.
+    if (0 === $expiration) {
```
mxr576, line 104: @phdhiren What about other places? (e.g: app caches) Why just here? Configurable cache expiration is a fundemantal features of Apigee Edge module.

phdhiren: Thanks for the input @mxr576 
Let me add separate ticket to check this across Apigee Edge.

PR #588, opened 2021-05-19
Disabled email address field on edit profile page

raakesh-blokhra:
Internally it was decided to prevent users from editing their email addresses. This code will disable the email field on edit profile page to prevent users from editing their email address.


phdhiren (changes requested):
1. Can we have `form-disabled` class for the element so it is visible understood to the end user that its read-only.
2. If admin is allowed to change email of the user?
3. Can we have test case to support this change?

raakesh-blokhra:
Ok using disabled instead of readonly. 
The admins should also not be able to edit email addresses. 
Closing this PR. Creating a new one since this didn't trigger the checks.

Issue #315, opened 2020-01-03
Not able to login with Social Auth Google module, If Developer exists in Apigee Edge but not in developer portal


varuntit101:
Trying to login with google in developer portal :-
First scenario:-  Developer doesn't exist on Apigee Edge, In this case, as expected a new User is created in developer portal as well a Developer is created in Edge.

Second scenario:- Developer already exist on Edge, but not in Developer portal, In this case, when I am trying to login I am getting following error:-

TypeError: Argument 1 passed to Drupal\apigee_edge\Entity\Developer::setFirstName() must be of the type string, null given, called in /srv/bindings/55b4fa54bbfb4b3cb2ca4d91de385c2a/code/modules/contrib/apigee_edge/src/UserDeveloperConverter.php on line 107 in Drupal\apigee_edge\Entity\Developer->setFirstName() (line 368 of /srv/bindings/55b4fa54bbfb4b3cb2ca4d91de385c2a/code/modules/contrib/apigee_edge/src/Entity/Developer.php)

**Expected behavior**
User should be created in developer portal, If a developer already exists with the same email in Edge.

**Version Info**
Social Auth Google - 8.x-2.0
Apigee Edge - 8.x-1.3
Drupal core - 8.7.11

PFA:-
![screencapture-dev-ferguson-d8-pantheonsite-io-admin-reports-dblog-event-152-2020-01-03-23_22_04](https://user-images.githubusercontent.com/9864661/71739688-29312680-2e80-11ea-8e04-af8dea462a3e.png)

arunz6161:
@varuntit101 could you please run "[developer sync](https://www.drupal.org/docs/8/modules/apigee-edge/synchronize-developers-with-apigee-edge)" to ensure that developers are copied over from edge to Drupal and then let us know if the issue persists.

PR #592, opened 2021-06-11
Fix - developer portal fails to recognize Apigee Registered Developers while using SSO

### LINKED ISSUE #586: Developer Portal Fails to recognize Apigee Registered Developers
We have two Developer Portals [ I'll be addressing them `DP1` and `DP2` hereafter ] which are connected to the same Edge Environment.

**Scenario**

- The developer `X` does not exist on Apigee Edge. This user has never accessed either `DP1` or `DP2`. 
- This new user `X` logs into `DP1` for the first time, therefore, a new user profile `X` is created in `DP1` as well the user `X` is registered as a Developer in Edge.  
-  Now, user `X` heads to `DP2` and tries to login for the first time in `DP2`. Since `DP2` treats `X` as a new user, `DP2` tries to register a new developer profile in Edge. `X` already exists in Edge, however Drupal does not know that the developer is in Edge.
- Instead of syncing with 'X' developer profile, `DP2` is throwing the following error:

`§ TypeError: Argument 1 passed to Drupal\apigee_edge\Entity\Developer::setFirstName() must be of the type string, null given, called in /var/www/html/devportal/web/modules/contrib/apigee_edge/src/UserDeveloperConverter.php on line 107 in Drupal\apigee_edge\Entity\Developer->setFirstName() `
						
**Issues with Existing Solutions**

The `developer sync` cron job in `DP2` would regularly create Drupal accounts for all those Edge Developers unavailable in `DP2`. However, only a small amount of those synced users access the portal. A large number of users would never access the portal in their lifetime. 

**Possible Solutions**

If a developer exists in Edge, the developer profile should be automatically detected and synced by Developer Portal when the user tries to login to the developer portal for the first time.

shishir-intelli:
Closes #586 and #315 

Fixes below issues.
1. Whenever a user logs in using SSO, if First Name, Last Name is null or if the developer already exists in Apigee during this process, the site throws
 `§ TypeError: Argument 1 passed to Drupal\apigee_edge\Entity\Developer::setFirstName() must be of the type string, null given, called in /var/www/html/devportal/web/modules/contrib/apigee_edge/src/UserDeveloperConverter.php on line 107 in Drupal\apigee_edge\Entity\Developer->setFirstName()`.

2. If developer does not exist on both Apigee Edge as well as Drupal portal, then the developer is not registered on Apigee and throws error while doing a POST request to Apigee.
`"error": {
        "code": 400,
        "message": "Following elements are required: Email, User Name, First Name, Last Name",
        "status": "INVALID_ARGUMENT",
        "details": [...`
 
The above issue occurs when First-name/Last-name is null, fixed by providing a default value.

### DIFF
--- a/src/UserDeveloperConverter.php
+++ b/src/UserDeveloperConverter.php
@@ -103,6 +103,13 @@ class UserDeveloperConverter implements UserDeveloperConverterInterface {
     foreach (static::DEVELOPER_PROP_USER_BASE_FIELD_MAP as $developer_prop => $base_field) {
       $setter = 'set' . ucfirst($developer_prop);
       $getter = 'get' . ucfirst($developer_prop);
+
+      // Default value for firstname lastname if null.
+      if ($user->get($base_field)->value === NULL && ($base_field === "first_name" || $base_field === "last_name")) {
+        $base_field_value = $developer->{$getter}() !== NULL ? $developer->{$getter}() : ucfirst($developer_prop);
+        $user->set($base_field, $base_field_value);
+      }
+
       if ($user->get($base_field)->value !== $developer->{$getter}()) {
         $developer->{$setter}($user->get($base_field)->value);
         $successful_changes++;


raakesh-blokhra (approved):
Looks good.

PR #598, opened 2021-06-24
Fix 587 - throw exception when change the email address using PUT /developers for ApigeeX

### LINKED ISSUE #587: Disable editing user email address for logged in user
## Description
When a user edits their email address, it creates a new account in the backend. There has been planning done internally at b/182997472 and b/187928954 and was decided to disable the email textfield on the user edit form. This will not allow a user to change their email address in the Drupal instance.

## Steps to Reproduce
1. Go to Edit Profile
2. Change Email Address
3. Scroll down and Click on Save

## Actual Behavior
What happened after the steps to reproduce.
A new user is created in the backend and the changes are saved on Drupal.
## Expected Behavior
User shall not be able to change their email address.
## Version Info
This can be the version you can see on admin/modules in Drupal or 
the output of this command: `composer show`.  Add Drupal core and
other version information if needed.

8+

shishir-intelli:
Closes #587 

`PUT /developers` throwing `INVALID_ARGUMENT -  rest.interceptor.name_mismatch` error while updating user email address on
Apigee X and hybrid runtime versions 1.5.0 and 1.5.1.

To prevent user email address getting updated on Drupal portal and not on `Apigee edge` due to above issue, we are throwing exception on Drupal portal to prevent updating the email address on Drupal, which creates new account on` Apigee edge` if user again tries to update the same account.

This issue is related to [this](https://github.com/apigee/apigee-client-php/pull/154)

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -1371,6 +1371,14 @@ function apigee_edge_user_presave(UserInterface $account) {
           $logger->error("Unable to save @developer developer's developer id on user.", $context);
         }
       }
+      elseif ($previous->getEdgeErrorCode() === Developer::APIGEE_EDGE_ERROR_CODE_DEVELOPER_EMAIL_MISMATCH) {
+        // Apigee X and hybrid runtime versions 1.5.0 and 1.5.1 a call to change the developer's email address will not work so need
+        // to prevent user email update on drupal as well.
+        // @see https://github.com/apigee/apigee-client-php/issues/153
+        // @see https://github.com/apigee/apigee-edge-drupal/issues/587
+        \Drupal::messenger()->addMessage(t($previous->getMessage()), 'error');
+        throw new \Exception(t($previous->getMessage()));
+      }
     }
     else {
       $context += Error::decodeException($exception);
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -31,6 +31,7 @@ use Drupal\apigee_edge\Element\StatusPropertyElement;
 use Drupal\apigee_edge\Entity\ApiProduct;
 use Drupal\apigee_edge\Entity\AppInterface;
 use Drupal\apigee_edge\Entity\Developer;
+use Drupal\apigee_edge\Exception\DeveloperUpdateFailedException;
 use Drupal\apigee_edge\Exception\UserDeveloperConversionNoStorageFormatterFoundException;
 use Drupal\apigee_edge\Exception\UserDeveloperConversionUserFieldDoesNotExistException;
 use Drupal\apigee_edge\Form\DeveloperSettingsForm;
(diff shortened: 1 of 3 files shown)


giteshk (approved):
LGTM

apigee_edge.module
```diff
@@ -1371,6 +1371,14 @@ function apigee_edge_user_presave(UserInterface $account) {
           $logger->error("Unable to save @developer developer's developer id on user.", $context);
         }
       }
+      elseif ($previous->getEdgeErrorCode() === Developer::APIGEE_EDGE_ERROR_CODE_DEVELOPER_EMAIL_MISMATCH) {
+        // Apigee X and hybrid runtime versions 1.5.0 and 1.5.1 a call to change the developer's email address will not work so need
+        // to prevent user email update on drupal as well.
```
mxr576: Nitpick: s/drupal/Drupal/g s/hybrid/Hybrid/g

Comment line is longer is than 80 characters.

apigee_edge.module
```diff
@@ -1371,6 +1371,14 @@ function apigee_edge_user_presave(UserInterface $account) {
...
+        // @see https://github.com/apigee/apigee-client-php/issues/153
+        // @see https://github.com/apigee/apigee-edge-drupal/issues/587
+        \Drupal::messenger()->addMessage(t($previous->getMessage()), 'error');
```
mxr576: Provide your own, human readable error message. Translating variables should be avoided, especially because the exception message can change anytime on the API backend therefore it becomes a moving target for translators.

apigee_edge.module
```diff
@@ -1371,6 +1371,14 @@ function apigee_edge_user_presave(UserInterface $account) {
...
+        throw new \Exception(t($previous->getMessage()));
```
mxr576: Exception messages are not translatable, also consider throwing a domain specific exception (the module's own exception) instead a built-in one, you could call it like DevelopeUpdateFailed.

src/Entity/Developer.php
```diff
@@ -65,6 +65,13 @@ class Developer extends EdgeEntityBase implements DeveloperInterface {
    */
   const APIGEE_EDGE_ERROR_CODE_DEVELOPER_DOES_NOT_EXISTS = 'developer.service.DeveloperDoesNotExist';
 
+  /**
+   * Developer email in the url doesnot match with the email in the request payload.
+   *
+   * @var string
+   */
+  const APIGEE_EDGE_ERROR_CODE_DEVELOPER_EMAIL_MISMATCH = 'rest.interceptor.name_mismatch';
```
mxr576: Maybe it should start with APIGEE_HYBRID or just APIGEE without "EDGE", since only the new Apigee products throws this exception code.

mxr576 (changes requested):

src/Entity/Developer.php
shishir-intelli: Updated as `APIGEE_HYBRID`

src/Exception/DeveloperUpdateFailedException.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+namespace Drupal\apigee_edge\Exception;
+
+use Apigee\Edge\Exception\ApiException;
+
+/**
+ * This exception is thrown when the developer profile update fails.
+ */
+class DeveloperUpdateFailedException extends ApiException implements ApigeeEdgeExceptionInterface {
```
mxr576: This is not part of the module's public API, make it `final` and properties `private`.
(Unfortunately, we haven't started to use these before the first stable release so every bit of the code can be considered a public API which makes maintenance way more complicated. Although, we can fix our mistakes on new codes.)

apigee_edge.module
```diff
@@ -1371,6 +1372,15 @@ function apigee_edge_user_presave(UserInterface $account) {
           $logger->error("Unable to save @developer developer's developer id on user.", $context);
         }
       }
+      elseif ($previous->getEdgeErrorCode() === Developer::APIGEE_HYBRID_ERROR_CODE_DEVELOPER_EMAIL_MISMATCH) {
+        // Apigee X and Hybrid runtime v1.5.0 and v1.5.1 a call to change the developer's
+        // email address will not work so need to prevent user email update on
+        // Drupal as well.
+        // @see https://github.com/apigee/apigee-client-php/issues/153
+        // @see https://github.com/apigee/apigee-edge-drupal/issues/587
+        \Drupal::messenger()->addMessage("Unable to update developer's profile.", 'error');
```
mxr576: This hook can be triggered in various contexts (eg.: update hooks, CLI commands, (JSON) API request, etc). Displaying a message with messenger only works for HTML requests/responses. Therefore calling a messenger here is not a good solution. (Also, this hook can be triggered in batch which means calling messenger here could trigger 1 to many messages on the UI, which may not be necessary.)
If a message should be displayed for this exception when an HTML response is expected then it should be done in an HTML exception subscriber.


This is one of the reasons why only logs are being created here.

An eye for one, I would not expose more information here for end-users, logs already contain enough information, end-user should just call customer support and report what happened.

mxr576 (changes requested):

apigee_edge.module
shishir-intelli: @mxr576
Removed the Drupal messenger and also updated the exception message, this is how it will look

image

@giteshk let me know if this looks good.

PR #599, opened 2021-07-05
[#583] Allow customization of management API URL for Apigee X/Hybrid

### LINKED ISSUE #583: Allow customization of management API URL for Apigee X/Hybrid
In order to integrate with external Identity Providers, it is common to proxy management API calls to add custom behavior. This might include mediating/orchestrating between the management API and another source. In the past this was easy by using Private Cloud settings and providing your own Management API URL.

Allow the Hybrid [Endpoint](https://github.com/apigee/apigee-client-php/blob/2.x/src/ClientInterface.php#L45) to be overriden using a variable, similar to Apigee Credentials.

1) Events trigger from Audit Logs allow us to create entries in an external IdP on POST requests, but cannot be used to enrich GET requests
2) The Apigee Drupal libraries could be forked and the hardcoded URL could be changed locally, although this causes problems when upgrading.
3) Storing all Apps in Apigee instead of the IdP - this only works in cases where all traffic goes through Apigee (but is the first choice if possible).

## Additional context 
This may also be useful for customers that apply additional security controls between their portal and the management API. It may also be useful in future if there is a self hosted control plane for deployments in regions with strict traffic locality requirements such as Russia/China/Switzerland.

phdhiren:


giteshk:
@phdhiren do we want to merge this?

phdhiren:
@giteshk 
Yes, we want to merge

PR #607, opened 2021-07-23
ApigeeX code for cuj7(change the billing type to default)

divya-intelli:
This PR calls the m10n hook which changes the billing type of the user to the default that is set in portal.


raakesh-blokhra (approved):
LGTM.

apigee_edge.module
```diff
@@ -1317,6 +1317,13 @@ function apigee_edge_user_presave(UserInterface $account) {
     }
     $developer = $result->getDeveloper();
     $developer->save();
+
+    // Check if apigee_m10n_add_credit module enabled and org is ApigeeX
+    // if yes call 'default_billingType' hook.
+    if (\Drupal::moduleHandler()->moduleExists('apigee_m10n_add_credit') && \Drupal::service('apigee_m10n.monetization')->isOrganizationApigeeXorHybrid()) {
```
mxr576, line 1323: Why does the Apigee Edge module need to know about the MINT module or ANY other module that depends on it? It does not. This is a misinterpretation of the dependency graph between components.

If you see a need to react on an event when a user is converted to a developer create a _generic_ event to _this_ module, a new public API that can be leveraged by ANY dependent. Do not add hidden/secret checks.

mxr576 (changes requested):

apigee_edge.module
mxr576: This is how we solved things in Drupal 7 and before but that is not necessary anymore in Drupal 8+.

Actually, I realized there is no need for a new event and there is a good reason why we haven't created one before. User -> Developer conversion is handled by `apigee_edge.converter.user_developer` service and in Drupal 8+ you can use a good-old Symfony trick called "service decorator" to exend or alter the public behavior of that service, in other words the created developer from a user. I would propose to replace the current solution with that in the apigee_m10n_add_credit module, no change is needed here related to this.

divya-intelli: Thanks @mxr576 , thats a great suggestion , I tried and worked as expected.I am going to close this PR and implement the change  in apigee_m10n_add_credit module.
Thanks you so much for your contribution.

divya-intelli:
Closing this PR as this functionality is moved to apigee_m10n_add_credit module.

apigee_edge.module
mxr576: thank you, happy to help anytime :)

PR #628, opened 2021-09-28
Fix for apiproduct access case issue

### LINKED ISSUE #627: Apiproduct is not visible if the access type is set to 'Public' or 'Private'
If the access attribute of an api product is set to 'Public' and 'Private' (notice the capitalization of the first character), it does not match the portal visibility.

divya-intelli:
Fixes #627.

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -515,7 +515,9 @@ function apigee_edge_api_product_access(EntityInterface $entity, $operation, Acc
   $result = AccessResult::allowedIfHasPermission($account, 'bypass api product access control');
   if ($result->isNeutral()) {
     // Attribute may not exists but in that case it means public.
-    $product_visibility = $entity->getAttributeValue('access') ?? 'public';
+    // Access attribute needs to be set to lower case
+    // to match the portal visibilities that is set in the config apigee_edge.api_product_settings.
+    $product_visibility = $entity->getAttributeValue('access') ? strtolower($entity->getAttributeValue('access')) : 'public';
     $visible_to_roles = \Drupal::config($config_name)->get('access')[$product_visibility] ?? [];
 
     // A user may not have access to this API product based on the current


apigee_edge.module
```diff
@@ -515,7 +515,9 @@ function apigee_edge_api_product_access(EntityInterface $entity, $operation, Acc
   $result = AccessResult::allowedIfHasPermission($account, 'bypass api product access control');
   if ($result->isNeutral()) {
     // Attribute may not exists but in that case it means public.
-    $product_visibility = $entity->getAttributeValue('access') ?? 'public';
+    // Access attribute needs to be set to lower case
+    // to match the portal visibilities that is set in the config apigee_edge.api_product_settings.
+    $product_visibility = $entity->getAttributeValue('access') ? strtolower($entity->getAttributeValue('access')) : 'public';
```
mxr576, line 520: I am currently :heavy_minus_sign:  on this change.

The management UI also sets lowercase visibility values so if values are not stored lowercased by the form on apigee_edge.api_product_settings then that feature needs to be fixed, not the access control implementation that uses the stored values from Apigee.

image

mxr576 (changes requested):

apigee_edge.module
mxr576: Probably it is not a good idea to change all access values on existing API products in an update hook. I would rather emphasize in release notes that the configuration UI worked incorrectly before and every project should validate the value in the `access` attribute. Why? Because they may unintentionally "leveraging" this misbehavior on their site by comparing `Public` with `public` or `Private` with `private`, and do not forget, this is an attribute, so literally, the value can be anything...
image

mxr576: (This also means that adding the suggested workaround can expose information on existing projects that should not be.)

giteshk: @mxr576 the backend team is going to change the case for the access attribute. 

This will be needed till they roll out the fix. We may be in a state where the attribute could be in different case for a little time.

giteshk: There is additional information on the ticket to make this field an ENUM to only accept certain values.

giteshk (approved):

PR #629, opened 2021-10-04
Fix 624 - updated the custom hook to use argument and fix minor issues

### LINKED ISSUE #624: Rollback or fix #609 or find an alternative solution
## Description
PR #609 was merged but there are several problems with it.

Approach 1) If we would like to keep the _current_ solution then it should be fixed, see review comments on https://github.com/apigee/apigee-edge-drupal/commit/1f15f9148f15066c6ba7584f8e9c56ac5ce654e7

Approach 2) Since even Drupal core is tending to get away Drupalism called "alter hooks" and use events, the SDKConnector should rather trigger an event instead of exposing an alter hook.

Approach 3) (This would be my preference) instead of extending the public API of the module "unnecessarily", I would verify what changes are needed to be made (if any) to allow a simple SDK connector _service decorator_ to change (extend or override) the user agent string. 

## Version Info
Latest master.

shishir-intelli:
Fix #624 with approach #1

1. Module name prefixed on custom hook.
2. Implemented return type.
3. Function updated with argument (`$user_agent`).
4. Removed custom hook implement from .module file.

### DIFF
--- a/apigee_edge.api.php
+++ b/apigee_edge.api.php
@@ -37,16 +37,12 @@ function hook_apigee_edge_app_listing_page_title_alter(\Drupal\Core\StringTransl
 }
 
 /**
- * Build user agent string.
+ * Extends the default user-agent information sent by the API client.
  *
- * @param array $user_agent
- *   The user agent string data.
- *
- * @return string
- *   A string with module info.
+ * @param array $user_agent_parts
+ *   The user agent string.
  */
-function hook_apigee_edge_user_agent_string(array &$user_agent) {
-  return $user_agent;
+function hook_apigee_edge_user_agent_string_alter(array &$user_agent_parts): void {
 }
 
 /**
--- a/apigee_edge.api.php
+++ b/apigee_edge.api.php
@@ -37,9 +37,16 @@ function hook_apigee_edge_app_listing_page_title_alter(\Drupal\Core\StringTransl
 }
 
 /**
- * Alters the user agent string.
+ * Build user agent string.
+ *
+ * @param array $user_agent
+ *   The user agent string data.
+ *
+ * @return string
+ *   A string with module info.
  */
-function hook_user_agent_string_alter() {
+function hook_apigee_edge_user_agent_string(array &$user_agent) {
+  return $user_agent;
 }
 
 /**
--- a/src/SDKConnector.php
+++ b/src/SDKConnector.php
@@ -255,13 +255,14 @@ class SDKConnector implements SDKConnectorInterface {
       if (!isset($edge_module_info['version'])) {
         $edge_module_info['version'] = '2.x-dev';
       }
-      $user_agent[] = $edge_module_info['name'] . '/' . $edge_module_info['version'];
+      $user_agent_parts[] = $edge_module_info['name'] . '/' . $edge_module_info['version'];
+      $user_agent_parts[] = 'Drupal/' . \Drupal::VERSION;
 
       // Get info from other modules.
-      $userAgent = \Drupal::moduleHandler()->invokeAll('apigee_edge_user_agent_string', [&$user_agent]);
-      $userAgent = !empty($userAgent) ? implode('; ', $userAgent) : implode('; ', $user_agent);
+      $userAgent = $this->moduleHandler->invokeAll('apigee_edge_user_agent_string_alter', [&$user_agent_parts]);
+      $userAgent = !empty($userAgent) ? implode('; ', $userAgent) : implode('; ', $user_agent_parts);
 
-      self::$userAgentPrefix = $userAgent . '; Drupal/' . \Drupal::VERSION;
+      self::$userAgentPrefix = $userAgent;
     }
 
     return self::$userAgentPrefix;
--- a/src/SDKConnector.php
+++ b/src/SDKConnector.php
@@ -250,8 +250,16 @@ class SDKConnector implements SDKConnectorInterface {
    */
   protected function userAgentPrefix(): string {
     if (NULL === self::$userAgentPrefix) {
-      $userAgent = \Drupal::moduleHandler()->invokeAll('user_agent_string_alter');
-      $userAgent = implode('; ', $userAgent);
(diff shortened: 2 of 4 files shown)


src/SDKConnector.php
```diff
@@ -250,8 +250,16 @@ private function buildCredentials(KeyInterface $key): CredentialsInterface {
...
-      $userAgent = implode('; ', $userAgent);
+      // apigee_edge module info.
+      $edge_module_info = $this->infoParser->parse($this->moduleHandler->getModule('apigee_edge')->getPathname());
+      if (!isset($edge_module_info['version'])) {
+        $edge_module_info['version'] = '2.x-dev';
+      }
+      $user_agent[] = $edge_module_info['name'] . '/' . $edge_module_info['version'];
+
+      // Get info from other modules.
+      $userAgent = \Drupal::moduleHandler()->invokeAll('apigee_edge_user_agent_string', [&$user_agent]);
```
mxr576: Module handler has to be dependency-injected to the service. It also should be added as a nullable dependency with failover and a warning that it becomes a mandatory parameter in the next major version (as you can see in Drupal core).

src/SDKConnector.php
```diff
@@ -250,8 +250,16 @@ private function buildCredentials(KeyInterface $key): CredentialsInterface {
...
+      $userAgent = !empty($userAgent) ? implode('; ', $userAgent) : implode('; ', $user_agent);
 
       self::$userAgentPrefix = $userAgent . '; Drupal/' . \Drupal::VERSION;
```
mxr576: Test coverage should be extended/added for the new alterable/extendable user agent.

src/SDKConnector.php
```diff
@@ -250,8 +250,16 @@ private function buildCredentials(KeyInterface $key): CredentialsInterface {
...
   protected function userAgentPrefix(): string {
     if (NULL === self::$userAgentPrefix) {
-      $userAgent = \Drupal::moduleHandler()->invokeAll('user_agent_string_alter');
-      $userAgent = implode('; ', $userAgent);
+      // apigee_edge module info.
+      $edge_module_info = $this->infoParser->parse($this->moduleHandler->getModule('apigee_edge')->getPathname());
+      if (!isset($edge_module_info['version'])) {
+        $edge_module_info['version'] = '2.x-dev';
+      }
+      $user_agent[] = $edge_module_info['name'] . '/' . $edge_module_info['version'];
```
mxr576: I would call it `$user_agent_parts`

apigee_edge.api.php
```diff
@@ -37,9 +37,16 @@ function hook_apigee_edge_app_listing_page_title_alter(\Drupal\Core\StringTransl
...
+ * Build user agent string.
+ *
+ * @param array $user_agent
+ *   The user agent string data.
+ *
+ * @return string
+ *   A string with module info.
  */
-function hook_user_agent_string_alter() {
+function hook_apigee_edge_user_agent_string(array &$user_agent) {
```
mxr576: I would call it `$user_agent_parts`

mxr576: Missing `void` return type.

apigee_edge.api.php
```diff
@@ -37,9 +37,16 @@ function hook_apigee_edge_app_listing_page_title_alter(\Drupal\Core\StringTransl
...
+  return $user_agent;
```
mxr576: If the variable is passed by reference, no return is needed. (I more real life-ish example would be better :) )

apigee_edge.api.php
```diff
@@ -37,9 +37,16 @@ function hook_apigee_edge_app_listing_page_title_alter(\Drupal\Core\StringTransl
 }
 
 /**
- * Alters the user agent string.
+ * Build user agent string.
```
mxr576: I would: Extends the default user-agent information sent by the API client.

src/SDKConnector.php
mxr576: It is still looks odd that the module version can be removed from the user agent string via an alter hook (yes, that is possible) but not the Drupal version suffix.
Maybe the alter hook should allow altering the complete string (e.g: `Apigee Edge/2.x-dev;Drupal/8.9.10`) and in an alter hook implementation the string should be parsed, exploded and imploded again as needed. (We intentionally introduced a format that can be easily parsed ;) ).

mxr576 (changes requested):

src/SDKConnector.php
mxr576: @shishir-intelli First I think we should get an agreement on this because other suggested changes could become unnecessary :)

mxr576: The other option is the alter hook still gets an array of user agent string parts (Component and version) that includes Apigee Edge version and Drupal version by default and a hook implementation can add new items to that array in any position, or even remove items. The user agent string would be still built in the SdkConnector to the expected format. (Probably that is the better option due to the previous sentence)

apigee_edge.api.php
mxr576: How about `hook_apigee_edge_user_agent_string_alter()`?

apigee_edge.api.php
```diff
@@ -37,9 +37,16 @@ function hook_apigee_edge_app_listing_page_title_alter(\Drupal\Core\StringTransl
...
+ *
+ * @param array $user_agent
+ *   The user agent string data.
+ *
+ * @return string
+ *   A string with module info.
```
mxr576: "The user agent string."

src/SDKConnector.php
shishir-intelli: Yes, By default Apigee Edge version and Drupal version in array and in hook implementation one can add new items or update the position, In our case m10 module info is placed at beginning of array. so now the format looks like 
`Apigee Monetization/8.x-1.11; Apigee Edge/8.x-1.24; Drupal/9.2.6 (Apigee Edge PHP Client 2.0.10; PHP/# )`

phdhiren (approved):
LGTM

PR #630, opened 2021-10-07
Fixes #576 to allow multiple protocols for callback URL

### LINKED ISSUE #576: Callback URL forces http/https
(basically recreating https://github.com/apigee/apigee-devportal-kickstart-drupal/issues/447)

Description
When editing app, it enforces callback url to begin with http or https, if trying to change the regexp pattern to allow .* for example it still forces http/https. Basically a native client will have its own reserved URL scheme and would require to register "myapp://" as their callback.

In addition, if you configure and app in Apigee edge with custom url scheme callback then the module crashes for the user who owns that app once visiting their list of apps in the portal

"Callback URL External site to which an app consumer is redirected to log in when using three-legged OAuth. Valid values must be prefixed with one of the following protocols: http:// or https://"

Steps to Reproduce

Apps> App > Edit > Set callback to myapp:// > Save (does not work)
or
Go to Apigee Edge, create an app with callback url myapp://
Go to Developer portal > Apps > Crash happens
Actual Behavior
Crash/Bug see above

Expected Behavior
Apigee edge module should allow the entry of custom callback URLs.  The ability to onboard native apps using a custom url is an important feature to our customers and is a component of three-legged OAuth with PKCE flow.

Screenshots:
104913869-8cf49780-598e-11eb-95d9-57912bdd5a2e

extra:
drupal core only allows a list of accepted protocols for routing found in protocol_filter list in web/stes/default/services.yml (/var/www/html/web/core/lib/Drupal/Core/Utility/UnroutedUrlAssembler.php)?
```
InvalidArgumentException: The URI 'sap://mynativeapp' is invalid. You must use a valid URI scheme. Use base: for a path, e.g., to a Drupal file that needs the base path. Do not use this for internal paths controlled by Drupal. in Drupal\Core\Utility\UnroutedUrlAssembler->assemble() (line 65 of /var/www/html/web/core/lib/Drupal/Core/Utility/UnroutedUrlAssembler.php).
```

giteshk:
Fixes #576

### DIFF
--- a/config/install/core.entity_view_display.developer_app.developer_app.default.yml
+++ b/config/install/core.entity_view_display.developer_app.developer_app.default.yml
@@ -14,7 +14,7 @@ content:
     region: content
     settings: {  }
     third_party_settings: {  }
-    type: uri_link
+    type: basic_string
   createdAt:
     type: timestamp_ago
     label: inline
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -383,10 +383,7 @@ function apigee_edge_entity_extra_field_info() {
  * Implements hook_field_formatter_info_alter().
  */
 function apigee_edge_field_formatter_info_alter(array &$info) {
-  // Allows using the 'uri_link' formatter for the 'app_callback_url'
-  // field type, which uses it as default formatter.
-  // @see \Drupal\apigee_edge\Plugin\Field\FieldType\AppCallbackUrlItem
-  $info['uri_link']['field_types'][] = 'app_callback_url';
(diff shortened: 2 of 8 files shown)


config/install/core.entity_view_display.developer_app.developer_app.default.yml
```diff
@@ -14,7 +14,7 @@ content:
     region: content
     settings: {  }
     third_party_settings: {  }
-    type: uri_link
+    type: basic_string
```
mxr576, line 17: This should require an update hook for existing sites, isn't it?

PR #636, opened 2021-10-25
Incorrect organization now returns Forbidden

divya-intelli:
Updated the test as tests was failing if incorrect organization is entered while setting up the connection as the error message is different.


raakesh-blokhra (approved):
LGTM

PR #639, opened 2021-11-10
Switching from PHP Serialization/Deserialization to JSON Encode/Decode

### LINKED ISSUE #641: Switching from PHP Serialization/Deserialization to JSON Encode/Decode
## Description
Replace PHP Serialization/Deserialization to JSON Encode/Decode. 

Drupal 8/9

shishir-intelli:
Closes #641 
Switching from PHP Serialization/Deserialization to JSON function.

### DIFF
--- a/src/OauthTokenFileStorage.php
+++ b/src/OauthTokenFileStorage.php
@@ -20,6 +20,7 @@
 namespace Drupal\apigee_edge;
 
 use Drupal\apigee_edge\Exception\OauthTokenStorageException;
+use Drupal\Component\Serialization\Json;
 use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\File\Exception\FileException;
 use Drupal\Core\File\FileSystemInterface;
@@ -171,7 +172,7 @@ final class OauthTokenFileStorage implements OauthTokenStorageInterface {
     try {
       $this->checkRequirements();
       // Write the obfuscated token data to a private file.
-      $this->fileSystem->saveData(base64_encode(serialize($data)), $this->tokenFilePath, FileSystemInterface::EXISTS_REPLACE);
+      $this->fileSystem->saveData(base64_encode(Json::encode($data)), $this->tokenFilePath, FileSystemInterface::EXISTS_REPLACE);
     }
     catch (FileException $e) {
       $this->logger->critical('Error saving OAuth token file.');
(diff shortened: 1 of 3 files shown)


mxr576:
Reported and accepted security issues must not be fixed in a public issue queue, it must be fixed in a private repo and only exposed in the public repo when a release with the fix is available. Also, the title of the PR/commit title and description must not imply that it fixes a security issue.
 :scream: :see_no_evil: (cc @giteshk )

src/OauthTokenFileStorage.php
```diff
@@ -263,7 +264,7 @@ private function getFromStorage(): array {
     $data = [];
     // Get the token data from the file store.
     if (file_exists($this->tokenFilePath) && ($raw_data = file_get_contents($this->tokenFilePath))) {
-      $data = unserialize(base64_decode($raw_data));
```
mxr576, line 266: I do not have access to the original security issue, but the token is coming from a trusted system  and if there is a MTM attack in place between Drupal and the Apigee instance, it is already pretty bad and this does not fix that problem. 
So my question is, does this fix really necessary?
https://www.php.net/manual/en/function.unserialize.php

json_decode() has a worse performance: https://grechin.org/2021/04/06/php-json-encode-vs-serialize-performance-comparison.html

giteshk: @mxr576 We evaluated that it was very highly unlikely that this was going to be an issue. It would need for someone to gain access to the servers to make this changes. As per the drupal security team it is not a bug and hence the direction to fix it. 

I do not see performance as an issue since we are probably caching the token (memory cache not drupal cache). 
@shishir-intelli to confirm the last statement.

shishir-intelli: @giteshk Yes correct, just storing the token in `oauth.dat` file using Json encode/decode. It wont impact the performance.

mxr576: But my point is that the data stored in the file gets decoded every time when a new sdk connector service is being initiated for making API requests. This may not have a performance impact ofc, I have just wanted call your attention that it may requires profiling. 

So it is not going to be promoted as a security fix after all?

giteshk: Yep it is not promoted as a security fix. 

Thank you for bringing into notice the potential impact this could have.

giteshk (approved):

PR #652, opened 2021-12-29
Updated test case according to Drupal 9.3

### LINKED ISSUE #651: Update test case for DeveloperTest
There are few string changes with Drupal 9.3.0 which leads to fail the test case developerCreateByAdminTest. It should be update as per latest.

Reference:

https://www.drupal.org/project/drupal/issues/3199972

https://github.com/apigee/apigee-edge-drupal/blob/18ebaa05866a72bb887c2a32a23ef2448027c8d4/tests/src/Functional/DeveloperTest.php#L381

phdhiren:
Fixes #651

https://www.drupal.org/project/drupal/issues/3199972


raakesh-blokhra (approved):
LGTM

Issue #677, opened 2022-03-29
Add Support for uuid on Team Entity


audave:
As a Drupal developer, I would like the Team Entity to have a uuid & uuid() function so that it plays nicely with Drupal Core's jsonapi module. 

Drupal core has inbuilt API functionality provided by the jsonapi module. This module relies on Entities having a uuid & uuid() function in order provide unique identifiers for each entity & through that a unique url to read/update entities. 

Quick fix available at https://github.com/apigee/apigee-edge-drupal/pull/678

PR #678, opened 2022-03-29
Add uuid() function to Team Entity so that this module places nicely …

audave:
Quick and dirty fix for issue raised in https://github.com/apigee/apigee-edge-drupal/issues/677


modules/apigee_edge_teams/apigee_edge_teams.services.yml
```diff
@@ -130,3 +130,12 @@ services:
     arguments: ['@entity_type.manager', '@logger.channel.apigee_edge_teams']
     tags:
       - { name: paramconverter }
+
+  paramconverter.apigee_edge_teams.entity_uuid:
+    parent: paramconverter.entity
+    class: Drupal\apigee_edge_teams\ParamConverter\InvitationsEntityUuidConverter
```
shishir-intelli, line 136: @audave 
I am not sure why we need a InvitationsEntityUuidConverter paramconverter which is already there in JsonAPI module (EntityUuidConverter)? 
This will also create a dependency of JsonAPI module.
IMO you should make use of uuid() and EntityUuidConverter instead of a new converter.

modules/apigee_edge_teams/src/Entity/Team.php
```diff
@@ -347,4 +347,14 @@ protected static function propertyToBaseFieldBlackList(): array {
     ]);
   }
 
+  /**
+   * This is required by jsonapi.
+   *
+   * This is here because the jsonapi module relies on a uuid to build end points and breaks if one is not present.
```
mxr576, line 353: Nitpick: JSONAPI, UUID

modules/apigee_edge_teams/src/Entity/Team.php
```diff
@@ -347,4 +347,14 @@ protected static function propertyToBaseFieldBlackList(): array {
...
+   * Creating a real UUID for this entity type is not possible as entity data is stored on Apigee.
```
mxr576, line 354: But it is not a problem because the JSONAPI _specification_ also allows non-valid UUID as universal identifiers until their uniqueness is ensured.
https://jsonapi.org/format/#crud-creating-client-ids

mxr576 (changes requested):
There is a potential and not negligible issue with adding this fix to the code: **potential information disclosure vulnerability.**


Team/Company on Apigee has no UUID like unique identifier, only a `name` that can be anything (it usually a machine name version of the display name, but that is also not enforced by the API gateway), therefore we have to use the `name` as unique id in Drupal.

How JSON API works OOTB that it exposes if a given resource is not available for the current user (see anonymous on this screenshot) and that resource contains the value returned by the `uuid()` method on the entity...

image

mxr576:
Can be unrelated but we also bumped into [this guard condition in Drupal core](https://github.com/drupal/core/blame/a42f6328924ca3b55127571f5a32d14fda8869d3/modules/jsonapi/src/Normalizer/JsonApiDocumentTopLevelNormalizer.php#L306-L308), so I wonder, if a team entity (with a non-uuid name) can be created with a JSON API with this fix or not. (Probably it will work because `name` field can be used in the request instead of `id`, but it worth double checking)

thursdaybw:
This patch causes the following error when installing the apigee_edge_teams module:


```
$ drush en -y apigee_edge_teams apigee_edge
The following module(s) will be enabled: apigee_edge_teams, apigee_edge, entity, key

 // Do you want to continue?: yes.                                                                                      


In SqlContentEntityStorage.php line 1588:
                                                                                                                         
  Exception thrown while performing a schema update. The 'uuid' field specification does not define 'not null' as TRUE.  
                                                                                                                         

In Schema.php line 702:
                                                                      
  The 'uuid' field specification does not define 'not null' as TRUE.  
 ```

PR #701, opened 2022-05-10
Disable email notification and approve

### LINKED ISSUE #691: Add team members intermittently not adding members to team
## Description
Sometimes when we try to add members to team it was not adding even though we get message saying it successfully added to the team .We cannot see that members under teams and also we cannot verify backend (apigee managment api) because we cannot get  member details using management api.

## Steps to Reproduce
1.Selected team one and removed member X
2. Selected team two and added the member X.
3. Removed X from team two
4. Added back member X to team two
5. Removed X from team two
6. Added  back member X to team two -->This is where its not adding .Looks like cache object queuing up and cause the issue .

## Actual Behavior
Member not adding members to team eventhough we got messgae :-The following developer has been invited to the team two team: X.

## Expected Behavior
We should be seeing member X under team team two.

## Screenshots
testteam_member_after adding pking
testteam_members
team one after remove user pking
Team one members

## Version Info
This can be the version you can see on admin/modules in Drupal or 
the output of this command: `composer show`.  Add Drupal core and
other version information if needed.

drupal/core                               9.3.12
drupal/apigee_edge                        2.0.2

phdhiren:
Resolves #691

### DIFF
--- a/modules/apigee_edge_teams/src/Entity/Storage/TeamInvitationStorage.php
+++ b/modules/apigee_edge_teams/src/Entity/Storage/TeamInvitationStorage.php
@@ -113,7 +113,12 @@ class TeamInvitationStorage extends SqlContentEntityStorage implements TeamInvit
 
     switch ($hook) {
       case 'insert':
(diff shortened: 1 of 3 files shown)


raakesh-blokhra (approved):
LGTM

PR #702, opened 2022-05-12
Access checking must not happen in the view layer

mxr576:
In addition, API products that are assigned to an app must be always displayed.

(They are also visible on the app edit form even if the current app owner does not have access to them.)


raakesh-blokhra (commented):
Thank you for your contributions as always. Can you please create issues related to the PR so we can understand the issue(s)?

mxr576:
Hm, sorry, I was a bit brief in the description of this PR, I wanted to fix that but today I realized that this access check here might works as it should - even if it should happen in different layer of the application - just it interferes with my custom app implementation where an app is not owned by the logged in developer (`_apigee_edge_user_has_an_app_with_product()`).

Closing this PR for now and I'll reopen if it becomes necessary.

PR #703, opened 2022-05-12
Add missing JS library dependency

mxr576:
for displaying a modal windows


mxr576:
My gut says the current code only works if some other code has already loaded this library (JS) dependency, but if not, modal is not being opened. This code ensures the library is always there when it is needed.

shishir-intelli (approved):
Thanks @mxr576,
LGTM

mxr576:
Thx @shishir-intelli , please merge it. I do not have permission to do it.

PR #704, opened 2022-05-12
Contruct the app keys url on the backend

mxr576:


mxr576:
The front-end should have "hard-coded" paths or try to guess paths when the backend generates those.

(This also allows extending the existing system much easier.)

shishir-intelli (approved):
Thanks @mxr576 for identifying this and thanks for your contribution.. 
Looks good to me.

PR #728, opened 2022-07-28
Fix for deprecation:Relying on entity queries to check access by default

### LINKED ISSUE #684: Relying on entity queries to check access by default is deprecated in drupal:9.2.0
Relying on entity queries to check access by default is deprecated in drupal:9.2.0 and an error will be thrown from drupal:10.0.0. Call \Drupal\Core\Entity\Query\QueryInterface::accessCheck() with TRUE or FALSE to specify whether access should be checked.

Change Record https://www.drupal.org/node/3201242

divya-intelli:
Closes #684


phdhiren (approved):
LGTM

PR #738, opened 2022-09-09
Fix Drush si profile installation error due to missing Apigee configuration

### LINKED ISSUE #731: Apigee Edge throws errors on site install.
## Description
When installing a new site with apigee_edge, that does not yet have a connection set up, the apigee_edge module throws hard errors when a connection key is not found.

## Steps to Reproduce
1. Install a site with `$ drush site:install apigee_devportal_kickstart`
2. View errors in output.

## Actual Behavior
You will see an error that a key to connect to apigee does not exist.

image

## Expected Behavior
If the key hasn't been set yet, the system shouldn't throw a hard error. How about a warning message?

## Version Info
Drupal core 9.4.5
Apigee_edge 2.0.4
DevPortal_Kickstart 2.1.0

shishir-intelli:
Fix #731

Fix error `Could not create a developer entity` while installation through `Drush si`. Due to missing Apigee edge config since `drush si` cannot set default values for the form.

### DIFF
--- a/tests/src/Functional/DeveloperAppFieldTest.php
+++ b/tests/src/Functional/DeveloperAppFieldTest.php
@@ -154,11 +154,15 @@ class DeveloperAppFieldTest extends ApigeeEdgeFunctionalTestBase {
       strtolower($this->randomMachineName()) => [
         'type' => 'float',
         'data' => [
-          ['value' => M_PI],
-          ['value' => M_E],
-          ['value' => M_EULER],
+          ['value' => round(M_PI, 5)],
+          ['value' => round(M_E, 5)],
+          ['value' => round(M_EULER, 5)],
         ],
-        'encoded' => implode(',', [M_PI, M_E, M_EULER]),
+        'encoded' => implode(',', [
+          round(M_PI, 5),
+          round(M_E, 5),
+          round(M_EULER, 5)
+        ]),
       ],
       strtolower($this->randomMachineName()) => [
         'type' => 'integer',
@@ -179,15 +183,15 @@ class DeveloperAppFieldTest extends ApigeeEdgeFunctionalTestBase {
         'type' => 'list_float',
         'settings' => [
           'settings[allowed_values]' => implode(PHP_EOL, [
-            M_PI,
-            M_E,
-            M_EULER,
+            round(M_PI, 5),
+            round(M_E, 5),
+            round(M_EULER, 5),
           ]),
         ],
         'data' => [
-          ['value' => M_PI],
+          ['value' => round(M_PI, 5)],
         ],
-        'encoded' => (string) M_PI,
+        'encoded' => (string) round(M_PI, 5),
       ],
       strtolower($this->randomMachineName()) => [
         'type' => 'list_integer',
--- a/tests/src/Functional/DeveloperAppFieldTest.php
+++ b/tests/src/Functional/DeveloperAppFieldTest.php
@@ -154,14 +154,14 @@ class DeveloperAppFieldTest extends ApigeeEdgeFunctionalTestBase {
       strtolower($this->randomMachineName()) => [
         'type' => 'float',
         'data' => [
-          ['value' => round(M_PI, 5)],
-          ['value' => round(M_E, 5)],
-          ['value' => round(M_EULER, 5)],
+          ['value' => round(M_PI, 10)],
+          ['value' => round(M_E, 10)],
+          ['value' => round(M_EULER, 10)],
         ],
         'encoded' => implode(',', [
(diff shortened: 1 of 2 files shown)


dpagini (commented):
Will this bit of code **ONLY** work when using the kickstart apigee/apigee-devportal-kickstart-drupal#583 profile? B/c this solution will still be a problem for anyone who wants to use this module but not use the profile...

shishir-intelli:
> Will this bit of code **ONLY** work when using the kickstart [apigee/apigee-devportal-kickstart-drupal#583](https://github.com/apigee/apigee-devportal-kickstart-drupal/pull/583) profile? B/c this solution will still be a problem for anyone who wants to use this module but not use the profile...

@dpagini 
No, IMO there wont be any issue,
This fix is only for non-interactive profile installations like `Drush si`, 

If someone is not using Apigee profile and using this module there wont be a problem because the user can configure the Apigee server setting on UI,
In-fact, the cause of this issue is while creating `developer` in Apigee edge during profile installation and not during module installation.

dpagini:
I'm not using the kickstart profile and simply installing drupal with `$ drush si`. This error is popping up when Drupal itself is creating/updating/saving the USER1 account. So this fix I don't expect will work for me.

Couldn't you, inside the `hook_user_save` just determine if the Apigee connection is active or not, and if it's not, don't try to send the user to apigee?

I would actually suspect right now that you could reproduce this same error with the following steps...
1. Install a blank Drupal site with `standard` profile.
2. Go to `Extend` page and install this module, `apigee_edge`.
3. Go to the User page and create a new user.

I would suspect that you get an error with the above steps since, after installing the module, you did not immediately navigate to the module configuration page and set up the connection...  I'm not 100% sure on that, but that's what it looks like to me.

So this PR wouldn't address that scenario... right?

shishir-intelli:
> So this PR wouldn't address that scenario... right?

Correct, This error is expected if `Apigee configuration` is not set and Drupal portal trying to create a developer in `Apigee Edge`.
This PR is only to prevent error which is thrown during `Drush:si` Apigee kickstart devportal installation. 
Also we can't determine if the Apigee connection is active or not because of missing Apigee config.

If you want to install a regular Drupal with Apigee edge module in it, I will advise you to do Drupal installation first and then follow the installation instructions on this page:
https://www.drupal.org/project/apigee_edge#Installing_24

You can also configure Apigee from `Drush` through this command:
`drush key-save apigee_edge_connection_default '{"auth_type":"basic","organization":"APIGEE_ORG","username":"APIGEE_EMAIL","password":"APIGEE_PASSWORD"}' --key-type=apigee_auth -y; drush --uri=https://SITE_URI uli`
Note : Replace _APIGEE_ORG_, _APIGEE_EMAIL_,  _APIGEE_PASSWORD_, _SITE_URI_ with proper values

dpagini:
> This PR is only to prevent error which is thrown during Drush:si Apigee kickstart devportal installation.

But that's not what the bug report is... To me, this should be entirely fixed in the `apigee_edge` module.

> Also we can't determine if the Apigee connection is active or not because of missing Apigee config.

Not sure what this means? If there is config missing, couldn't you assume the connection is not set up?

shishir-intelli:
> I would actually suspect right now that you could reproduce this same error with the following steps...
> 
> 1. Install a blank Drupal site with `standard` profile.
> 2. Go to `Extend` page and install this module, `apigee_edge`.
> 3. Go to the User page and create a new user.
> 
> I would suspect that you get an error with the above steps since, after installing the module, you did not immediately navigate to the module configuration page and set up the connection... I'm not 100% sure on that, but that's what it looks like to me.

@dpagini ,
 No, if the` Apigee Edge` connection setting is not configured then it wont allow you to create/update any developer in Drupal site, 
It will show an error message like in the screenshot below:
image

I was able to reproduce this issue only during `Drush:si`,

However, let me update this PR to check for active connection and handle the error accordingly.

dpagini:
Oh that's awesome, thank you so much. I will give this a try on my app and see if it addresses my issue...

tests/src/Functional/DeveloperAppFieldTest.php
```diff
@@ -154,11 +154,15 @@ protected function fieldStorageFormattersTest() {
       strtolower($this->randomMachineName()) => [
         'type' => 'float',
         'data' => [
-          ['value' => M_PI],
-          ['value' => M_E],
-          ['value' => M_EULER],
+          ['value' => round(M_PI, 10)],
```
phdhiren, line 157: Please create an issue to revert this test case in a separate ticket.

phdhiren (approved):
LGTM

PR #744, opened 2022-09-29
Fix to show teams more that 100

### LINKED ISSUE #743: Developer part of 100+ teams does not display in the UI
If a developer is a part of 100+ teams the UI only shows the first 100 .

divya-intelli:
Fixes #743

Added permission 'View extensive teams list ' to show all the teams that the developer is part of. 
Create a role for all those developers who have more than 100 teams and assign the new permission to the role so that the other developers are not affected. 

Note : This permission will have performance impact. This permission should be given to users who belong to more than 100 teams.

### DIFF
--- a/modules/apigee_edge_teams/src/Entity/TeamAccessHandler.php
+++ b/modules/apigee_edge_teams/src/Entity/TeamAccessHandler.php
@@ -98,14 +98,17 @@ final class TeamAccessHandler extends EntityAccessControlHandler implements Enti
           // (Reminder, anonymous user can not be member of a team.
           /** @var \Drupal\apigee_edge\Entity\DeveloperInterface|null $developer */
           $developer = $this->developerStorage->load($account->getEmail());
+          $developer_team_ids = $developer->getCompanies();
           $developer_team_access = FALSE;
-          if ($developer && in_array($entity->id(), $developer->getCompanies())) {
+
+          if ($developer && in_array($entity->id(), $developer_team_ids)) {
             $developer_team_access = TRUE;
           }
           else {
             // Check if current developer is a member of the team and has the permision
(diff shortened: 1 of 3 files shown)


modules/apigee_edge_teams/src/Entity/TeamAccessHandler.php
```diff
@@ -86,7 +98,22 @@ protected function checkAccess(EntityInterface $entity, $operation, AccountInter
...
           /** @var \Drupal\apigee_edge\Entity\DeveloperInterface|null $developer */
           $developer = $this->developerStorage->load($account->getEmail());
+          $developer_team_access = FALSE;
           if ($developer && in_array($entity->id(), $developer->getCompanies())) {
+            $developer_team_access = TRUE;
+          }
+          else {
+            // Check if current developer is a member of the team and has the permision
+            // to view more than 100 teams.
+            if ($account->hasPermission('view extensive team list')) {
```
phdhiren: Would like to add one more condition, if the developer's companies array is 100 then only it should proceed. Otherwise it makes no sense to check individually

modules/apigee_edge_teams/src/Entity/TeamPermissionProvider.php
```diff
@@ -75,6 +75,16 @@ public function buildPermissions(EntityTypeInterface $entity_type) {
       ]),
     ];
 
+    $permissions["view extensive team list"] = [
```
phdhiren, line 78: Can we add `{$entity_type_id}` instead of `team`?

phdhiren (changes requested):

phdhiren (approved):
LGTM

PR #755, opened 2022-11-07
Fix for drupal check prophesize method deprecation

### LINKED ISSUE #754: Github action fails for drupal check
## Description
Github action fails for  drupal check with following error:
'Call to deprecated method prophesize() of class PHPUnit\Framework\TestCase'
prophesize() funtion is deprecated in Phpunit 9.5.

divya-intelli:
Fixes #754
In Phpunit 9.5 prophesize method is deprecated.  Prophecy library is used for prophesize method.


shishir-intelli (approved):
LGTM

PR #756, opened 2022-11-07
Fix error sending email with capitalization to ApigeeX Org

### LINKED ISSUE #730: Error sending email with capitalization to Apigee.
## Description
Our organization is using SSO to log users into the developer portal. This creates a user in Drupal, and our identity provider is sending over the users email address. The email address gets saved with proper capitalization, so I am `David.Pagini@organization.com`. When the module takes this user and create them as a developer in Apigee, we are getting an error that email address must be lower case.

## Expected Behavior
I would expect that a Developer get created on the Apigee side.

## Screenshots
image

## Notes
I guess I would suggest that all users get set/synced to apigee with a lowercase email address.

## Version Info
Drupal core 9.4.5
Apigee_edge 2.0.4
DevPortal_Kickstart 2.1.0

shishir-intelli:
Closes #730 

As ApigeeX Org is not accepting capitalized email address, we are forcing lowercase email addresses for Apigee-X Org to prevent error.


apigee_edge.module
```diff
@@ -1270,6 +1270,36 @@ function template_preprocess_app_credential(array &$variables) {
   }
 }
 
+/**
+ * Forcing user email address to lowercase for ApigeeX/Hybrid Org.
+ * @see https://github.com/apigee/apigee-edge-drupal/issues/730
+ */
+function apigee_edge_user_email_check(&$account) {
```
mxr576, line 1274: Sorry, but I believe this should be a new[ field level constraint](https://www.drupal.org/docs/drupal-apis/entity-api/entity-validation-api/defining-constraints-validations-on-entities-andor-fields) which would run whenever and wherever `$user->validate()` or `$userl->mail->validate()` is called. This would be the standard Drupal/Symfony way or ensuring data consistency.

As a last result, a sanity check could still be executed in `apigee_edge_user_presave()` to ensure the above described field level validation was actually validated before entity save - because we all know that some IDP integration modules loves to bypass validating the user entity before they save it - and if an email still contains uppercase characters a `LogicException` (or a subclass of it) could be thrown to prevent user creation. (Which is still better than spending hours on figuring our why they is a Drupal user in the system without a developer account on the Apigee side.)

apigee_edge.module
```diff
@@ -1270,6 +1270,36 @@ function template_preprocess_app_credential(array &$variables) {
...
+    /* @var \Apigee\Edge\Api\Management\Entity\Organization $organization */
+    $organization = $org_controller->load($sdk_connector->getOrganization());
+
+    // Check if org is ApigeeX.
+    if ($organization && ('CLOUD' === $organization->getRuntimeType() || 'HYBRID' === $organization->getRuntimeType())) {
+       // Apigee X Org is only accepting lower case email address,
+       // therefore preventing the form from creating account
+       // with uppercase email address.
+       if (preg_match('/[A-Z]/', $email)) {
+        $account->set('mail', strtolower($email));
```
mxr576, line 1289: I have security concerns regarding this approach because I am quite sure that somewhere in the user create/update process it is checked whether the new/changed email address of Drupal user is already belongs to a developer account on Apigee, BUT if that runs before this code then somebody could easily take over a developer account of somebody else by using a uppercase version of the email address.

mxr576 (changes requested):

apigee_edge.module
shishir-intelli: Thank you @mxr576 for reviewing this PR.

Do you mean to say, that instead of forcefully converting the email address to lowercase we should put a constraint on the email field so the user gets an error message while user create/update if capitalized email address found for Apigee-X Org.

mxr576: Yes, exactly - And developer sync is also going to fail (it already works like this) when entity validation fails.

apigee_edge.module
sagar-lcm: I know of a case where the client's  LDP was using upper-case characters in every user's email address, which apparently was getting synced with Apigee. The client was reluctant to change the LDP configurations so wrote a hook to convert the new user's email address to lowercase and also an update hook to convert the existing user's email address to lowercase which in turn syncs the developer account Apigee as well.

dpagini:
Just wanted to check in on this issue and see if there's any further movement on this...?

Looking at this PR, I would not be expecting this module to control how I am storing a user email in my Drupal site. I guess, to me, the fix I would expect is simply that, when sending an email over to Apigee, that it's transformed to lowercase at that time. Leave it stored in whatever case it's stored in on Drupal... Is that not viable?

dpagini:
Hi Apigee team - Is there any direction on this pull request? This has been open for more than a year now, with little discussion. There are some valid discussions that are still open here. What are the next steps to move this issue forward?

shishir-intelli:
Hi @dpagini ,
Team is reviewing this PR for any changes and approve it.

kedarkhaire:
Hi @shishir-intelli I checked with other cases also, 
with Okta SAML login also, I checked with camelcase, It works fine and the email is converted to lowercase with message.

kedarkhaire (approved):
LGTM

boobaa:
Please note that this approach goes against the SMTP standard: https://datatracker.ietf.org/doc/html/rfc5321#section-2.4 clearly says:
> The local-part of a mailbox MUST BE treated as case sensitive.

If this gets merged, all kind of breaks are expected, not only in Drupal, but also in receiving emails, etc.

kedarkhaire:
Hi @boobaa 

We are planning to add the following to avoid the duplicate developers with same email ids.

During new user creation - Add a duplicate email id check (Includes a check with lists of existing developers in that X org)
When a existing user is logged in -  Add a duplicate email id check + Add a status report message of system requirements + Restrict users to show the API keys.

If you have any other additional to-dos, we are open for a discussion over here.

mxr576:
Do not forget about email changing in the Drupal UI, that should be also handled... But not even Drupal core handles that (still) https://www.drupal.org/node/85494

Even if think this auto-lowercasing is a bad idea... Let me try to imagine how this could work securely... _Before_ the lowercased email is saved, the ownership must be confirmed via opening a verification link from the mailbox or similar.
Developers are going to have plenty ways to shoot themselves on foot if one system uses upperscase/mixed case emails the other silently converts them to lowercase, just imagine that email stored in the idp gets used for loading a suppose to be existing developer on Apigee.

apigee_edge.module
```diff
@@ -1267,6 +1267,36 @@ function template_preprocess_app_credential(array &$variables) {
...
+        \Drupal::messenger()->addWarning(t('Apigee X Org is only accepting lowercase email address.'));
```
mxr576: So what...? I mean this message is going to be seen by end users. It does not say that they should change the format of their email address....it also does not explain that the system auto fixed that...and as an end user I probably should not know that my API provider is using Apigee X.

dpagini:
I'm not sure I agree with this approach, either. 
I don't think we should be changing the case of an email on the Drupal side. This is very heavy handed. 
Maybe this is not something that should even be fixed in the module, but instead on the Apigee X Org end to accept mixed case email addresses...? Maybe that's a bigger change, but based on some of the links shared above, it seems like this is a problem with the API endpoint, and not with the integration.

apigee_edge.module
shishir-intelli: Updated.

shishir-intelli:
This issue has been resolved here https://github.com/apigee/apigee-edge-drupal/pull/1050

PR #758, opened 2022-11-10
Fix for dev dependencies error

### LINKED ISSUE #757: Github action fails when installing dev dependencies
## Description
Github action fails when installing dev dependencies as it tries to install both behat/mink-browserkit-driver ^2.1 and friends-of-behat/mink-browserkit-driver ^1.4

divya-intelli:
Fixes #757 
"drupal/drupal-extension" was added when moving from travis to circle ci (https://github.com/apigee/apigee-edge-drupal/commit/4020919d08122aaff496036523abef0c8c2aa7d2) also later behat.yml was removed. As currently we are using Github action we don't need this library and can be safely removed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,6 @@
         "bex/behat-screenshot": "^1.2",
         "cweagans/composer-patches": "^1.6",
         "drupal/core-dev": "^9.3",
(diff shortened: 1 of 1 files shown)


shishir-intelli (approved):
LGTM

PR #768, opened 2022-12-20
End of support for PHP 7.4

### LINKED ISSUE #767: End of support for PHP 7.4
End of support for PHP 7.4

## Additional context 
https://www.php.net/supported-versions.php

phdhiren:
Closes #767

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,7 @@
     "type": "drupal-module",
     "description": "Apigee Edge for Drupal.",
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": "^8.0",
         "ext-json": "*",
         "apigee/apigee-client-php": "^2.0.16",
         "drupal/core": "^9.3",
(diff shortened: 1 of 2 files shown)


shishir-intelli (approved):
LGTM

composer.json
```diff
@@ -4,7 +4,7 @@
     "type": "drupal-module",
     "description": "Apigee Edge for Drupal.",
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": "^8.0",
```
dpagini, line 7: Should this really be done as part of a bugfix release? I would say this type of change should be more at least a minor version upgrade, or even a major version.

My project is still (unfortunately) using PHP 7.4, and now I cannot upgrade to the security release published yesterday without upgrading my whole project to PHP8 (which has other dependencies).
Should this be a **hard requirement**?

phdhiren: @dpagini 
PHP 7.4 was not actively supported from Nov'21 and now security updates are also not maintained from Nov'22 and already EOL. 
https://www.php.net/supported-versions.php

PR #770, opened 2022-12-22
add php-8.2 to GH-actions

gosunilgo:
add php-8.2 to GH-actions


mxr576:
Compatibility with PHP 8.2 should be also marked and tested explicitly, see #816

Issue #761, opened 2022-11-17
Drupal 10 compatibility.


dpagini:
## Is your feature request related to a problem? Please describe.
Release a module version that is compatible with Drupal 10.

## Describe the solution you would like
Need to have this module ship a D10/PHP8.1 compatible version before we can update our site to Drupal 10.

## Describe alternatives you have considered
Not using the Apigee suite of modules?

## Additional context 
N/A

VladimirAus:
Running `./vendor/bin/drush -l default -y upgrade_status:analyze apigee_edge`

### Result

```

================================================================================
Apigee Edge,  2.0.6
Scanned on Sat, 07/01/2023 - 19:36

FILE:
modules/contrib/apigee_edge/modules/apigee_edge_teams/src/Controller/TeamMembers
List.php

STATUS         LINE                           MESSAGE                           
--------------------------------------------------------------------------------
Check manually 118  Relying on entity queries to check access by default is     
                    deprecated in drupal:9.2.0 and an error will be thrown from 
                    drupal:10.0.0. Call                                         
                    \Drupal\Core\Entity\Query\QueryInterface::accessCheck() with
                    TRUE or FALSE to specify whether access should be checked.  
--------------------------------------------------------------------------------

FILE:
modules/contrib/apigee_edge/modules/apigee_edge_teams/src/Entity/Form/TeamForm.p
hp

STATUS         LINE                           MESSAGE                           
--------------------------------------------------------------------------------
Check manually 156  Relying on entity queries to check access by default is     
                    deprecated in drupal:9.2.0 and an error will be thrown from 
                    drupal:10.0.0. Call                                         
                    \Drupal\Core\Entity\Query\QueryInterface::accessCheck() with
                    TRUE or FALSE to specify whether access should be checked.  
--------------------------------------------------------------------------------

FILE:
modules/contrib/apigee_edge/modules/apigee_edge_teams/src/Entity/Storage/TeamApp
Storage.php

STATUS         LINE                           MESSAGE                           
--------------------------------------------------------------------------------
Check manually 38   Class Drupal\apigee_edge_teams\Entity\Storage\TeamAppStorage
                    extends @internal class                                     
                    Drupal\apigee_edge\Entity\Storage\AppStorage.               
--------------------------------------------------------------------------------

FILE: modules/contrib/apigee_edge/apigee_edge.libraries.yml

STATUS         LINE                           MESSAGE                           
--------------------------------------------------------------------------------
[...]

PR #782, opened 2023-01-07
#761 Drupal 10 compatibility

VladimirAus:
Fixes all issues mentioned in #761 except `@internal` warning.

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -103,9 +103,9 @@ jobs:
     - name: Install Drupal
       run: |
         cd drupal   
-        php -d sendmail_path=$(which true); vendor/bin/drush --yes -v site-install minimal --db-url="$SIMPLETEST_DB"
-        vendor/bin/drush en apigee_edge -y
-        vendor/bin/drush rs 8000 &
+        php -d sendmail_path=$(which true); vendor/bin/drush --yes -v site:install minimal --db-url="$SIMPLETEST_DB"
+        vendor/bin/drush pm:enable apigee_edge -y
+        vendor/bin/drush runserver 8000 &
 
     - name: "PHPCS"
       run: |
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -86,6 +86,7 @@ jobs:
         composer config --no-plugins allow-plugins.wikimedia/composer-merge-plugin true
         composer config --no-plugins allow-plugins.composer/package-versions-deprecated true
         composer config --no-plugins allow-plugins.cweagans/composer-patches true
+        composer config --no-plugins allow-plugins.php-http/discovery true        
         composer require --dev drush/drush
         composer config minimum-stability dev
         composer require drupal/rules:3.0.0-alpha7
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -29,12 +29,12 @@ jobs:
       fail-fast: false
       matrix:
         php-version:
-          - "8.0"
           - "8.1"
         drupal-core:
           # Should update the following as the minimum supported version from Drupal.org
           - "9.4.x"
           - "9.5.x"
+          - "10.0.x"
 
     steps:
 
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -62,17 +62,17 @@ jobs:
       id: composercache
       run: |
         cd drupal
-        echo "::set-output name=dir::$(composer config cache-files-dir)"
+        echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
 
     - name: Cache composer dependencies
-      uses: actions/cache@v2
+      uses: actions/cache@v3
       with:
         path: ${{ steps.composercache.outputs.dir }}
         key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
         restore-keys: ${{ runner.os }}-composer-
 
     - name: Checkout apigee_edge module
-      uses: actions/checkout@v3.0.0
+      uses: actions/checkout@v3
       with:
         path: drupal/modules/contrib/apigee_edge
 
--- a/apigee_edge.info.yml
+++ b/apigee_edge.info.yml
@@ -3,7 +3,7 @@ description: Apigee Edge Drupal integration.
 package: Apigee
 
 type: module
-core_version_requirement: ^8.7.7 || ^9
+core_version_requirement: ^9.4 || ^10.0
 
 dependencies:
   - drupal:file
@@ -12,8 +12,8 @@ dependencies:
   - drupal:filter
   - drupal:options
   - key:key
-  - drupal:system (>=8.7.0)
+  - drupal:system (>=9.4.0)
 
 configure: apigee_edge.settings
 
-php: "7.1"
+php: "8.0"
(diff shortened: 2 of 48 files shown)


modules/apigee_edge_teams/src/Controller/TeamMembersList.php
```diff
@@ -116,6 +116,7 @@ public function overview(TeamInterface $team) {
     if (!empty($members)) {
       $user_storage = $this->entityTypeManager()->getStorage('user');
       $uids = $user_storage->getQuery()
+        ->accessCheck(FALSE)
```
mxr576, line 121: Personally, I'd require a comment above _every_ `->accessCheck(FALSE)` that explains clearly why the code author thought once that disabling access checking on the query level is necessary and does not lead to an information disclosure vulnerability or worse... because this is what is being risked every time.

apigee_edge.info.yml
```diff
@@ -12,8 +12,8 @@ dependencies:
   - drupal:filter
   - drupal:options
   - key:key
-  - drupal:system (>=8.7.0)
+  - drupal:system (>=9.4.0)
```
mxr576: Do we really need to limit the supported version still?

composer.json
```diff
@@ -4,10 +4,10 @@
     "type": "drupal-module",
     "description": "Apigee Edge for Drupal.",
     "require": {
-        "php": "^8.0",
+        "php": ">=8.0",
```
mxr576: Can you explain why this was changed? They means the same, aren't they?

(We should either be more strict regarding supported PHP versions and do not just believe that our code runs on the next PHP minor version or keep the current wishful constraint. Either way, IMO, that should be decided and change in a separate issue/PR.)

mxr576 (changes requested):

composer.json
mxr576: This is also not aligned with the current test matrix configuration [here](https://github.com/apigee/apigee-edge-drupal/pull/782/files#diff-a73bb6555480a5ee79ae276a3f5d71a08fa316e09a4a8da7b643cf1e92c97df9L32).

What do want to achieve? 
I believe we would like to mark that this version Apigee Edge module only supports `~8.1.0 || ~8.2.0` or even 8.0.x still (for on-prem customers on RHEL?). In that case the strict constraint should be `~8.0.0 || ~8.1.0 || ~8.2.0`.

But again, I'd settle adding/dropping PHP versions in a separate (preliminary?) PR.

apigee_edge.info.yml
VladimirAus: `once` library wouldn't be working with anything earlier than 9.2.
Happy to drop it though.

composer.json
VladimirAus: Again, not critical as Drupal 10 can run only on `8.1+`.
Happy to revert.

modules/apigee_edge_teams/src/Controller/TeamMembersList.php
VladimirAus: Comments added.

composer.json
mxr576: Yep, but Drupal 9 compatibility was not and should not be dropped yet.

apigee_edge.info.yml
mxr576: composer.json guarantees the version, whoever tries to run Drupal without Composer in 2023,must not be lucky person. 😉

composer.json
VladimirAus: It's not dropped. But I'm sure we can rely on php restrictions from the core. Removing php constrain all together.

apigee_edge.info.yml
VladimirAus: You'd be surprised. 
The apigee project I worked on started without the composer until I joined. 😀

.github/workflows/php.yml
```diff
@@ -29,13 +29,12 @@ jobs:
       fail-fast: false
       matrix:
         php-version:
-          - "7.4"
-          - "8.0"
           - "8.1"
```
giteshk, line 32: We are not ready to drop support for 7.4 and 8.0 for Drupal 9.
We are evaluating if we will support 10.x only on 8.1 

We will add 10.x knowing that these tests will fail, till the client SDK is updated to use Symfony 6

```
       include:
          - php-version: "8.1"
            drupal-core: "10.0.x"
          - php-version: "8.1"
            drupal-core: "10.1.x"
        php-version:
          - "7.4"
          - "8.0"
          - "8.1"
        drupal-core:
          # Should update the following as the minimum supported version from Drupal.org
          - "9.4.x"
          - "9.5.x" 
 ```

To make your tests run for 10.x you may have to update the rules module

`composer require 'drupal/rules:3.x-dev@dev'`

File: apigee_edge.info.yml [hunk 5]
```diff
@@ -12,8 +12,7 @@ dependencies:
   - drupal:filter
   - drupal:options
   - key:key
-  - drupal:system (>=8.7.0)
 
 configure: apigee_edge.settings
 
-php: "7.1"
+php: "8.0"
```
giteshk, line 18: Can you add 7.4 as the minimum php version we support.


Past June when Drupal 10.x is ready we will update this to 8.0

composer.json
```diff
@@ -16,17 +16,16 @@
         "apigee/apigee-mock-client-php": "^1.1.1",
         "behat/mink-extension": "^2.0",
         "bex/behat-screenshot": "^1.2",
-        "cweagans/composer-patches": "^1.6",
-        "drupal/core-dev": "^9.4",
-        "drush/drush": "^9.0 || ^10.0",
+        "cweagans/composer-patches": "^1.7",
+        "drupal/core-dev": "^9.4 || ^10.0",
+        "drupal/drupal-extension": "master-dev",
```
giteshk, line 21: I could test this with the following version: 
```
"drupal/drupal-extension": "^4.2.1 || ~5",
```

Also removed 
```
        "behat/mink-extension": "^2.0",
        "bex/behat-screenshot": "^1.2",
```

composer.json
```diff
@@ -4,10 +4,10 @@
     "type": "drupal-module",
     "description": "Apigee Edge for Drupal.",
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": ">=8.1",
```
giteshk, line 7: Lets leave 7.4 support till we are ready to deprecate it.

giteshk:
@VladimirAus,

I have tested the PR in my repository and have noted some of my observations.
Code rebased and minor modifications => https://github.com/giteshk-org/apigee-edge-drupal/tree/pr-782
Results of test run => https://github.com/giteshk-org/apigee-edge-drupal/actions/runs/4706077631/jobs/8347171374

We should definitely add support for Drupal 10.x and PHP 8.1 testing. 

We will make sure the branch protection rules are updated to not require, Drupal 10.x  tests to pass till we finish 10.x compatibility issues.

Thank you for your contribution.  After the tests pass we can definitely merge it in.

@mxr576 thank you for the review.

giteshk (changes requested):

mxr576:
@VladimirAus Are you planning to address remaining todos here and what is the plan to move forward? 

I assume myself and our team cannot push changes to your issue branch, that is why I am asking.

*Update*: hm, I have just found this suggestion about splitting this monster PR to smaller ones, which makes sense to me. Based on our D10 compatibility fixes on other projects, there must be some changes that can be introduced in atomic commits/PRs. What do you think?

https://github.com/apigee/apigee-edge-drupal/issues/783#issuecomment-1381452532

divya-intelli:
Closing this issue as all the changes have been added in 3.x.

PR #790, opened 2023-02-01
Support for entity query access check

shishir-intelli:
Fix api_product entity query access issue by adding support for entity query access check.


phdhiren (approved):
LGTM

PR #803, opened 2023-02-28
Checking if the organisation is ApigeeX for AppGroup

shishir-intelli:
Checking if the organisation is ApigeeX for AppGroup.


phdhiren (approved):
LGTM

PR #806, opened 2023-02-28
Fix organization exception

shishir-intelli:
Fix organization exception by returning false for ApigeeX org.


divya-intelli (approved):
LGTM

PR #808, opened 2023-03-03
Warrant that queues executed by cron

### LINKED ISSUE #802: Background developer sync queues are not executed via CRON
## Description
Setting up background developer sync and executing CRON doesn't process the apigee_edge_jobs queue items.

## Apigee Info
Apigee EDGE, although it seems like limited only to Drupal.

## Steps to Reproduce
1. Install a Drupal site with the apigee module.
2. Enable and setup the apigee edge module.
3. Make sure you have a few developers to sync on apigee.
4. Go to the "/admin/config/apigee-edge/developer-settings/sync" page and press the "Background developer sync" button.
5. Executing `drush queue-list` shows that there is one item in the queue.
6. Execute CRON via `drush cron` or on the UI (/admin/config/system/cron).
7. Check with `drush queue-list` again, now there are two items in the queue. (If you pressed the "Background developer sync" button multiple times, the number of newly added items will be duplicated).
8. Execute CRON again and check with `drush queue-list`, the queue items are not processed.

## Actual Behavior
Apigee edge job queue items are not getting processed.

## Expected Behavior
Apigee edge job queue items should be processed during CRON.

## Notes
It seems like that one of the problems that the `@QueueWorker` annotation is missing the cron key, without that Drupal won't execute the queue worker: \Drupal\Core\Cron::processQueues():172

However adding the `cron = {"time" = 60}` to the queue worker there seems to be another problem after running CRON, the developers are still not being synced, but the queue items are processed like this with no result.

## Version Info
apigee/apigee-client-php: 2.0.17
drupal/core-recommended: 9.5.3

mxr576:
Closes #802 

By using the default processing time limit.

### DIFF
--- a/src/Plugin/QueueWorker/JobQueueWorker.php
+++ b/src/Plugin/QueueWorker/JobQueueWorker.php
@@ -32,6 +32,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  * @QueueWorker(
  *   id = "apigee_edge_job",
  *   title = "Apigee Edge job runner",
+ *   cron = {"time" = \Drupal\Core\Queue\QueueWorkerManagerInterface::DEFAULT_QUEUE_CRON_TIME}
  * )
  */
 class JobQueueWorker extends QueueWorkerBase implements ContainerFactoryPluginInterface {


shishir-intelli (approved):
LGTM

@mxr576 ,
Thank you for your contribution.

PR #809, opened 2023-03-03
App creation fails without API Products

### LINKED ISSUE #766: (Developer) app creation fails with keymanagement.service.ApiProductsRequired
## Description
This is another issue that we have bumped into with the same customer as in #764. However in this case it is 99% sure that the issue is caused by the `features.keymanagement.disable.unbounded.permissions` feature flag set TRUE on the Apigee org. One of the things that this flag change is:

> When creating a developer or company app, the management API requires that the app be associated with an API product. (The management UI already enforces this.)

(source: https://docs.apigee.com/release/notes/180202-apigee-edge-public-cloud-release-notes)

This feature flag is a 2018 invention, yet (AFAIK) nobody has ever reported that when that feature flag is enabled then the way how this Drupal module creates an app does not work anymore. 

This module first tries to create an "empty app"  without and after assigning a credential (with selected API products) to the created app. With this feature flag enabled it MUST do this in one API call, which is technically possible on the Apigee side.

See: https://github.com/apigee/apigee-edge-drupal/blob/2.0.6/src/Entity/Form/AppForm.php#L133-L175

## Apigee Info
Apigee Edge

## Steps to Reproduce

Have the given feature flag enabled on an organization and create a developer app.

## Actual Behavior

App creation fails with an error message and the following in the logs

```
{
    "code": "keymanagement.service.ApiProductsRequired",
    "message": "API products required",
    "contexts": []
}

```

## Expected Behavior

The app is created.

## Version Info
2.0.16

phdhiren:
caused by features.keymanagement.disable.unbounded.permissions flag

Fixes #766

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -95,7 +95,6 @@ jobs:
         composer config --json extra.merge-plugin.require '["modules/contrib/apigee_edge/composer.json"]'
         composer config platform.php ${{ matrix.php-version }}
         composer config --json extra.patches."drupal/core" '{ "Support entities that are neither content nor config entities": "https://www.drupal.org/files/issues/2020-12-02/3042467-50.patch"}'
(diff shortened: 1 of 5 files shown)


shishir-intelli (approved):
LGTM

PR #815, opened 2023-03-08
Sort API products on app forms by API product display name

### LINKED ISSUE #811: Sort API Products
We have installed Drupal 9 Apigee Kickstart  with connection to a  Apigee Edge Private Cloud instance.
Is there a way to  sort the list of API Products that are displayed in the  "Add App" form?

mxr576:
This PR rolls back the change that was made in #812 and properly fixes #811.

End users do not want to see API products sorted by their `machine name` rather by their `display name` (which is their customer facing naming).

This implementation also ensures that sorting happens where it is actually needed.

Before:

image
image

After:

image
image

The difference is noticeable but I'd like to explain why `Zero product` was on the top before... because its machine name was `An API product`.

### DIFF
--- a/modules/apigee_edge_teams/src/Entity/Form/TeamAppFormTrait.php
+++ b/modules/apigee_edge_teams/src/Entity/Form/TeamAppFormTrait.php
@@ -221,10 +221,8 @@ trait TeamAppFormTrait {
         return $this->getTeamMemberApiProductAccessHandler()->access($api_product, 'assign', $team);
       };
     }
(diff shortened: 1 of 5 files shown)


giteshk (approved):

giteshk:
LGTM

giteshk:
@mxr576 , Can you take a look at why this test is failing?
https://github.com/apigee/apigee-edge-drupal/actions/runs/4405851484

PR #825, opened 2023-03-23
Support for AppGroup membership

shishir-intelli:
Added support for AppGroup membership which stores the team members locally.

TODO : Save membership information on AppGroup's `_apigee_reserved__memberships` custom attribute.

### DIFF
--- a/src/Entity/DeveloperCompaniesCacheInterface.php
+++ b/src/Entity/DeveloperCompaniesCacheInterface.php
@@ -42,6 +42,25 @@ interface DeveloperCompaniesCacheInterface {
   public function getCompanies(string $id): ?array;
 
   /**
+   * Returns appgroups of a developer.
+   *
+   * @param string $id
+   *   Developer id.
+   *
+   * @return string[]|null
+   *   Array of appgroup names or NULL if information is not yet available.
+   */
+  public function getAppGroups(string $id): ?array;
+
+  /**
+   * Saves developers' appgroups to cache.
+   *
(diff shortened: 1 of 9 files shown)


src/Entity/DeveloperCompaniesCacheInterface.php
```diff
@@ -41,6 +41,25 @@ interface DeveloperCompaniesCacheInterface {
    */
   public function getCompanies(string $id): ?array;
 
+  /**
```
giteshk, line 44: While this may work. I am not sure if this is the best approach.

We should create AppGroup Entities here.

giteshk (approved):
Please add a ticket or TODO to fix 

DeveloperCompaniesCacheInterface.php

src/Entity/DeveloperCompaniesCacheInterface.php
shishir-intelli: The above function only works for Edge and not for ApigeeX, for ApigeeX we have `getAppGroup()` function however  AppGroup caching has not yet implemented which is mentioned [here](https://github.com/apigee/apigee-edge-drupal/pull/825/files#diff-dce6159a29b9c2d2be06ba9fc3c7f60810512762696d33c4c14d20cee4b5c44fR282:~:text=it%20to%20the-,local,-cache%20so%20we) as TODO.

I will create a separate ticket for creating `AppGroup` entity for Edge module.

shishir-intelli: Created ticket https://github.com/apigee/apigee-edge-drupal/issues/830 for above issue

PR #840, opened 2023-05-24
Fix issue which removes users from all team if user is removed from one team

### LINKED ISSUE #822: Removing user from one team removes the same user from all the teams
## Description
If a user has an administrative access to multiple teams, and that user is removed from one team that they are member of, the same user will be removed from all the teams that they are member of.

## Apigee Info
Apigee Edge

## Steps to Reproduce

**Sign in with a non-super-admin user**
1. Go to '/teams' page
2. Click on 'Add team' button
3. Create 3 or 4 teams, just to make sure that this is a valid bug, to see that the same user loses access to all teams, and not just one or two teams.
    - Teams created:
        - teams_test_1
        - teams_test_2
        - teams_test_3
        - teams_test_4

5. In the 'Operations' column there are two options: (1) 'Teams apps' and (2) 'Members' - click on 'Members' option from operations on one of the newly created teams
6. You will end up at the '/teams/{your_team_name}/members' URL: find your account and from the 'Operations' section click on 'Remove' and 'Confirm' and delete your account from that team

    - I have deleted the user from 'teams_test_4'

7. You will land on an error page, indicating that you have lost your access to that team
8. Navigate to the '/teams' page again, where you will see that from the 'Operations' column the option 'Members' is gone for all the teams (that were created in step 3)
9. To double check that the access to the specific teams has gone, you can go to the '/teams/{your_team_name}/members' URL and you will get a 403 error that shows that you have lost access to that team

**Sign in with super-admin user**

1. Go to '/teams/teams_test_1/members': you will see the user there
2. Go to '/teams/teams_test_2/members': you will see the user there
3. Go to '/teams/teams_test_3/members': you will see the user there
4. Go to '/teams/teams_test_4/members': there are no users here, as the only user was deleted (step 6 above)

## Actual Behavior
If a user who has access to team administration, and this user is added to multiple teams, and this user is removed from one of the teams, this user will loose access to all the teams that they were a member of.

With super-admin user, however, it can be seen that the user was removed only from one team, and in further teams that same user is listed as a member.

## Expected Behavior
Leaving one team should mean that the user loses access to that team and that team only.

## Version Info
apigee_edge_teams 2.0.8

shishir-intelli:
Closes #822 
Fix issue - When deleting one users from one Team and if that user is added on multiple teams then that user gets deleted from all Teams.

### DIFF
--- a/modules/apigee_edge_teams/src/TeamMembershipManager.php
+++ b/modules/apigee_edge_teams/src/TeamMembershipManager.php
@@ -142,18 +142,22 @@ final class TeamMembershipManager implements TeamMembershipManagerInterface {
       $controller->removeMember($developer);
       // Remove team member's roles from Drupal.
       if (array_key_exists($developer, $users_by_mail)) {
-        /** @var \Drupal\apigee_edge_teams\Entity\TeamMemberRoleInterface[] $team_member_roles_in_teams */
-        $team_member_roles_in_teams = $team_member_role_storage->loadByDeveloper($users_by_mail[$developer]);
-        foreach ($team_member_roles_in_teams as $team_member_roles_in_team) {
-          try {
+        /** @var \Drupal\user\Entity\User $account */
+        $account = user_load_by_mail($users_by_mail[$developer]->getEmail());
+        $team_entity = $this->entityTypeManager->getStorage('team')->load($team);
+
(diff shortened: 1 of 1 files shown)


divya-intelli (approved):
LGTM

PR #853, opened 2023-06-06
Implemented AppGroup Team Membership Sync

shishir-intelli:
Team Membership sync,
Durning Team member sync all the member details (email id and roles) would be stored in database from `__apigee_reserved__developer_details` attribute.

If developer email doesn't exists in Drupal portal, a invitation will be sent to those developer and the record will be shown in `Pending Invitations` tab.


divya-intelli (approved):
LGTM

PR #884, opened 2023-07-11
Expiry date hence changed to now and developer renamed to members

shishir-intelli:
- App key expiry date format 'hence' replaced with 'now'. 
- In invite team member page, the title `Developers` replaced with `Email address(es)` as the members of an Apigee edge/X are not necessarily `developers`.
- 
b/289746352


divya-intelli (approved):
LGTM

PR #885, opened 2023-07-11
Post create team redirection and changed internal name as machine name

shishir-intelli:
- Changed `Internal name` as `Machine name` 
- Following the team creation, user is redirected to manage the members and apps for that team.

b/289746352


divya-intelli (approved):
LGTM

PR #848, opened 2023-06-02
Tweak TeamMemberRole.php so that it does not cause errors with the JS…

audave:
…ONAPI
https://github.com/apigee/apigee-edge-drupal/issues/845

PR #887, opened 2023-07-25
Fix team member role install schema error due to duplicate UUID

shishir-intelli:
Fix for the issue occurred because of the PR https://github.com/apigee/apigee-edge-drupal/pull/848,

In PR https://github.com/apigee/apigee-edge-drupal/pull/848, after adding UUID in `entity_keys` which was required for JSONAPI for `team_member_role`, the schema was showing duplicate UUID field data, which caused error while installing `team_member_role` table. 

Fixes error:
`Drupal\Core\Database\SchemaException: The 'uuid' field specification does not define 'not null' as TRUE.`


divya-intelli (approved):
LGTM

PR #919, opened 2023-08-28
Extract user deletion into a new service

### LINKED ISSUE #915: Provide a programming API for opting out/fine tuning developer removal on user removal
## Is your feature request related to a problem?

The Apigee Edge module [nicely integrates with the user cancellation API](https://github.com/apigee/apigee-edge-drupal/blob/3.0.1/apigee_edge.module#L1420-L1487) in Drupal and it follows the development workflow that [the API enforces on Drupal developers](https://github.com/drupal/core/blob/10.1.2/modules/user/user.api.php#L25-L27), more specifically, it implements `hook_user_delete()` for handling user removal instead of reacting on `user_cancel_delete` cancel method. 

This API provides an ideal approach for canceling/removing user account on the Drupal HTML UI, however, user removal can happen programmatically and when that happens Drupal developers may not always want to remove developer accounts from Apigee as well (and that destroy their API keys). (When the Apigee Monetization feature is enabled, a developer cannot be removed in given circumstances, the Monetization API does not allow that.)

The problem with this approach (again enforced by the user cancellation API) that without additional tweaks there is no way to opt out or tweak how the developer removal should happen when a user is deleted. Should it be executed synchronously (ideal for removing one user) or asynchronously in batch (ideal for bulk removal)?

Extract the current implementation of `apigee_edge_user_delete()` to a new service and call that service inside the method. Just simply introducing a service (class+interface) should give enough ammunition for downstream Drupal developers via the service decorator pattern to tweak if and how the developer removal happens when a user is deleted.

mxr576:
Introduce a programming API for downstream developers that allows customizing what should happen when a user is deleted in Drupal.

Co-authored-by: @FCsongradi 

Closes #915

### DIFF
--- a/modules/apigee_edge_teams/src/User/RemoveTeamRolesWithUserSynchronousUserRemovalHandler.php
+++ b/modules/apigee_edge_teams/src/User/RemoveTeamRolesWithUserSynchronousUserRemovalHandler.php
@@ -1,0 +1,64 @@
+<?php
+
+/**
+ * Copyright 2023 Google Inc.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * version 2 as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+namespace Drupal\apigee_edge_teams\User;
+
+use Drupal\apigee_edge\User\UserRemovalHandlerInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Utility\Error;
+use Drupal\user\UserInterface;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
+
+/**
+ * Ensures team roles of the removed user also get deleted.
+ */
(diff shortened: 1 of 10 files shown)


mxr576:
Just spotted `apigee_edge_teams_user_delete()` :O so that should be also moved to use the new API.

cc @FCsongradi

src/User/PostUserDeleteActionPerformerInterface.php
```diff
@@ -23,18 +23,22 @@
...
 /**
- * Contract for triggering an Apigee specific reaction on user removal.
+ * Contract for triggering Apigee specific reactions _after_ user removal.
+ *
+ * It is important to know that when implementations are called the user
+ * entity is already removed and the database transaction is closed, there
+ * is no way to roll that back.
  */
-interface UserRemovalHandlerInterface {
+interface PostUserDeleteActionPerformerInterface {
```
mxr576, line 32: Is is better? I mean, naming is hard... but I think it is better to be explicit and descriptive in names (in domain driven style) when designing public APIs. If we like this I can squash this change to other commits.

modules/apigee_edge_teams/src/User/RemoveTeamRolesWithUserSynchronousUserRemovalHandler.php
```diff
@@ -0,0 +1,64 @@
...
+  /**
+   * RemoveTeamRolesWithUserSynchronousUserRemovalHandler constructor.
+   */
+  public function __construct(private readonly UserRemovalHandlerInterface $decorated, private readonly EntityTypeManagerInterface $entityTypeManager, private readonly LoggerInterface $logger) {}
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __invoke(UserInterface $account): void {
+    ($this->decorated)($account);
```
mxr576: Double checked and this was the original order as well, the hook in apigee_edge was called sooner than in this module - which makes total sense.

shishir-intelli (approved):
Checked for Apigee Edge and X both.
LGTM

PR #926, opened 2023-09-05
Added changes for AppGroup Apps pagination

shishir-intelli:
Implemented pagination for Apigee-X AppGroup Apps.
Fix issue showing 1000+ apps for Apigee X.
Closes https://github.com/apigee/apigee-client-php/issues/305

### DIFF
--- a/modules/apigee_edge_teams/src/Entity/Controller/TeamAppController.php
+++ b/modules/apigee_edge_teams/src/Entity/Controller/TeamAppController.php
@@ -20,6 +20,7 @@
 
 namespace Drupal\apigee_edge_teams\Entity\Controller;
 
+use Apigee\Edge\Api\ApigeeX\Controller\AppByOwnerControllerInterface as XAppByOwnerControllerInterface;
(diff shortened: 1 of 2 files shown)


modules/apigee_edge_teams/src/Entity/Controller/TeamAppController.php
```diff
@@ -20,6 +20,7 @@
 
 namespace Drupal\apigee_edge_teams\Entity\Controller;
 
+use Apigee\Edge\Api\ApigeeX\Controller\AppByOwnerControllerInterface as XAppByOwnerControllerInterface;
```
giteshk, line 23: can we call this AppByOwnerControllerInterface and drop the X

shishir-intelli: We already have `AppByOwnerControllerInterface as EdgeAppByOwnerControllerInterface;` (line 25) for Edge, to prevent ambiguity we have Apigee X file name as XAppByOwnerControllerInterface

divya-intelli (approved):
LGTM

PR #929, opened 2023-09-08
Fix isOrganizationApigeeX() on null issue

### LINKED ISSUE #928: Saving multivalue field that is part of the Team Entity results in an error.
## Description
Saving multivalue field that is part of the Team Entity results in an error.

## Apigee Info
Apigee Edge

## Steps to Reproduce
Enable apigee_edge_teams
Add a new field that enables having multiple values /admin/config/apigee-edge/team-settings/fields
Add a new team add with least 2 values into our new field.

## Expected behaviour
Team will save.
Field values will be save

## Actual behaviour
Hit Save, will result in a 500 error.

Error message in logs is
`Error: Call to a member function isOrganizationApigeeX() on null in Drupal\apigee_edge_teams\Entity\Form\TeamForm->buildEntity() (line 165 of /app/web/modules/contrib/apigee_edge/modules/apigee_edge_teams/src/Entity/Form/TeamForm.php).`

Replicated using Apigee Kickstart distro

shishir-intelli:
Closes #928

### DIFF
--- a/modules/apigee_edge_teams/src/Entity/Form/TeamForm.php
+++ b/modules/apigee_edge_teams/src/Entity/Form/TeamForm.php
@@ -159,6 +159,9 @@ class TeamForm extends FieldableEdgeEntityForm implements EdgeEntityFormInterfac
     /** @var \Drupal\apigee_edge_teams\Entity\TeamInterface $team */
     $team = parent::buildEntity($form, $form_state);
 
+    if (NULL === $this->orgController) {
+      $this->orgController = \Drupal::service('apigee_edge.controller.organization');
+    }
     // ADMIN_EMAIL_ATTRIBUTE is a required field for monetization.
     // We add to any team to make sure team creation works for mint orgs even
     // if they do not enable the m10n teams module.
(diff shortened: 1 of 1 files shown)


modules/apigee_edge_teams/src/Entity/Form/TeamForm.php
```diff
@@ -159,6 +159,9 @@ public function buildEntity(array $form, FormStateInterface $form_state) {
     /** @var \Drupal\apigee_edge_teams\Entity\TeamInterface $team */
     $team = parent::buildEntity($form, $form_state);
 
+    if (NULL === $this->orgController) {
+      $this->orgController = \Drupal::service('apigee_edge.controller.organization');
```
mxr576: This fixes the null problem, but how this can be NULL when it is DI-d above?

shishir-intelli: This issue is coming in 3x only, when added new Team field with multi-value field setting.

mxr576: I saw the report, but by reading the code in this class I still do not see why/how this can be NULL, when the service is injected above. 
Depending on `\Drupal::service(()` is classes only acceptable as a BC bridge that gets removed in the next major version. Is this going to be removed or have to stay?

shishir-intelli: This issue was caused by `$orgController` private access specifier.
Resolved

divya-intelli (approved):
LGTM

modules/apigee_edge_teams/src/Entity/Form/TeamForm.php
mxr576: Ahh so the actual issue was related to https://www.drupal.org/project/drupal/issues/3110266

PR #937, opened 2023-09-20
Bump Client PHP requirement version

shishir-intelli:
Bump Client PHP requirement version to 2.1.1 to support Appgroup paginations.


divya-intelli (approved):
LGTM

PR #940, opened 2023-09-25
Resolved the validation for monitized enabled org for enabling teams for ApigeeX and edge

### LINKED ISSUE #936: Apigee X: Monetization vs Teams requirement check is a dead code (?)
## Description
https://github.com/apigee/apigee-edge-drupal/blob/8466a97f3cacf4224b473b3b8d10891476cbb5f6/modules/apigee_edge_teams/apigee_edge_teams.install#L44-L57

## Apigee Info
Apigee X with Monetization enabled.

## Steps to Reproduce
Enable the Apigee Edge Teams module on a Monetization enabled Apigee X org.

## Actual Behavior
The module was enabled with no requirement error.

## Expected Behavior
Well the Teams functionality is available on Apigee X as AppGroups but based on our discussion the AppGroup+Monetization integration missing at this moment, so probably the **fixed** requirement check should be added to https://github.com/apigee/apigee-m10n-drupal/blob/2.1.0/modules/apigee_m10n_teams/apigee_m10n_teams.install

## Version Info
3.0.1

shishir-intelli:
Closes #936 
Resolved the dead code for validating for Apigee X with Monetization enabled org while enabling Teams module, 
also added error message on all Teams related pages if the module is already enabled (_In case organisation switched from Edge to X_).

### DIFF
--- a/modules/apigee_edge_teams/apigee_edge_teams.install
+++ b/modules/apigee_edge_teams/apigee_edge_teams.install
@@ -43,7 +43,7 @@ function apigee_edge_teams_requirements($phase) {
       $organization = $org_controller->load($sdk_connector->getOrganization());
       if ($organization && $org_controller->isOrganizationApigeeX()) {
         // AppGroup APIs are not supported in Apigee X / Hybrid orgs with monetization enabled.
-        if ($organization->getAddonsConfig() || TRUE === $organization->getAddonsConfig()->getMonetizationConfig()->getEnabled()) {
+        if ($organization->getAddonsConfig() && $organization->getAddonsConfig()->getMonetizationConfig() && TRUE === $organization->getAddonsConfig()->getMonetizationConfig()->getEnabled()) {
           $url = [
             ':url' => 'https://cloud.google.com/apigee/docs/api-platform/publish/organizing-client-app-ownership?hl=en#appgroups-limitations-and-known-issues',
           ];
--- a/modules/apigee_edge_teams/apigee_edge_teams.install
+++ b/modules/apigee_edge_teams/apigee_edge_teams.install
@@ -41,18 +41,36 @@ function apigee_edge_teams_requirements($phase) {
       $org_controller = \Drupal::service('apigee_edge.controller.organization');
       /* @var \Apigee\Edge\Api\Management\Entity\Organization $organization */
       $organization = $org_controller->load($sdk_connector->getOrganization());
-      if ($organization && !OrganizationFeatures::isCompaniesFeatureAvailable($organization) && OrganizationFeatures::isMonetizationEnabled($organization)) {
-        $url = [
-          ':url' => 'https://cloud.google.com/apigee/docs/api-platform/get-started/compare-apigee-products#unsupported-apis',
-        ];
-        $message = ($phase == 'runtime') ?
-          t("The Teams module functionality is not available for monetization enabled org on Apigee X / Hybrid and should be uninstalled, because <a href=':url' target='_blank'>AppGroup APIs are not supported in Apigee X / Hybrid orgs with monetization enabled</a>.", $url) :
-          t("The Teams module functionality is not available for monetization enabled org on Apigee X / Hybrid because <a href=':url' target='_blank'>AppGroup APIs are not supported in Apigee X / Hybrid orgs with monetization enabled</a>.", $url);
-        $requirements['apigee_edge_teams_not_supported'] = [
-          'title' => t('Apigee Edge Teams'),
-          'description' => $message,
-          'severity' => REQUIREMENT_ERROR,
-        ];
+      if ($organization && $org_controller->isOrganizationApigeeX()) {
+        // AppGroup APIs are not supported in Apigee X / Hybrid orgs with monetization enabled.
+        if ($organization->getAddonsConfig() || TRUE === $organization->getAddonsConfig()->getMonetizationConfig()->getEnabled()) {
+          $url = [
+            ':url' => 'https://cloud.google.com/apigee/docs/api-platform/get-started/compare-apigee-products#unsupported-apis',
+          ];
+          $message = ($phase == 'runtime') ?
+            t("The Teams module functionality is not available for monetization enabled org on Apigee X / Hybrid and should be uninstalled, because <a href=':url' target='_blank'>AppGroup APIs are not supported in Apigee X / Hybrid orgs with monetization enabled</a>.", $url) :
+            t("The Teams module functionality is not available for monetization enabled org on Apigee X / Hybrid because <a href=':url' target='_blank'>AppGroup APIs are not supported in Apigee X / Hybrid orgs with monetization enabled</a>.", $url);
+          $requirements['apigee_edge_teams_not_supported'] = [
+            'title' => t('Apigee Edge Teams'),
+            'description' => $message,
+            'severity' => REQUIREMENT_ERROR,
+          ];
+        }
+      } else {
+        // Edge company APIs are not supported in Apigee Hybrid orgs.
+        if ($organization && !OrganizationFeatures::isCompaniesFeatureAvailable($organization)) {
+          $url = [
+            ':url' => 'https://cloud.google.com/apigee/docs/api-platform/get-started/compare-apigee-products#unsupported-apis',
+          ];
+          $message = ($phase == 'runtime') ?
+            t("The Apigee Edge Teams module functionality is not available for your org and should be uninstalled, because <a href=':url' target='_blank'>Edge company APIs are not supported for Apigee Hybrid orgs</a>.", $url) :
+            t("The Apigee Edge Teams module functionality is not available for your org because <a href=':url' target='_blank'>Edge company APIs are not supported for Apigee Hybrid orgs</a>.", $url);
+          $requirements['apigee_edge_teams_not_supported'] = [
+            'title' => t('Apigee Edge Teams'),
+            'description' => $message,
+            'severity' => REQUIREMENT_ERROR,
+          ];
+        }
       }
     }
     catch (\Exception $exception) {
--- a/modules/apigee_edge_teams/apigee_edge_teams.install
+++ b/modules/apigee_edge_teams/apigee_edge_teams.install
@@ -45,7 +45,7 @@ function apigee_edge_teams_requirements($phase) {
         // AppGroup APIs are not supported in Apigee X / Hybrid orgs with monetization enabled.
         if ($organization->getAddonsConfig() || TRUE === $organization->getAddonsConfig()->getMonetizationConfig()->getEnabled()) {
           $url = [
-            ':url' => 'https://cloud.google.com/apigee/docs/api-platform/get-started/compare-apigee-products#unsupported-apis',
+            ':url' => 'https://cloud.google.com/apigee/docs/api-platform/publish/organizing-client-app-ownership?hl=en#appgroups-limitations-and-known-issues',
           ];
           $message = ($phase == 'runtime') ?
             t("The Teams module functionality is not available for monetization enabled org on Apigee X / Hybrid and should be uninstalled, because <a href=':url' target='_blank'>AppGroup APIs are not supported in Apigee X / Hybrid orgs with monetization enabled</a>.", $url) :
--- a/modules/apigee_edge_teams/apigee_edge_teams.install
+++ b/modules/apigee_edge_teams/apigee_edge_teams.install
@@ -56,21 +56,6 @@ function apigee_edge_teams_requirements($phase) {
             'severity' => REQUIREMENT_ERROR,
           ];
         }
-      } else {
-        // Edge company APIs are not supported in Apigee Hybrid orgs.
-        if ($organization && !OrganizationFeatures::isCompaniesFeatureAvailable($organization)) {
-          $url = [
-            ':url' => 'https://cloud.google.com/apigee/docs/api-platform/get-started/compare-apigee-products#unsupported-apis',
-          ];
-          $message = ($phase == 'runtime') ?
-            t("The Apigee Edge Teams module functionality is not available for your org and should be uninstalled, because <a href=':url' target='_blank'>Edge company APIs are not supported for Apigee Hybrid orgs</a>.", $url) :
-            t("The Apigee Edge Teams module functionality is not available for your org because <a href=':url' target='_blank'>Edge company APIs are not supported for Apigee Hybrid orgs</a>.", $url);
-          $requirements['apigee_edge_teams_not_supported'] = [
-            'title' => t('Apigee Edge Teams'),
-            'description' => $message,
-            'severity' => REQUIREMENT_ERROR,
-          ];
-        }
       }
     }
     catch (\Exception $exception) {
--- a/modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
+++ b/modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
@@ -1,0 +1,97 @@
+<?php
+
+/*
+ * Copyright 2023 Google Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License version 2 as published by the
+ * Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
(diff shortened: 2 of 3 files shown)


modules/apigee_edge_teams/apigee_edge_teams.install
```diff
@@ -41,18 +41,36 @@ function apigee_edge_teams_requirements($phase) {
...
-          'severity' => REQUIREMENT_ERROR,
-        ];
+      if ($organization && $org_controller->isOrganizationApigeeX()) {
+        // AppGroup APIs are not supported in Apigee X / Hybrid orgs with monetization enabled.
+        if ($organization->getAddonsConfig() || TRUE === $organization->getAddonsConfig()->getMonetizationConfig()->getEnabled()) {
+          $url = [
+            ':url' => 'https://cloud.google.com/apigee/docs/api-platform/get-started/compare-apigee-products#unsupported-apis',
+          ];
+          $message = ($phase == 'runtime') ?
+            t("The Teams module functionality is not available for monetization enabled org on Apigee X / Hybrid and should be uninstalled, because <a href=':url' target='_blank'>AppGroup APIs are not supported in Apigee X / Hybrid orgs with monetization enabled</a>.", $url) :
```
mxr576, line 51: I am really uncertain if this is actually the case as I wrote in the original issue. 

I have a monetization enabled trial Apigee X organization and the teams functionality is working without any issues. What should be missing is the Teams (AppGroups) and Monetization integration. This was the reason why I suggested to move this check to the Monetization Teams module. 

Can you please double check if there is any issue with the **base** Teams functionality on a Monetized Apigee X org?

modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
```diff
@@ -0,0 +1,97 @@
...
+ * Validates that Apigee X Team is enabled on every Team page request.
+ */
+class ValidateApigeeXTeamEnabledSubscriber implements EventSubscriberInterface {
+
+  /**
+   * The SDK connector service.
+   *
+   * @var \Drupal\apigee_edge\SDKConnectorInterface
+   */
+  private $connector;
```
mxr576: Suggestion: use native typehints by default.

modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
```diff
@@ -0,0 +1,97 @@
...
+use Drupal\apigee_edge\SDKConnectorInterface;
+use Drupal\Core\Messenger\MessengerInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\RequestEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * Validates that Apigee X Team is enabled on every Team page request.
+ */
+class ValidateApigeeXTeamEnabledSubscriber implements EventSubscriberInterface {
```
mxr576: Suggestion: mark class `final`.

modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+   * If monetization enabled in Apigee X org alert the user.
+   *
+   * @param \Symfony\Component\HttpKernel\Event\RequestEvent $event
+   *   The event.
+   */
+  public function validateApigeexTeamEnabled(RequestEvent $event) {
+    /** @var \Symfony\Component\Routing\Route $current_route */
+    if (($current_route = $event->getRequest()->get('_route')) && (strpos($current_route, 'team') !== FALSE)) {
```
mxr576: >  (strpos($current_route, 'team') !== FALSE)

This looks way to invasive to me, I'd be as specific as possible, like starts with `apigee_edge_team` or `entity.team`...

mxr576: Should only react on HTML requests, no other, like JSON via JSONAPI.

modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
```diff
@@ -0,0 +1,97 @@
...
+      $organization = $this->orgController->load($this->connector->getOrganization());
+      if ($organization && $this->orgController->isOrganizationApigeeX()) {
+        if ($organization->getAddonsConfig() || TRUE === $organization->getAddonsConfig()->getMonetizationConfig()->getEnabled()) {
+          $this->messenger->addError('The Teams module functionality is not available for monetization enabled org on Apigee X / Hybrid and should be uninstalled');
```
mxr576: A sentence should end with full stop.

mxr576 (changes requested):

modules/apigee_edge_teams/apigee_edge_teams.install
shishir-intelli: > I have a monetization enabled trial Apigee X organization and the teams functionality is working without any issues.

Ideally this shouldn't work as of now, because it is not completely supported in Apigee X (_but due the above code validation issue it got bypassed_). 
So basically for now we are blocking the teams module installing/enabling in Drupal with the rectified code `PR 940` and if in case, the Teams module is already enabled then a error message will be shown on all teams related pages (_ex: teams, team/{team_name}, teams settings etc_), rest will work as it is.
Added same code in M10n teams module as well.

@giteshk Can you please confirm.

giteshk: Thank you @shishir-intelli. I agree this would be the right approach till we have appgroup support in monetization.

modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+   * If monetization enabled in Apigee X org alert the user.
+   *
+   * @param \Symfony\Component\HttpKernel\Event\RequestEvent $event
+   *   The event.
+   */
+  public function validateApigeexTeamEnabled(RequestEvent $event) {
+    /** @var \Symfony\Component\Routing\Route $current_route */
+    if (($current_route = $event->getRequest()->get('_route')) && (strpos($current_route, 'entity.team') !== FALSE || strpos($current_route, 'settings.team') !== FALSE)) {
```
mxr576: It just occurred to me, is this a message that everybody should see or only users with certain admin permission(s)?

mxr576: Also, the code still reacts to non-HTML requests if I am not mistaken.

shishir-intelli: > It just occurred to me, is this a message that everybody should see or only users with certain admin permission(s)?

Everybody as of now.

shishir-intelli: > Also, the code still reacts to non-HTML requests if I am not mistaken.

Done

mxr576: > >   It just occurred to me, is this a message that everybody should see or only users with certain admin permission(s)?

> Everybody as of now.

Well, I am skeptical if Apigee customers are going to like that their customers are going to see messages like this if they still keep using the Teams feature on their site - which seems to be working. 

My opinion that the goal should be reaching out the right audience and convincing them to disable this feature.

image
(The screenshot was made as a simple logged in user.)

shishir-intelli: I agree with you. 
We can show different notice message for simple logged in user or no message at all for non-admin?

@giteshk What do you think?

modules/apigee_edge_teams/apigee_edge_teams.install
mxr576: > Ideally this shouldn't work as of now, because it is not completely supported in Apigee X

So are we try to block enabling the Teams module on Apigee X because of these entries [from known limitations ](https://cloud.google.com/apigee/docs/api-platform/publish/organizing-client-app-ownership?hl=en#appgroups-limitations-and-known-issues)...?

> Metrics
> Quota enforcement
> Updates to the scope for API keys

I am just still trying to understand why this kill switch is here and not **only in apigee_m10n_teams** if the actual problem/limitations arises on a Monetization enabled organization (around Monetization features)? Why other customers on Apigee X should not use this feature now?

giteshk: Yes, We have to enforce the limitations.

Enabling Teams Module on monetized Apigee X orgs will set an expectation that Team Apps will follow the restrictions that Monetization Plans that users may have purchased.

We will enable this once AppGroup support for Monetization is announced.

mxr576: ~~That basically means that customers who use the Teams feature on Apigee Edge cannot be migrated to Apigee X, they are blocked.~~

~~This is a game changer in a conversation with a customer who uses team without Monetization and would like to move to Apigee X in this year...~~

shishir-intelli: > That basically means that customers who use the Teams feature on Apigee Edge cannot be migrated to Apigee X, they are blocked.

No, They can migrate to Apigee X and use Teams module as well, only if Monetization is not enabled in the organisation which they are using.

eg:
```
        "monetizationConfig": {
            "enabled": false
        },
```

If 
```
        "monetizationConfig": {
            "enabled": true
        },
```
Monetization is enabled(TRUE) in the Org then the customer wont be able to use Teams module and Monetization Teams module in Apigee X.

mxr576: Ahh right, sorry, my bad, one morning coffee was not enough today...

modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
shishir-intelli: @giteshk and @mxr576 
As discussed, we will implement this only for admin and no error/notice message for non-admin users.

shishir-intelli: Done

modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * If monetization enabled in Apigee X org alert the user.
+   *
+   * @param \Symfony\Component\HttpKernel\Event\RequestEvent $event
+   *   The event.
+   */
+  public function validateApigeexTeamEnabled(RequestEvent $event) {
+    // Check only for html request and admin users.
+    if (($this->currentUser->id() == 1 || $this->currentUser->hasRole('administrator')) && $event->getRequest()->getRequestFormat() === 'html') {
```
mxr576: >  $this->currentUser->hasRole('administrator'))

Well, depending on a bespoke role as a conditional is considered as an anti pattern in Drupal. Multiple roles (personas) could have a set of a common permissions that should trigger this warning, I'd rather try to identify those. 

Why I am also flagging this because it is common best practice creating dedicated subadministrator roles on Drupal projects for customers.

Potential permissions to check instead:
* ~~`administer apigee edge`~~ just because I have this permission I cannot manage installed dependencies on a site
* `administer site configuration` or `administer modules` instead? Since the latter can deactivate the module on a project that does not leverage Drupal 8+'s configuration as code approach for deployments.


and the `$this->currentUser->id() == 1` should become redundant since the current special behavior of UID1 that it has all permissions

mxr576 (changes requested):

modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
```diff
@@ -0,0 +1,111 @@
...
+ * Validates that Apigee X Team is enabled on every Team page request.
+ */
+final class ValidateApigeeXTeamEnabledSubscriber implements EventSubscriberInterface {
+
+  /**
+   * The current user.
+   *
+   * @var \Drupal\Core\Session\AccountInterface
+   */
+  private $currentUser;
```
mxr576:
```suggestion
  private AccountInterface $currentUser;
```

modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
mxr576:
```suggestion
    if ($this->currentUser->hasPermission('administer modules') && $event->getRequest()->getRequestFormat() === 'html') {
```

`user.permissions` should be a default cache.context... so hopefully we do not need to bubble up cacheability info here... :crossed_fingers:

shishir-intelli: @mxr576 ,
Thank you for the suggestion.

modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
```diff
@@ -0,0 +1,111 @@
...
+   */
+  public function validateApigeexTeamEnabled(RequestEvent $event) {
+    // Check only for html request and admin users.
+    if ($this->currentUser->hasPermission('administer modules') && $event->getRequest()->getRequestFormat() === 'html') {
+      /** @var \Symfony\Component\Routing\Route $current_route */
+      if (($current_route = $event->getRequest()->get('_route')) && (strpos($current_route, 'entity.team') !== FALSE || strpos($current_route, 'settings.team') !== FALSE)) {
+        $organization = $this->orgController->load($this->connector->getOrganization());
+        if ($organization && $this->orgController->isOrganizationApigeeX()) {
+          if ($organization->getAddonsConfig() || TRUE === $organization->getAddonsConfig()->getMonetizationConfig()->getEnabled()) {
+            $this->messenger->addError('The Teams module functionality is not available for monetization enabled org on Apigee X / Hybrid and should be uninstalled.');
```
mxr576:
```suggestion
            $this->messenger->addError($this->t('The Teams module functionality is not available for monetization enabled org on Apigee X / Hybrid and should be uninstalled.'));
```

Ahh, almost missed this one... and inject the string translation service.

modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
```diff
@@ -0,0 +1,111 @@
...
+    $this->messenger = $messenger;
+  }
+
+  /**
+   * If monetization enabled in Apigee X org alert the user.
+   *
+   * @param \Symfony\Component\HttpKernel\Event\RequestEvent $event
+   *   The event.
+   */
+  public function validateApigeexTeamEnabled(RequestEvent $event) {
```
mxr576: Last nitpicks, missing method return types inside this class.
```suggestion
  public function validateApigeexTeamEnabled(RequestEvent $event): void {
```

modules/apigee_edge_teams/src/EventSubscriber/ValidateApigeeXTeamEnabledSubscriber.php
```diff
@@ -0,0 +1,111 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents() {
```
mxr576:
```suggestion
  public static function getSubscribedEvents(): array {
```

mxr576 (commented):
Otherwise LGTM

divya-intelli (approved):
LGTM

PR #950, opened 2023-10-05
EntityQuery: Add support for filtering by attribute value

### LINKED ISSUE #949: Entity query: support filtering by attribute value
Extend entity query support on Apigee entities with an option to filter them by attribute `value.` 

Example.: `$query->condition('attribute_value.access', 'private');`

Later this feature can be extended to support filtering by other nested values as well.  "." is also used a seperator in Drupal core for accessing a specific property inside a field in entity query.

mxr576:
Closes #949 

TODOs

- [ ] Extend test coverage in https://github.com/apigee/apigee-edge-drupal/blob/3.x/tests/src/Unit/ConditionTest.php (#helpWanted)


shishir-intelli (approved):
LGTM

boobaa:
Original version of this PR works great for filtering on attribute names that does NOT contain any underscores. If an attribute contains underscores (eg. `foo_bar`), then the original version was looking for a "normalized" attribute name (eg. `fooBar`).

The updated version allows underscores in attribute names, by NOT normalizing the attribute name if the Condition's `$property` starts with `AttributeValue.`.

Issue #944, opened 2023-09-27
Improve performance of app listing pages


mxr576:
## Description
We were debugging performance issues related to app listing pages for an unnamed Apigee customer. As part of that process, we enabled the Apigee Edge Debug module and inspected outgoing API calls.

We realized that developer and team app listing pages send as many API calls as many apps they list for different reasons, which basically leads to the ORM's infamous [N+1 problem](https://planetscale.com/blog/what-is-n-1-query-problem-and-how-to-solve-it),

So for a start, we do not and cannot permanently cache app credentials... therefore we had/have tricky solutions in the code to fetch them on demand from the API gateway when needed.

### Problem 1: When display modes are in use

When display modes are configured to be used for display apps on `admin/config/apigee-edge/app-settings/developer-apps/display-settings` or on `/admin/config/apigee-edge/app-settings/team-apps/display-settings` then the root cause of the problem is the the selected display mode ("Default" by default) also used for rendering entries in the list - even if a minimal subset of the fields are rendered on the list - and the Default display mode displays the "Credentials" field that triggered on-demand fetching on API key(s) in an app.

This can be reproduced by setting the display mode and refreshing the app listing page of the user and checking logs generated by the debug module. The number of API calls to `/apps/[uuid]` endpoint with an Apigee Edge instance connection should raise your concerns and that is created by `$entity->getCredentials() ` in `apigee_edge_entity_view()` [here](https://github.com/apigee/apigee-edge-drupal/blob/3e360ab0673016062e15425156016f1039d007ee/apigee_edge.module#L390-L402).

When Credentials field is removed from the display mode the same problem arises, also originating in `apigee_edge_entity_view()` but triggered by the `$app_warnings_checker->getWarnings($entity)` [here](https://github.com/apigee/apigee-edge-drupal/blob/3e360ab0673016062e15425156016f1039d007ee/apigee_edge.module#L452-L465).

#### Proposed solution

Entity list builder MUST use a dedicated view mode for listing apps in the list or should not use any. Most probably the refactoring should start in `\Drupal\apigee_edge\Entity\ListBuilder\EdgeEntityListBuilder::render()` that involves `\Drupal\apigee_edge\Entity\ListBuilder\EdgeEntityListBuilder::getDisplaySettings()` and returns the same display mode as it is used on viewing a single app entity.

[...]

PR #957, opened 2023-10-11
POC: Render app credential warnings with BigPipe

mxr576:
This solves Problem 1 from #944.


mxr576:
So this is the first version of the solution and it is subject to changes, especially around wording.

The main idea that we would keep the old way until 4.0.0 because downstream projects may (and most probably) customized how app credentials are rendered. 

Also, if you would wonder, this also works without Big Pipe having enabled, it just renders everything synchronously, so we do not need to make it a requirement - maybe a warning on admin/reports/status page for that would be useful if it is not enabled.

apigee_edge.module
```diff
@@ -138,6 +139,12 @@ function apigee_edge_theme() {
       'render element' => 'elements',
       'base hook' => 'apigee_secret',
     ],
+    'apigee_app_credential_warnings' => [
```
mxr576, line 142: This should superior flexibility for downstream projects to customize these.

templates/apigee-app-credential-warnings-placeholder.html.twig
```diff
@@ -0,0 +1,5 @@
+<div class="app-credential-warnings">
```
mxr576, line 1: TODO add docblocks

templates/apigee-app-credential-warnings.html.twig
```diff
@@ -0,0 +1,5 @@
+<div class="app-credential-warnings">
```
mxr576, line 1: TODO add docblocks

src/LazyBuilder/AppWarningsLazyBuilder.php
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    return [];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function trustedCallbacks(): array {
+    return ['lazyBuilder'];
```
mxr576, line 69: (Why on earth invokable classes are not supported...???!)

mxr576:
@divya-intelli @giteshk what is your opinion about the suggested approach? Does this PR worth to be continued? Anything that you would change on the base idea?

config/schema/apigee_edge.schema.yml
```diff
@@ -66,6 +66,9 @@ apigee_edge.common_app_settings:
     callback_url_placeholder:
       type: label
       label: 'Placeholder for a Callback URL'
+    app_credential_warnings_bc_mode:
+      type: boolean
+      deprecated: "The 'app_credential_warnings_bc_mode' config schema is deprecated in Apigee Edge 3.0.3 and will be removed from Apigee Edge 4.0.0."
```
mxr576, line 71: Todo adjust version number before merge.

divya-intelli:
> @divya-intelli @giteshk what is your opinion about the suggested approach? Does this PR worth to be continued? Anything that you would change on the base idea?

@mxr576 , its good to have this feature which will speed up the page. 
Thank you and appreciate for your time. 
We will review once the PR is ready.

PR #960, opened 2023-10-16
Added kernel test case for Appgroup Team Invitation and EntityControllerCache

shishir-intelli:
For AppGroup, added kernel test case for TeamInvitation and EntityControllerCache files.


divya-intelli (approved):
LGTM

PR #965, opened 2023-10-26
Remove Drupal core patch

### LINKED ISSUE #964: Drupal core patch changes Entity API
## Description
The patch added in #931 adds two methods to EntityInterface (getOriginal and setOriginal). Classes that implement EntityInterface will need to implement these two methods or be declared as abstract.

## Steps to Reproduce
1. Make sure that in your project's root composer.json `extra.enable-patching` is set to `true`.
2. Require drupal/apigee_edge 3.0.2.
3. Make sure that the "Add a method to access the original property" patch is applied to Drupal core.
4. Require and enable an extension that has a class that implements EntityInterface (example: drupal/simple_oauth).
5. Configure and use the other extension as described in its documentation.

## Actual Behavior
PHP Fatal error.

## Expected Behavior
Updating to drupal/apigee_edge 3.0.2 should not cause other extensions to fail.

## Version Info
3.0.2

mxr576:
Closes #964

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,11 +33,6 @@
             "services": {
                 "drush.services.yml": "^12"
             }
-        },
-        "patches": {
-            "drupal/core": {
-                "Add a method to access the original property": "https://www.drupal.org/files/issues/2023-07-22/2839195-105.patch"
-            }
         }
     }
 }


mxr576:
The CI pipeline should keep working... hopefully nothing else gets broken in runtime.

divya-intelli (approved):
LGTM

PR #968, opened 2023-11-03
Setting up Apigee X environment to run test cases

shishir-intelli:
Setting up the Apigee X environment at runtime for testcase, so that we can segregate the Apigee Edge and X test case.


divya-intelli (approved):
LGTM

PR #980, opened 2023-11-15
Skipping Functional and FunctionalJavascript test for Hybrid Org

shishir-intelli:
Added check to skip test if the org is Hybrid for Apigee Edge Functional and FunctionalJavascript test cases.


divya-intelli (approved):
LGTM

PR #904, opened 2023-08-01
Code coverage changes for 3.x branch

divya-intelli:
Fixes #903

PR #986, opened 2023-11-21
Trigger code coverage for Drupal 10.1 and php 8.1

shishir-intelli:
Fix code coverage not getting triggered for 10x.


shishir-intelli:
Already created PR #904

PR #988, opened 2023-11-21
Added check for field_ui module enabled durning getting field prefix

### LINKED ISSUE #983: Teams lose field values if field_ui is enabled/disabled
## Description
To replicate 
- Start With core field_ui module enabled
- add some fields to the Team entity. 
- Create/edit a team and add a value to the newly created field.
- View team -> value should present
- Disable field_ui module
- view team -> field value has _vanished_
- Re-enable  field_ui module
- view team -> field value has reappeared

Below is the offending code:
_$this->config->get('field_ui.settings')->get('field_prefix');_
Returns a different value depending if field_ui is enabled.

` /**
   * Returns the field UI's field name prefix.
   *
   * @return string
   *   Prefix of the field.
   */
  protected function getFieldPrefix(): string {
    return (string) $this->config->get('field_ui.settings')->get('field_prefix');
  }`

shishir-intelli:
Closes #983 
Added check for field_ui module enabled durning getting field prefix


audave:
This will prevent future errors related to #983, but I can imagine it breaking a lot of existing sites. 

Anyone who currently has field_ui disabled, quite a common setting on productions instances, will find that all their fields will lose data as they are now looking for different attribute name on apigee.

shishir-intelli:
> This will prevent future errors related to #983, but I can imagine it breaking a lot of existing sites.
> 
> Anyone who currently has field_ui disabled, quite a common setting on productions instances, will find that all their fields will lose data as they are now looking for different attribute name on apigee.

@audave,  I agree to what you said,
IMO, we can check if `field_ui` module is enabled, 
If yes, return the field name prefix else check/search for the value `'field_'` in field name, if exists return default value `'field_'` or else return `""`.
Any suggestions?

audave:
The problem is not so much what the prefix is but that it changes depending on the state of a different module.

In my opinion there is no need to account for a prefix at all. This would lead to drupal fields being stored as apigee attributes with the name field_myfieldname - but unless you are looking directly at the management API I don't see why that would be a problem.

Whatever the final solution is, it is probably worth writing an update hook to copy existing data to the new attribute.

shishir-intelli:
We have added a dependency of Field UI module and rolled back the previous changes which will prevent such issues in future.

divya-intelli (approved):
LGTM

audave:
I do not thing this approach is a good idea.

In general the field_ui module is a UI module, its there help you configure fields on your content types (or teams). At almost every place I have worked in the last 12 years it has been disabled in production (and enabled in dev). 

This approach forces people to enable a UI module in production, which they wouldn't otherwise do, and it shouldn't be relevant to the functioning of the apigee module.

Also.

Lots of sites will loose their data when this is implemented. As stated in my previous comment, the problem is that the prefix changes. 

So anyone who currently does not have the field_ui module enabled will have all the field data stored as an apigee attribute field_myfieldname -> After they are forced to enabled the field_ui module as part of this change, this will no longer be the correct attribute name on apigee.

apigee_edge.install
```diff
@@ -389,3 +389,11 @@ function apigee_edge_update_9002() {
   $edge_settings['content']['callbackUrl'] = $new_edge_settings['content']['callbackUrl'];
   $config_storage->write('core.entity_view_display.developer_app.developer_app.default', $edge_settings);
 }
+
+/**
+ * Install Field UI module to maintain dependency.
+ */
+function apigee_edge_update_9003() {
```
mxr576, line 396: The update hook for Drupal 10+ should start with 10.

https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_update_N/10

(Probably old update hooks should have been removed when 3.x branch started and Drupal core major version was bumped)

shishir-intelli: > The update hook for Drupal 10+ should start with 10.

Thanks! Noted 👍 

> (Probably old update hooks should have been removed when 3.x branch started and Drupal core major version was bumped)

We will take care regarding this next time.

shishir-intelli:
@audave 
> I do not thing this approach is a good idea.
> 
> In general the field_ui module is a UI module, its there help you configure fields on your content types (or teams). At almost every place I have worked in the last 12 years it has been disabled in production (and enabled in dev).
> 
> This approach forces people to enable a UI module in production, which they wouldn't otherwise do, and it shouldn't be relevant to the functioning of the apigee module.
> 
> Also.
> 
> Lots of sites will loose their data when this is implemented. As stated in my previous comment, the problem is that the prefix changes.
> 
> So anyone who currently does not have the field_ui module enabled will have all the field data stored as an apigee attribute field_myfieldname -> After they are forced to enabled the field_ui module as part of this change, this will no longer be the correct attribute name on apigee.

Because in Apigee management UI [Custom attribute name size](https://cloud.google.com/apigee/docs/api-platform/reference/limits) has limits, that is why we remove the prefix to increase the field name size.

Solution (Including the changes in the PR):
We can provide a configurable setting form in Drupal portal to edit the value of the field prefix
(_Ideally updating the original field prefix value in field_ui.settings.yml_),
and set the default prefix value as `field_`, 

So that anyone who currently do not have the `field_ui` module enabled in Drupal and has `field prefix` or `no prefix` already stored in Apigee attribute can manage the field_prefix in Drupal portal through that configurable setting form to prevent data lose when `field_ui` module is forcefully enabled after this changes.

PR #997, opened 2023-11-30
Fix error - Creation of dynamic property $testTokenData is deprecated

shishir-intelli:
Fix error `Creation of dynamic property $testTokenData is deprecated` durning running Apigee X test cases.


divya-intelli (approved):
LGTM

PR #1005, opened 2023-12-08
Fix for Null value compatibility with PHP 8.1 & above

### LINKED ISSUE #972: Null value compatibility with PHP 8.1

kedarkhaire:
Fixes #972


kedarkhaire:
Closing the PR as there wont be bug fixes for 2x only security fixes qould be there.

PR #1011, opened 2023-12-14
Added AppGroup functional test case for team list builder

shishir-intelli:
Added AppGroup functional test case for team list builder


divya-intelli (approved):
LGTM

Issue #998, opened 2023-11-30
Renaming Apigee Edge to Apigee


shishir-intelli:
## Is your feature request related to a problem? Please describe.
Product name has been updated from `Apigee Edge` to `Apigee`.

## Describe the solution you would like
For now, 

- Change title & descriptions in module.info files.
- Change title and descriptions in the Form fields.
- Change title and descriptions in the routes, links, menu etc yml files.
- Change title and descriptions in the classes, methods, properties comments.
- Change in README.md and Contribution.md files.

## Expected Behavior

- Module descriptions are updated.
- Drupal documentation pages are updated.
- Drupal.org module page
- Github README.md files

PR #1020, opened 2024-01-03
Rename form fields  titles and description from Apigee Edge to Apigee

shishir-intelli:
#998 
Rename form fields titles and description from Apigee Edge to Apigee


divya-intelli (approved):
LGTM

PR #1025, opened 2024-01-11
Fix test case failing due to getCacheTags() on null error

shishir-intelli:
Fix test case failing due to getCacheTags() on null error


divya-intelli (approved):
LGTM

Issue #1028, opened 2024-01-17
Repeatedly APIs called durning AppGroup and member edit


shishir-intelli:
## Description
High latency because of multiple API calls happening for edit app groups.

## Apigee Info
Apigee X

shishir-intelli:
We are working on improving the cache mechanism for AppGroup.

PR #1030, opened 2024-01-25
Fix redirection to canonical issue and preventing API calls

shishir-intelli:
#1028 
Fix redirection to canonical issue and preventing API calls which was causing latency.


divya-intelli (approved):
LGTM

PR #1031, opened 2024-01-26
Acceptance link issue when the same mail address due to a case-sensitive error.

vpa24:
When inviting a new member to my team, consider the following example: the user's email is [hello_Word@example.com](mailto:hello_Word@example.com), but I entered [Hello_word@example.com](mailto:Hello_word@example.com). When they click on the acceptance link, they can't join the team due to a **case-sensitive error**.

In my code, it first retrieves the user object using the `user_load_by_mail` function with the recipient's email. If the user object is found, it retrieves the original user's email; otherwise, it obtains the recipient's email.

### DIFF
--- a/modules/apigee_edge_teams/src/Entity/TeamInvitation.php
+++ b/modules/apigee_edge_teams/src/Entity/TeamInvitation.php
@@ -190,7 +190,9 @@ class TeamInvitation extends ContentEntityBase implements TeamInvitationInterfac
    * {@inheritdoc}
    */
   public function getRecipient(): ?string {
-    return $this->get('recipient')->value;
+    $recipientEmail = $this->get('recipient')->value;;
+    $user = user_load_by_mail($recipientEmail);
+    return $user ? $user->getEmail() : $recipientEmail;
   }
 
   /**


modules/apigee_edge_teams/src/Entity/TeamInvitation.php
```diff
@@ -190,7 +190,9 @@ public function setStatus(int $status): TeamInvitationInterface {
    * {@inheritdoc}
    */
   public function getRecipient(): ?string {
-    return $this->get('recipient')->value;
+    $recipientEmail = $this->get('recipient')->value;;
```
mxr576, line 193: There is an extra `;` at the end of the line.

vpa24: Thanks

modules/apigee_edge_teams/src/Entity/TeamInvitation.php
```diff
@@ -190,7 +190,9 @@ public function setStatus(int $status): TeamInvitationInterface {
...
+    $user = user_load_by_mail($recipientEmail);
```
mxr576, line 194: Loading user approach looks quite expensive and if the email server case sensitive then it could also lead to problems.

Case sensitivity is a delicate topic anyway... 

https://www.googlecloudcommunity.com/gc/Apigee/Apigee-Edge-to-Apigee-X-migration-INVALID-ARGUMENT-email-address/m-p/641748/highlight/true#M77284

PR #1034, opened 2024-02-01
Removing TeamInvitationQueryAccessSubscriber for AppGroup member edit/add

shishir-intelli:
#1028 
To prevent unnecessary API calls to AppGroup/Teams listing we are returning the `TeamInvitationQueryAccessSubscriber` for Apigee-X organization because we are storing member information in database tables and no need to query Teams from Apigee.


divya-intelli (approved):
LGTM

PR #1050, opened 2024-04-11
Adding validation for email-id field

### LINKED ISSUE #730: Error sending email with capitalization to Apigee.

kedarkhaire:
Fixes #730 
For Hybrid instance (Org type = ApigeeX), 
add field validation for user email field on all operations. Validation will be to check the string is in lowercase. 
Should be triggered on new user creation, user edit, login, SAML auth, drupal default login & LDAP login.

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -338,6 +338,31 @@ function apigee_edge_entity_base_field_info_alter(&$fields, EntityTypeInterface 
 }
 
 /**
+ * Implements hook_form_FORM_ID_alter() for install_configure_form().
+ *
+ * Allows the profile to alter the site configuration form.
+ */
+function apigee_edge_form_install_configure_form_alter(&$form, FormStateInterface $form_state) {
+  $form['#validate'][] = 'apigee_edge_checkemail';
+}
+
+function apigee_edge_checkemail(&$form, FormStateInterface $form_state) {
+  try {
+    $org_controller = \Drupal::service('apigee_edge.controller.organization');
+    // Check if org is ApigeeX.
+    if ($org_controller->isOrganizationApigeeX()) {
+      if (preg_match('/[A-Z]/', $form_state->getValue(['account', 'mail']))) {
+        $form_state->setErrorByName('account][mail', 'This email address accepts only lowercase characters.');
+      }
+   }
+ }
+ catch (\Exception $exception) {
+   // If not able to connect to Apigee Edge.
+   Drupal::logger('apigee_edge')->error($exception->getMessage());
+ }
+}
+
+/**
  * Implements hook_entity_extra_field_info().
  */
 function apigee_edge_entity_extra_field_info() {
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -326,7 +326,7 @@ function apigee_edge_entity_base_field_info_alter(&$fields, EntityTypeInterface 
     $mail = $fields['mail'];
     $mail->setRequired(TRUE);
     $mail->addConstraint('DeveloperMailUnique');
-    $mail->addConstraint('LowercaseEmail');
+    $mail->addConstraint('DeveloperLowercaseEmail');
 
     // Add a bundle to these fields to allow other modules to display them
     // as configurable (fields added through the UI or configuration do have a
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -718,6 +718,8 @@ function apigee_edge_form_user_form_alter(&$form, FormStateInterface $form_state
   // connection validation apigee_edge_module_implements_alter() ensures that
   // form_alter hook is called in the last time.
   $validation_functions[] = 'apigee_edge_form_user_form_developer_email_validate';
+  // Validates whether the provided email address is in lowercase.
+  $validation_functions[] = 'apigee_edge_email_field_validate';
   $form['#validate'] = array_merge($validation_functions, $form['#validate']);
 }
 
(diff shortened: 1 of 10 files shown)


apigee_edge.module
```diff
@@ -718,6 +718,8 @@ function apigee_edge_form_user_form_alter(&$form, FormStateInterface $form_state
   // connection validation apigee_edge_module_implements_alter() ensures that
   // form_alter hook is called in the last time.
   $validation_functions[] = 'apigee_edge_form_user_form_developer_email_validate';
+  // Validates whether the provided email address is in lowercase.
+  $validation_functions[] = 'apigee_edge_email_field_validate';
```
mxr576: Form validation can be bypassed when a user is created via JSONAPI. Entity or field level constraints are the only solutions that all components should obey.

kedarkhaire: Hi @mxr576 Sure, I will add field constraints for email field.

src/Plugin/Validation/Constraint/LowercaseEmailConstraintValidator.php
```diff
@@ -0,0 +1,41 @@
...
+      // First check if the organization is ApigeeX.
+      try {
+        /** @var \Drupal\apigee_edge\SDKConnectorInterface $sdk_connector */
+        $sdk_connector = \Drupal::service('apigee_edge.sdk_connector');
+        $org_controller = \Drupal::service('apigee_edge.controller.organization');
+        /** @var \Apigee\Edge\Api\Management\Entity\Organization $organization */
+        $organization = $org_controller->load($sdk_connector->getOrganization());
+
+        // Check if organization is ApigeeX.
+        if ($organization && ('CLOUD' === $organization->getRuntimeType() || 'HYBRID' === $organization->getRuntimeType())) {
```
giteshk: Can this check for ApigeeX instance be moved to the module file instead of here?

src/Plugin/Validation/Constraint/LowercaseEmailConstraintValidator.php
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validate($value, Constraint $constraint) {
+    foreach ($value as $item) {
+      // First check if the organization is ApigeeX.
+      try {
+        /** @var \Drupal\apigee_edge\SDKConnectorInterface $sdk_connector */
+        $sdk_connector = \Drupal::service('apigee_edge.sdk_connector');
```
mxr576: Dependency injection is possible in constraint validators as well. 

See: `\Drupal\Core\Entity\Plugin\Validation\Constraint\ValidReferenceConstraintValidator`

src/Plugin/Validation/Constraint/LowercaseEmailConstraintValidator.php
mxr576: What would be the benefit of that? The constraint is added in "build-time", however this should be a runtime check because what if a site gets re-connected to an Apigee X org after it was connected to Apigee Edge.

mxr576:
```suggestion
        if ($controller->isOrganizationApigeeX()) {
```

https://github.com/apigee/apigee-edge-drupal/blob/0f0e59aff7cc680dd9ec49825728af94df904b84/src/Entity/Controller/OrganizationController.php#L151

This also makes the whole try-catch dead code, if I am not mistaken.

mxr576: What may help on potential performance issues caused by fetching org information multiple times (let's say when a batch operation is running) is storing the fact about whether this is an Apigee X org in an in-memory cache (or adding in-memory cache to the org controller if it does not have it yet).

src/Plugin/Validation/Constraint/LowercaseEmailConstraintValidator.php
kedarkhaire: Updated the code. Thanks for the suggestion.

src/Plugin/Validation/Constraint/LowercaseEmailConstraintValidator.php
kedarkhaire: Updated the code. Thanks for the suggestion.

src/CliService.php
```diff
@@ -81,7 +81,7 @@ public function createEdgeRoleForDrupal(
     string $password,
     ?string $base_url,
     ?string $role_name,
-    ?bool $force
+    ?bool $force,
```
mxr576: IMO these changes should go to a dedicated PR that gets merged independently and before this one.

apigee_edge.module
```diff
@@ -336,6 +337,31 @@ function apigee_edge_entity_base_field_info_alter(&$fields, EntityTypeInterface
...
+/**
+ * Implements hook_form_FORM_ID_alter() for install_configure_form().
+ *
+ * Allows the profile to alter the site configuration form.
+ */
+function apigee_edge_form_install_configure_form_alter(&$form, FormStateInterface $form_state) {
+  $form['#validate'][] = 'apigee_edge_checkemail';
+}
+
+function apigee_edge_checkemail(&$form, FormStateInterface $form_state) {
```
mxr576:
```suggestion
/**
*  @internal
*/
function _apigee_edge_site_install_form_check_email(array &$form, FormStateInterface $form_state): void {
```

The benefit is marking this as internal with all power, nobody should use this function outside of the module.

But it still deserves a better doc block.

apigee_edge.module
```diff
@@ -336,6 +337,31 @@ function apigee_edge_entity_base_field_info_alter(&$fields, EntityTypeInterface
...
+  try {
+    $org_controller = \Drupal::service('apigee_edge.controller.organization');
+    // Check if org is ApigeeX.
+    if ($org_controller->isOrganizationApigeeX()) {
```
mxr576: If the code logic from LowercaseEmailConstraintValidator would be decoupled to a dedicated service then the code could be more DRY because the validation logic would be centralized into one source of truth that both code could leverage.

apigee_edge.module
kedarkhaire: Noted, will apply the change

src/CliService.php
kedarkhaire: Updated. moved to #1051

apigee_edge.module
kedarkhaire: Updated

shishir-intelli (approved):
LGTM

PR #1073, opened 2024-08-20
Bump PHP client version to 3.0.6 & Drupal core version to 10.2

kedarkhaire:
Version bump for PHP Client and Drupal core


shishir-intelli (approved):
LGTM

PR #1089, opened 2024-10-16
Bump PHP client version to 3.0.8

kedarkhaire:
Bump PHP client version to 3.0.8


kedarkhaire:
Not necessary as no breaking change.

Issue #1098, opened 2024-11-22
Drupal Apigee API Product RBAC not filtering API Products based on Roles


urbanenomad:
## Apigee Info
We have both apigee hybrid and apigee-x instances that we want to support

## Steps to Reproduce
1. Enable apigee API product RBAC
![Screenshot 2024-11-21 at 9 29 54 AM](https://github.com/user-attachments/assets/55665a96-7c76-4c07-be81-82533567c1a2)

2. Configured Custom Attribute on some test apigee products
![Screenshot 2024-11-21 at 9 28 53 AM](https://github.com/user-attachments/assets/2f36654d-5526-4c11-a57d-00bc2f800040)

## Actual Behavior
* Non-admin account that was set to one of the attirbute roles still shows all products even more than if we did just products by visiblity.  It now shows all the products.  
![Screenshot 2024-11-22 at 12 07 00 PM](https://github.com/user-attachments/assets/a31e6fa6-da4a-41c7-b8ea-76188e52a689)

* We also did not enable "Bypass API product access control" nor did we enable "Show API products with missing or empty attribute to everyone"

## Expected Behavior
We would have hoped to see when a developer trys to create an App they only see the API products that have their role in the custom attribute

## Version Info
Drupal 10.3.6

Modules: 
Apigee: 3.0.10
Apigee API Catalog 3.0.8

kedarkhaire:
Hi @urbanenomad 
Thanks for creating a ticket here. We will look into it.
Thanks!

urbanenomad:
Is there any updates on this?

urbanenomad:
Hey so I think I figured out the issue at the following lines.

https://github.com/apigee/apigee-edge-drupal/blob/1e480cfb112892a54f4480565e355c9d53c260ac/modules/apigee_edge_apiproduct_rbac/apigee_edge_apiproduct_rbac.module#L70

the line should read like the following.

$result = AccessResult::forbidden("{$operation} is not allowed on {$entity->label()} API product.");

Same for this line.

https://github.com/apigee/apigee-edge-drupal/blob/1e480cfb112892a54f4480565e355c9d53c260ac/modules/apigee_edge_apiproduct_rbac/apigee_edge_apiproduct_rbac.module#L97

$result = AccessResult::forbidden("{$operation} is not allowed on {$entity->label()} API product.");

This resulted with the correct output on the App Create page.

kedarkhaire:
Hi @urbanenomad 
[...]

PR #1101, opened 2024-12-12
Update Accessresult for ApiProducts listing

urbanenomad:
Changed from neutral to forbidden so that API Products do not show up on the assign operation or Create App when using the RBAC service to control access to API Products.  This is to solve issue https://github.com/apigee/apigee-edge-drupal/issues/1098

### DIFF
--- a/modules/apigee_edge_apiproduct_rbac/apigee_edge_apiproduct_rbac.module
+++ b/modules/apigee_edge_apiproduct_rbac/apigee_edge_apiproduct_rbac.module
@@ -67,7 +67,7 @@ function apigee_edge_apiproduct_rbac_api_product_access(EntityInterface $entity,
 
     if (empty($entity->getAttributeValue($rbac_attribute_name))) {
       if ('assign' === $operation) {
-        $result = AccessResult::neutral("{$operation} is not allowed on {$entity->label()} API product.");
+        $result = AccessResult::forbidden("{$operation} is not allowed on {$entity->label()} API product.");
       }
       elseif ($config->get('grant_access_if_attribute_missing')) {
         $result = AccessResult::allowed();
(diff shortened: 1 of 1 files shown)


kedarkhaire:
Hi @urbanenomad Thanks for creating the PR, we will check the implementation, but also it is mandatory that all tests pass in your repository, to make sure that the other working functionalities are unharmed.

Thanks!

urbanenomad:
I am a bit confused on some of these errors.  There seems to be some kind of JWT token error in PHP 81./Drupal 10.3.x/X

"1) Drupal\Tests\apigee_edge_teams\Functional\ApigeeX\AccessTest::testAccess
Drupal\Core\Entity\EntityStorageException: ***"error":"invalid_grant","error_description":"Invalid JWT Signature."***". 

I am not sure why this error is occuring.  I don't think I did any changes to the JWT signature?  Do I need to test the code against a working apigee-x instance?

kedarkhaire:
> ror is occuring. I don't think I did any changes

Hi @urbanenomad 
Yes the issue is not related to the changes you made, we just need to update the credentials for test. I will do that tomorrow with the working ones, and that issue will be solved.

I saw other issue on my test fork. Can you have a look at this [link](https://github.com/kedarkhaire/apigee-edge-drupal/actions/runs/12346964604/job/34453339562#step:16:223).

urbanenomad:
I don't have an active apigee edge instance that seems to be something I need to test these unit tests.  Is there a public one I can use to validate the unit tests from my fork?

urbanenomad:
@kedarkhaire another thing I see is that your kedarkhaire fork there seems to be 2 errors for similar issues for both AccessTest::testAcccess as well as UiTest::testUi.  But in the original repo there only seems to be one error for the UiTest::testUi.  Both seems to be based on the same error where it looks like the team_name doesn't seem to exist which is causing the undefined array key error?  Also my apigee-x instance I have for testing the drupal portal seems to have monetization turned on so apigee-teams module does not seem to work in my portal since it is not designed to work with monetization turned on.  Is there another apigee instance edge or x that we can use to test this?

kedarkhaire:
Hi @urbanenomad 
Thanks for your inputs, I will check and update.

kedarkhaire (approved):
LGTM

Issue #1106, opened 2025-01-09
PHPCS update on drupal/coder - Drupal.NamingConventions.ValidClassName.NoUpperAcronyms error


kedarkhaire:
## Description
On PHP CS check - with new update of drupal/coder v8.3.27, we get new error of `Drupal.NamingConventions.ValidClassName.NoUpperAcronyms` on following files
`/src/Plugin/ApigeeFieldStorageFormat/JSON.php` & `/src/Plugin/ApigeeFieldStorageFormat/CSV.php`

Issue #1103, opened 2024-12-18
Deprecated errors with recent Apigee Edge update 


hollyw2025:
## Description
Hello, We ran into a couple of deprecated errors with recent Apigee Edge update. 
I open a case at the drupal site: 
https://www.drupal.org/project/apigee_dev_portal/issues/3494359 

## Apigee Info
We connected to Apigee Edge at the back end. 

composer -V: 2.8.4, which is the most recent version 

Drupal modules: 
Drupal core 10.4.0
drupal/apigee_edge: 3.0.10

## Steps to Reproduce
error message showed up at the team list page

## Expected Behavior
I created a local patch which could be applied using Git patch, but can't use Composer update to apply.  I attached the patch. 
Could you please fix the related deprecated code from your end per the error message? 

Here are the error list: 

Deprecated: Optional parameter $config_factory declared before required parameter $app_warnings_checker is implicitly treated as a required parameter in /var/www/web/modules/contrib/apigee_edge/src/Entity/ListBuilder/AppListBuilder.php on line 97

Deprecated: Optional parameter $entity_type_bundle_info declared before required parameter $org_controller is implicitly treated as a required parameter in /code/web/modules/contrib/apigee_edge/modules/apigee_edge_teams/src/Entity/Storage/TeamMemberRoleStorage.php on line 97 

Deprecated: Optional parameter $module_handler declared before required parameter $org_controller is implicitly treated as a required parameter in /var/www/web/modules/contrib/apigee_edge/modules/apigee_edge_teams/src/Controller/TeamMembersList.php on line 102

## Screenshots
will add 
[deprecatedparams_gitpatchversion.patch](https://github.com/user-attachments/files/18190479/deprecatedparams_gitpatchversion.patch)
[fix-php8-deprecated-params.patch](https://github.com/user-attachments/files/18190480/fix-php8-deprecated-params.patch)
[deprecatedparams_composerupdateversion.patch](https://github.com/user-attachments/files/18190481/deprecatedparams_composerupdateversion.patch)
![Screenshot 2024-12-16 at 3 54 17 PM](https://github.com/user-attachments/assets/9b08f617-f961-4697-8c1a-92bcccd89f64)

hollyw2025:
Thank you, waiting for the next deployment

kedarkhaire:
Till the new deployment is released, you can use this [patch](https://github.com/apigee/apigee-edge-drupal/pull/1102.patch).

[...]

PR #1110, opened 2025-01-16
Moving 3x fixes to 4x branch

### LINKED ISSUE #1098: Drupal Apigee API Product RBAC not filtering API Products based on Roles

kedarkhaire:
Fixes #1098 , #1103 & #1106 for 4.x branch

### DIFF
--- a/modules/apigee_edge_apiproduct_rbac/apigee_edge_apiproduct_rbac.module
+++ b/modules/apigee_edge_apiproduct_rbac/apigee_edge_apiproduct_rbac.module
@@ -67,7 +67,7 @@ function apigee_edge_apiproduct_rbac_api_product_access(EntityInterface $entity,
 
     if (empty($entity->getAttributeValue($rbac_attribute_name))) {
       if ('assign' === $operation) {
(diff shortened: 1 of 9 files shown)


shishir-intelli (approved):
LGTM

PR #1111, opened 2025-01-20
Version bump for PHP 8.3, Apigee PHP Client 4.x & Drupal 11

kedarkhaire:
Version bump for PHP 8.3, Apigee PHP Client 4.x & Drupal 11.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,10 +4,10 @@
     "type": "drupal-module",
     "description": "Apigee for Drupal.",
     "require": {
-        "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
+        "php": "~8.3.0",
         "ext-json": "*",
-        "apigee/apigee-client-php": "~3.0.7",
-        "drupal/core": "^10.2",
+        "apigee/apigee-client-php": "4.x-dev",
+        "drupal/core": "^11.1",
         "drupal/entity": "^1.0",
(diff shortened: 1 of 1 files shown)


divya-intelli (approved):
LGTM

divya-intelli (approved):
LGTM

composer.json
```diff
@@ -4,10 +4,10 @@
     "type": "drupal-module",
     "description": "Apigee for Drupal.",
     "require": {
-        "php": "~8.1.0 || ~8.2.0 || ~8.3.0",
+        "php": "~8.3.0",
```
mxr576, line 7: Why PHP < 8.3.0 gets dropped when they are still under active security support?  Are there any metrics indicates that on-prem customers are already on PHP 8.3.0 at least?

https://www.php.net/supported-versions.php

mxr576: ahh, probably false alarm since 4.x is for Drupal 11 compatibility, so there is no question about minimum requirements. Sorry for the noise.

kedarkhaire: Hi @mxr576 
This is for a new branch which will be used for only Drupal 11, 3.x branch will keep PHP <= 8.3.0 support 
Minimum requirement of PHP for Drupal 11 is v8.3

https://www.drupal.org/docs/getting-started/system-requirements/php-requirements

PR #1102, opened 2024-12-17
Changes for deprecated message Optional parameter $config_factory being declared before required parameter $app_warnings_checker

kedarkhaire:
Fixes #1091 & #1103

PR #1132, opened 2025-02-16
Deperecation error while installing the site fixed

sainath1991:


shishir-intelli (changes requested):
@sainath1991 
Thank you for your contribution.

It appears that this PR is incomplete. Could you please update the docblock for the constructor to reflect the new parameter order? 
Additionally, ensure that the createInstance method is updated to match the new parameter order in the constructor.

kedarkhaire:
Hi @sainath1991 
The issue is already fixed in this PR - https://github.com/apigee/apigee-edge-drupal/pull/1102

We will be releasing a new version in next week which will have this change, so till that time can you use the above [patch](https://patch-diff.githubusercontent.com/raw/apigee/apigee-edge-drupal/pull/1102.patch) for your error-free installation.

Closing this PR for now as it won't be required.

Thanks!

PR #1145, opened 2025-05-12
Clean up finished jobs after batch processing (#1144)

### LINKED ISSUE #1144: The apigee_edge_job gets HUUUGE because nothing deletes from it
## Description
The `apigee_edge_job` table may get out of control in its size (eg. over a hundred gigabytes) because nothing DELETEs from it.

## Apigee Info
Apigee Edge is being used in the particular case this become visible, but X and OPDK might also be affected.

## Steps to Reproduce
1. Check the number of entries in the table (eg. via `drush sqlq "select count(id), status from apigee_edge_job group by status order by status"`). Note that most entries having the status of 5 (`Job::FINISHED`).
2. Go to `/admin/config/apigee-edge/developer-settings/sync` and click on 'Run developer sync'.
3. Alternatively, use `drush apigee-edge:sync`.
4. Wait until the process finishes.
5. Check the number of entries in the table again.

## Actual Behavior
The number of entries got higher (by the number of developers), and most of the increases are falling into the `Job::FINISHED` status (5).

## Expected Behavior
When the process has finished, only those entries should be kept that couldn't be dealt with. Finished entries should be gone.

## Screenshots
N/A

## Notes
Removing jobs from the table right when they're finished might be problematic, since the [batch executor](https://github.com/apigee/apigee-edge-drupal/blob/3.x/src/Controller/DeveloperSyncController.php#L183-L200) calculates with the finished jobs.

Also worth noting the same happens when [syncing teams](https://github.com/apigee/apigee-edge-drupal/blob/3.x/modules/apigee_edge_teams/src/Controller/TeamMemberSyncController.php#L183-L200).

## Version Info
```
```

However, initial investigation reveals only the apigee_edge module seems to matter; and its latest versions (3.0.11 and 4.0.1) are both still affected.

boobaa:
Fixes #1144

### DIFF
--- a/src/JobExecutorInterface.php
+++ b/src/JobExecutorInterface.php
@@ -99,4 +99,12 @@ interface JobExecutorInterface {
    */
   public function countJobs(?string $tag = NULL, ?array $statuses = NULL): int;
 
+  /**
+   * Cleans up finished jobs in the queue.
+   *
+   * @param string $tag
+   *   Tag to filter with.
+   */
+  public function cleanup(string $tag): void;
+
 }
--- a/src/JobExecutorInterface.php
+++ b/src/JobExecutorInterface.php
@@ -104,6 +104,8 @@ interface JobExecutorInterface {
    *
    * @param string $tag
    *   Tag to filter with.
+   *
+   * @todo Handle race conditions.
    */
   public function cleanup(string $tag): void;
 
--- a/modules/apigee_edge_teams/src/Controller/TeamMemberSyncController.php
+++ b/modules/apigee_edge_teams/src/Controller/TeamMemberSyncController.php
@@ -148,6 +148,7 @@ class TeamMemberSyncController extends ControllerBase {
       'operations' => [
         [[static::class, 'batchGenerateJobs'], [$tag]],
         [[static::class, 'batchExecuteJobs'], [$tag]],
+        [[static::class, 'batchCleanupJobs'], [$tag]],
       ],
       'finished' => [static::class, 'batchFinished'],
     ];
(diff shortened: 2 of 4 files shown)


src/JobExecutorInterface.php
```diff
@@ -99,4 +99,12 @@ public function cast(Job $job);
    */
   public function countJobs(?string $tag = NULL, ?array $statuses = NULL): int;
 
+  /**
+   * Cleans up finished jobs in the queue.
+   *
+   * @param string $tag
+   *   Tag to filter with.
+   */
+  public function cleanup(string $tag): void;
```
mxr576, line 110: If this interface is considered a public API then it cannot be modified within this major version. As commented code with a todo for modifying the interface in 5.x  can be added here.

kedarkhaire:
Hi @boobaa 
We reviewed the PR, and we have one suggestion to add in this implementation.

Instead of deleting all the entries, keep the details for last 1 month in the table with created or updated column.

Thanks!

src/JobExecutorInterface.php
kedarkhaire: @mxr576 Are we planning to release this change in next major release 5.x or do we need to make changes to this PR for current version release i.e. 4.x?

mxr576: Currently there is two supported major version of this module if I am not mistaken. Releasing a new major version would just create more work both on upstream and downstream. 

So I think this PR should be fixed to be compatible with 4.x by having this method only as comment present on the interface in 4.x  with a todo for 5.x

kedarkhaire: Okay, @boobaa 
Can you please update the required todos as comment, so that we can merge this PR ?

kedarkhaire: Updated

kedarkhaire (approved):
LGTM

Issue #1147, opened 2025-05-15
Changing allowed_values of list_text fields on Apigee entities potentially times out


boobaa:
## Description
Drupal core [has introduced a change](https://www.drupal.org/project/drupal/issues/2521800) (in 10.2.1 if I'm not mistaken) to the admin UI for changing allowed_values for list_text fields (in `\Drupal\options\Plugin\Field\FieldType\ListItemBase::storageSettingsForm()`): for every allowed_value, the affected entity type's storage is being asked if there is an entity with the given value on the given field, and if so, the Remove button is disabled. (Before this change, there was no such check, so the Remove button was always available, even if there was an entity having the given value on the given field.)

Asking local storage (Drupal DB) is much cheaper than asking Apigee for such values, but that's only one part of the problem. In case of Apigee entities, this check MUST use the Apigee Edge Mgmt API to retrieve anything from that storage. This API does NOT allow conditional retrieval; IOW, when it's asked for those entities that already have a given value in the given field, it returns no entities at all, therefore rendering this core functionality useless. For the `team` entity type, this can be seen from the logs (with the apigee_edge_debug.module enabled): the `GET /v1/organizations/test-boobaa/companies?expand=true HTTP/1.1` request is sent to Apigee Edge, without pagination, without filtering.

Additionally, this very same request is being sent to the Mgmt API for X times during a single page load, where X is the number of already-present allowed_values. One might ask why the response is NOT cached. The answer lies in `\Drupal\apigee_edge\Entity\Storage\EdgeEntityStorageBase::doLoadMultiple()` which is NOT given any IDs to load (as it should load all entities with the given field having the given value), in which case it MUST NOT respond from its cache as there might be some new entries in the storage in between the API requests (well, at least theoretically).

These things together might lead to PHP timeouts on pages like `/admin/config/apigee-edge/team-settings/fields/team.team.field_my_list_text_field` if the field already has so many allowed_values (like 100+ for a country field).

## Apigee Info
We have seen this happening on both Apigee Edge and OPDK, but theoretically this is also present with Apigee X (given the module uses the same PHP SDK, therefore the same Mgmt API).

## Steps to Reproduce
1. Add a list_string field to any fieldable Apigee entity type (eg. team) with a handful of allowed_values.
2. Enable apigee_edge_debug.module.
3. Visit the same field's edit page.
4. Visit `/admin/reports/dblog` to see that Apigee has been called X many times during the last page load.

## Actual Behavior
[...]

PR #1148, opened 2025-05-16
Cache Apigee entity ids in memory during one page load

boobaa:
Addresses #1147

I can't say it *solves* that issue, because if there is a change in Apigee during a single page load, then this solution might be misleading. OTOH, I have no idea how to get around that *without* asking Apigee again and again.

So it seems to be a chicken-or-egg problem for me:
- we either want to be theoretically correct in all cases (therefore asking Apigee all the time), or
- we try to avoid timeouts and accept that this solution only works theoretically correctly in those cases when there's no change in Apigee during a single page load.

I, for one, am leaning towards the latter, for better UX.

### DIFF
--- a/src/Entity/Storage/EdgeEntityStorageBase.php
+++ b/src/Entity/Storage/EdgeEntityStorageBase.php
@@ -75,6 +75,11 @@ abstract class EdgeEntityStorageBase extends DrupalEntityStorageBase implements 
   protected $systemTime;
 
   /**
+   * Indicates that all entities should be/have been loaded in this request.
+   */
+  private bool $allEntitiesHaveBeenLoaded = FALSE;
+
+  /**
    * Constructs an EdgeEntityStorageBase instance.
    *
    * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
@@ -468,10 +473,48 @@ abstract class EdgeEntityStorageBase extends DrupalEntityStorageBase implements 
         Cache::invalidateTags([$this->entityTypeId . ':values']);
       }
     }
+    // When resetting cache for ANY of the entities, cache entry for ALL
+    // entities MUST also be cleared to prevent cache poisoning eg. after a
+    // loadMultiple(); save(); loadMultiple(); cycle.
+    $this->memoryCache->delete($this->entityTypeId . '-all-entity-ids');
     // We do not clear the entity controller's cache here because our main goal
     // with the entity controller cache to reduce the API calls that we
     // send to Apigee Edge. Although we do delete the entity controller's cache
     // when it is necessary, like in loadUnchanged().
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function loadMultiple(?array $ids = NULL) {
+    // Without $ids the caller wants to load ALL entities. To speed things up,
+    // let's assume the storage does not change during a page load; IOW, load
+    // all the entities for the first time and store them in a memory cache.
+    // For all subsequent calls for the same page load, return entities from
+    // that memory cache instead (without making API calls to Apigee).
+    if ($ids === NULL) {
+      $all_known_entity_ids = $this->memoryCache->get($this->entityTypeId . '-all-entity-ids');
+      if ($all_known_entity_ids !== FALSE) {
+        return parent::loadMultiple($all_known_entity_ids->data);
+      }
+      $this->allEntitiesHaveBeenLoaded = TRUE;
+      $return = parent::loadMultiple($ids);
+      $this->allEntitiesHaveBeenLoaded = FALSE;
+      return $return;
+    }
+
+    return parent::loadMultiple($ids);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setStaticCache(array $entities) {
+    if ($this->allEntitiesHaveBeenLoaded) {
+      $ids = array_map(static fn(DrupalEdgeEntityInterface $entity) => $entity->id(), $entities);
+      $this->memoryCache->set($this->entityTypeId . '-all-entity-ids', $ids);
+    }
+    parent::setStaticCache($entities);
   }
 
 }
--- a/tests/src/Kernel/Entity/ListBuilder/AppListBuilderTest.php
+++ b/tests/src/Kernel/Entity/ListBuilder/AppListBuilderTest.php
@@ -327,6 +327,7 @@ class AppListBuilderTest extends ApigeeEdgeKernelTestBase {
           'apps' => [
             $this->approvedAppWithApprovedCredential,
             $this->approvedAppWithOneRevokedCredential,
+            $this->approvedAppWithAllRevokedCredential,
             $this->revokedAppWithRevokedCredential,
             $this->approvedAppWithExpiredCredential,
             $this->revokedAppWithExpiredCredential,


src/Entity/Storage/EdgeEntityStorageBase.php
```diff
@@ -74,6 +74,11 @@ abstract class EdgeEntityStorageBase extends DrupalEntityStorageBase implements
    */
   protected $systemTime;
 
+  /**
+   * Are we loading all entities or just some/one?
+   */
+  private bool $loading_all = FALSE;
```
mxr576:
```suggestion
  private bool $allEntitiesHasBeenLoaded = FALSE;
```

src/Entity/Storage/EdgeEntityStorageBase.php
```diff
@@ -474,4 +479,38 @@ public function resetCache(?array $ids = NULL) {
...
+    // let's assume the storage does not change during a page load; IOW, load
+    // all the entities for the first time and store them in a memory cache.
+    // For all subsequent calls for the same page load, return entities from
+    // that memory cache instead (without making API calls to Apigee).
+    if ($ids === NULL) {
+      $all_cache_ids = $this->memoryCache->get($this->entityTypeId . '-all-cache-ids');
+      if ($all_cache_ids !== FALSE) {
+        return parent::loadMultiple($all_cache_ids->data);
+      }
+      $this->loading_all = TRUE;
```
mxr576:
```suggestion
      $this->allEntitiesHasBeenLoaded = TRUE;
```

src/Entity/Storage/EdgeEntityStorageBase.php
```diff
@@ -474,4 +479,38 @@ public function resetCache(?array $ids = NULL) {
...
+      $return = parent::loadMultiple($ids);
+      $this->loading_all = FALSE;
```
mxr576:
```suggestion
      $this->allEntitiesHasBeenLoaded = FALSE;
```

src/Entity/Storage/EdgeEntityStorageBase.php
```diff
@@ -474,4 +479,38 @@ public function resetCache(?array $ids = NULL) {
...
+
+    return parent::loadMultiple($ids);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setStaticCache(array $entities) {
+    if ($this->loading_all) {
+      $ids = array_map(fn($entity) => $entity->id(), $entities);
```
mxr576:
```suggestion
      $ids = array_map(static fn(\Drupal\apigee_edge\Entity\EdgeEntityInterface $entity) => $entity->id(), $entities);
```

src/Entity/Storage/EdgeEntityStorageBase.php
```diff
@@ -474,4 +479,38 @@ public function resetCache(?array $ids = NULL) {
...
+      $this->memoryCache->set($this->entityTypeId . '-all-cache-ids', $ids);
```
mxr576:
```suggestion
      $this->memoryCache->set($this->entityTypeId . '-all-entity-ids', $ids);
```

src/Entity/Storage/EdgeEntityStorageBase.php
```diff
@@ -474,4 +479,38 @@ public function resetCache(?array $ids = NULL) {
...
+   * {@inheritdoc}
+   */
+  public function loadMultiple(?array $ids = NULL) {
+    // Without $ids the caller wants to load ALL entities. To speed things up,
+    // let's assume the storage does not change during a page load; IOW, load
+    // all the entities for the first time and store them in a memory cache.
+    // For all subsequent calls for the same page load, return entities from
+    // that memory cache instead (without making API calls to Apigee).
+    if ($ids === NULL) {
+      $all_cache_ids = $this->memoryCache->get($this->entityTypeId . '-all-cache-ids');
```
mxr576:
```suggestion
      $all_cache_ids = $this->memoryCache->get($this->entityTypeId . '-all-entity-ids');
```

mxr576:
```suggestion
      $all_known_entity_ids = $this->memoryCache->get($this->entityTypeId . '-all-cache-ids');
```

maybe this variable name would be more descriptive

src/Entity/Storage/EdgeEntityStorageBase.php
```diff
@@ -74,6 +74,11 @@ abstract class EdgeEntityStorageBase extends DrupalEntityStorageBase implements
    */
   protected $systemTime;
 
+  /**
+   * Are we loading all entities or just some/one?
```
mxr576:
```suggestion
   * Indicates that all entities should be/have been loaded in this request.
```

mxr576 (commented):
Added some cosmetical changes

mxr576:
I think this could work, what I would still double check if the following sequence could still return stale data from the cache or cache invalidation works properly:

* `loadMultiple()`
* `delete('foo')` or `update(`foo`)`
* `loadMulitple()`  --- This must not return the deleted "foo" entity OR must not return the oudated "foo" entity after the update.

boobaa:
> I think this could work, what I would still double check if the following sequence could still return stale data from the cache or cache invalidation works properly:
> 
> * `loadMultiple()`
> * `delete('foo')` or `update(`foo`)`
> * `loadMulitple()`  --- This must not return the deleted "foo" entity OR must not return the oudated "foo" entity after the update.

This was a valid concern, thanks: it was also surfaced by some automated test that confirmed its presence. Now it's addressed, too.

kedarkhaire:
Hi @boobaa Is this PR ready for review ?

boobaa:
Yes.

kedarkhaire:
Hi @boobaa I have checked the issue by following the mentioned steps, I was able to reproduce the issue & also checked the PR,
Applied the patch and and followed the steps to reproduce.

I see you tried to cache the entity, but it is not making any change in output neither it is reducing the time to load the page also I see the data is loaded from API instead of cache on 2nd load.

This API is taking much time to load the page `/apps?expand=true&includeCred=true`, we need to skip calling this api on certain pages, this will solve the issue.

Thanks!

mxr576:
I think the PR title should be adjusted to "Cache Apigee entity ids in memory during one page load"

kedarkhaire (changes requested):
Hi @boobaa 
Testcases are not passing - https://github.com/kedarkhaire/apigee-edge-drupal/actions/runs/15881205329/job/44781735125#step:13:43

Please check & update the test cases also as per the changes.

Thanks!

boobaa:
I tried setting up a local environment following CONTRIBUTING.md, but there failing test ratio is like 50+% so I'm unable to reproduce the failure we're seeing on the CI. Also, I'm not sure that document is up to date.

boobaa:
Took some time, but here's a fix – tho I must admit I don't really understand what's going on here. I was having the impression that loading all entities from a storage twice, without having anything between those two loads, would yield the same entities (at least the same amount of entities). Turns out that's not the case with the `developer_app` entities.

The `$entity_type_manager->getListBuilder(static::ENTITY_TYPE)->render()` call (in `\Drupal\Tests\apigee_edge\Kernel\Entity\ListBuilder\AppListBuilderTest::testAppWarnings()`) loads the developer apps TWICE:

- first time in `\Drupal\Core\Entity\EntityListBuilder::render()`, then
- second time in `\Drupal\apigee_edge\Entity\ListBuilder\AppListBuilder::render()`.

Now that's the problem:

1. Without the changes from this PR at all, the first one returns 5 items while the second one returns 6, thus the test was a PASS.
2. With the PR, but without this last commit both the first and the second ones return 5 items, thus the test was a FAIL.
3. With this last commit, both return 6 items, thus the test is a PASS again.

So I have no idea what's exactly wrong here:

- Before the PR, the test might have been an incorrect PASS – but we cannot prove it, because even without changes from this PR but adding the missing line introduced by the last commit the test would still be a PASS.
- There might be something fishy in the `developer_app` storage, which this PR shed some light upon, but without actually fixing the problem in that storage.
- Something else.

As the CI now reports all tests PASS, I don't know how to proceed now: we can either consider this solved, or start investigating why feeding only 5 items into a storage will make the `load()` call return 6 items for non-first calls. Which is totally different kind of a problem this PR tried to address… :shrug:

kedarkhaire (approved):
LGTM

PR #1149, opened 2025-05-22
Remove entity save on config form

### LINKED ISSUE #733: Drupal 11 with Devportal on install, gives issue on GCP credentials page
## Description
On install page while configuring GCP credentials, the next step throws installation issue.
Symfony\Component\DependencyInjection\Exception\RuntimeException: You have requested a synthetic service ("kernel"). The DIC does not know how to construct this service. in Symfony\Component\DependencyInjection\ContainerBuilder->createService() (line 1093 of vendor/symfony/dependency-injection/ContainerBuilder.php).

## Expected Behavior
Installation should be completed without any errors.

## Notes
Update the forms deprecated settings is install sets.

## Version Info
Drupal 11

kedarkhaire:
Fixes https://github.com/apigee/apigee-devportal-kickstart-drupal/issues/733 issue.

Saving entity on Config form creates synthetic service call, which throws issue on Drupal 11 Devportal installation.

Storing key values & other required is done by Keys module save & submit() function.

Thanks!

### DIFF
--- a/src/Form/AuthenticationForm.php
+++ b/src/Form/AuthenticationForm.php
@@ -183,6 +183,8 @@ class AuthenticationForm extends KeyEditForm {
     // @see https://www.drupal.org/project/key/issues/3048562
     $status = parent::save($form, $form_state);
 
+    // Save the authentication key entity id to module's configuration.
+    $this->configFactory->getEditable(static::CONFIG_NAME)->set('active_key', $this->entity->id());
     // Override the redirect destination.
     $form_state->setRedirect('apigee_edge.settings');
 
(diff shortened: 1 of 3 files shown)


src/Form/AuthenticationForm.php
```diff
@@ -183,8 +183,6 @@ public function save(array $form, FormStateInterface $form_state) {
     // @see https://www.drupal.org/project/key/issues/3048562
     $status = parent::save($form, $form_state);
 
-    // Save the authentication key entity id to module's configuration.
-    $this->configFactory->getEditable(static::CONFIG_NAME)->set('active_key', $this->entity->id())->save();
```
shishir-intelli, line 187: Why are we removing this? how do we save the `active_key` if not here?

shishir-intelli (changes requested):

src/Form/AuthenticationForm.php
kedarkhaire: Updated.
After Configs need to be set(), not save().

shishir-intelli (approved):
LGTM

PR #1153, opened 2025-06-18
Eliminate repated sync of developer/users with no reason

### LINKED ISSUE #1152: `drush apigee:sync` always updates users
## Description

When you run `drush apigee:sync` multiple times you may notice that logs always says  that users were updated, which should not happen after a successful sync has happened before unless either the developer account or the related Drupal user were updated and their changed time is different.

## Apigee Info
N/A

## Steps to Reproduce
* `drush apigee:sync`
* `drush apigee:sync`
* `drush apigee:sync`
* `drush apigee:sync`
* ...

## Actual Behavior
`Updating user (john.doe@example.com) from Apigee Edge.` appears as many times a `drush apigee:sync` was executed.

## Expected Behavior

An entry related to `john.doe@example.com` should only appear once, either when it was created or updated, but not more.

## Screenshots
N/A

## Version Info
Apigee Edge module 4.0.2

mxr576:
Closes #1152 

TODO
- [x] Code clean up would be nice but the code as-is should also work

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -1365,6 +1365,7 @@ function apigee_edge_user_presave(UserInterface $account) {
     }
     $developer = $result->getDeveloper();
     $developer->save();
+    \Drupal::getContainer()->get('apigee_edge.dev_sync.last_update_tracker')->set($developer->getEmail(), \Drupal::time()->getCurrentTime());
   }
   catch (\Exception $exception) {
     $previous = $exception->getPrevious();
@@ -1386,6 +1387,7 @@ function apigee_edge_user_presave(UserInterface $account) {
           // "originalEmail" property's value on the entity.
           $developer->enforceIsNew(TRUE);
           $developer->save();
+          \Drupal::getContainer()->get('apigee_edge.dev_sync.last_update_tracker')->set($developer->getEmail(), \Drupal::time()->getCurrentTime());
         }
         catch (\Exception $exception) {
           $context = [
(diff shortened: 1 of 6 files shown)


mxr576:
Okay, so the same trick is not going to work inside `\Drupal\apigee_edge\Job\DeveloperCreateUpdate::executeRequest()` because the last modified date on the developer is a immutable and calculated on the server (as it should), so we either store the last sync date:
* in an attribute, which could lead to sync issues when multiple Drupal instances syncs from the same Apigee instance :stop_sign: 
* in the Drupal's database - this may work :heavy_check_mark: 

or else...?

kedarkhaire (changes requested):
Hi @mxr576 
Testcases are not passing - https://github.com/kedarkhaire/apigee-edge-drupal/actions/runs/15880850942/job/44780502524#step:13:43

Please check & update the test cases also as per the changes.

Thanks!

mxr576:
Thanks for the notice. I need to investigate what happens and what would be a proper fix

kedarkhaire:
> Thanks for the notice. I need to investigate what happens and what would be a proper fix

Marking this PR to draft as testcase fix need to be implemented.

mxr576:
Hm, previously failing tests got fixed in HEAD? :thinking: 

This looks unrelated:

```
1) Drupal\Tests\apigee_edge_apiproduct_rbac\FunctionalJavascript\ApiProductRoleBasedAccessAuthenticatedInternalTest::testApiProductAccess
WebDriver\Exception\JavaScriptError: javascript error: jQuery is not defined
```

Issue #1150, opened 2025-06-13
Loading some team-related pages takes ages


boobaa:
## Description
More than one of our customers reported that loading some team-related pages takes ages (5+ seconds, sometimes over a minute which causes PHP timeouts). Those customers are common in having a relatively high number of teams (10k+), with relatively low number of members in each (<20). Some already-identified examples:
- `/user/{uid}/team-invitations`
- `/teams/{team}/members`
- `/teams`
- `/admin/config/apigee-edge/team-settings/fields/team.team.field_list_text` (which is a field with the type of list_text, having 10-100 allowed_values).

By the help of apigee_edge_debug.module, we could identify that Apigee Edge is requested 11 times for 10 allowed_values on this last one – this triggered us submitting #1147 and helped to alleviate the problem somewhat: the other pages are still loading slow.

This debug module also told us that Apigee Edge itself is NOT the bottleneck: some 11 MB of JSON data was returned in about 400 msecs (for an OPDK customer; for cloud customers it was a bit higher, but still not the bottleneck).

We disabled apigee_edge_debug.module (as it also takes a considerable amount of time for only _logging_ this amount of data), and started debugging ourselves. Some pain points we have found so far, after applying #1148 for a quick remedy:
- Loading some pages (`/user/{uid}/team-invitations` and `/teams/{team}/members` in particular) takes ages only for the first time; subsequent page loads are fast (as they're coming from Drupal's cache).
- Both the `/teams` page and `/admin/config/apigee-edge/team-settings/fields/team.team.field_list_text` rely on a `$teamStorage->loadMultiple(NULL)` call at some point. Because of that, a `GET /v1/organizations/{organization}/companies?expand=true` request is sent to Apigee Edge, and the returned 10k+ teams are then instantiated as Drupal entities, which are then cached. Now this *caching* is the culprit: `\Drupal\apigee_edge\Entity\Storage\EdgeEntityStorageBase::setPersistentCache` calls `\Drupal\pgsql\Driver\Database\pgsql\Upsert::execute` for EACH team, for EACH page load.

At this point we decided to kill the team storage cache (by using a `\Drupal\Core\Cache\NullBackend` as its `cache_backend`). Now everything loads acceptably quickly (<2s even with cold caches), except the `/teams` page (5+ secs, even with warm caches).

## Apigee Info
[...]

PR #1155, opened 2025-06-19
Optimize Apigee entity caching

mxr576:
This PR addresses performance issues from inserting Apigee entities individually into persistent cache (#1150) and fixes the ignored cache expiration value of 0, which should disable caching entirely.

**Key fixes:**
- Replace individual cache->set() with batched setMultiple() operations
- Add configurable chunk sizes for optimal performance
- Respect cache expiration value 0 to disable persistent caching
- **Fix long-standing bug where Team apps ignored cache expiration settings**

### DIFF
--- a/modules/apigee_edge_teams/src/Entity/Storage/TeamAppStorage.php
+++ b/modules/apigee_edge_teams/src/Entity/Storage/TeamAppStorage.php
@@ -23,6 +23,7 @@ namespace Drupal\apigee_edge_teams\Entity\Storage;
 use Drupal\Component\Datetime\TimeInterface;
 use Drupal\Core\Cache\CacheBackendInterface;
 use Drupal\Core\Cache\MemoryCache\MemoryCacheInterface;
+use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\apigee_edge\Entity\AppInterface;
 use Drupal\apigee_edge\Entity\Controller\AppControllerInterface;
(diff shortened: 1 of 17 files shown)


mxr576:
@boobaa to be precise, https://github.com/apigee/apigee-edge-drupal/pull/1155/commits/48fd46a7d6cdc02f3a95142c90249553e529149c is not a syntax error, but PHP 8.3 language level - which is the minimum requirement of 4.x - it is only a syntax error when the diff is applied on <= PHP 8.2. 

I would consider keep using PHP 8.3 features in 4.x and when this gets backported to 3.x, get rid of those, but up to ppl with commit rights.

modules/apigee_edge_teams/src/Entity/Storage/TeamAppStorage.php
```diff
@@ -73,10 +74,18 @@ class TeamAppStorage extends AppStorage implements TeamAppStorageInterface {
    * @param \Drupal\apigee_edge\Entity\Controller\OrganizationControllerInterface $org_controller
    *   The organization controller service.
    */
-  public function __construct(EntityTypeInterface $entity_type, CacheBackendInterface $cache_backend, MemoryCacheInterface $memory_cache, TimeInterface $system_time, TeamAppControllerFactoryInterface $team_app_controller_factory, AppControllerInterface $app_controller, OrganizationControllerInterface $org_controller) {
+  public function __construct(EntityTypeInterface $entity_type, CacheBackendInterface $cache_backend, MemoryCacheInterface $memory_cache, TimeInterface $system_time, TeamAppControllerFactoryInterface $team_app_controller_factory, AppControllerInterface $app_controller, OrganizationControllerInterface $org_controller, protected ?ConfigFactoryInterface $config = NULL) {
```
kedarkhaire, line 79: Parameter $config is not described in comment as per PHPCS - Drupal.Commenting.FunctionComment.ParamMissingDefinition

kedarkhaire (changes requested):
Hi @mxr576 
I see PHPCS changes - https://github.com/kedarkhaire/apigee-edge-drupal/actions/runs/15879063934/job/44774171588#step:10:46

Can you please update the PR for this fix.
Thanks!

modules/apigee_edge_teams/src/Entity/Storage/TeamAppStorage.php
mxr576: thx for calling my attention to it, updated the PR.

kedarkhaire (approved):
LGTM

PR #1165, opened 2025-07-15
Optimize Apigee entity caching for 3x branch

### LINKED ISSUE #1150: Loading some team-related pages takes ages

kedarkhaire:
Fixes #1150 for 3x branch

### DIFF
--- a/apigee_edge.install
+++ b/apigee_edge.install
@@ -393,7 +393,7 @@ function apigee_edge_update_9002() {
 /**
  * Set cache_insert_chunk_size for API Products, Developers, Developer apps.
  */
-function apigee_edge_update_11401(): void {
+function apigee_edge_update_10401(): void {
   $configs = [
     'apigee_edge.api_product_settings',
     'apigee_edge.developer_settings',
--- a/apigee_edge.install
+++ b/apigee_edge.install
@@ -389,3 +389,17 @@ function apigee_edge_update_9002() {
   $edge_settings['content']['callbackUrl'] = $new_edge_settings['content']['callbackUrl'];
   $config_storage->write('core.entity_view_display.developer_app.developer_app.default', $edge_settings);
 }
+
+/**
+ * Set cache_insert_chunk_size for API Products, Developers, Developer apps.
+ */
+function apigee_edge_update_11401(): void {
+  $configs = [
+    'apigee_edge.api_product_settings',
+    'apigee_edge.developer_settings',
+    'apigee_edge.developer_app_settings',
+  ];
+  foreach ($configs as $config) {
+    \Drupal::configFactory()->getEditable($config)->set('cache_insert_chunk_size', 100)->save(TRUE);
+  }
+}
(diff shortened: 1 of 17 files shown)


divya-intelli (approved):
LGTM

PR #1167, opened 2025-07-15
Save 'active_key' config when saving the general settings form

### LINKED ISSUE #1166: 'active_key' must be saved into configuration when saving the Apigee general settings form
## Description
#1149 introduced a regression, because if the `save()` method is not called then the `active_key` won't be saved into configuration and Apigee connection won't work, because with the missing configuration the module doesn't know where to get the credentials from.

## Steps to Reproduce
1. Install the Apigee edge module with 4.0.2 version.
2. Go to `/admin/config/apigee-edge/settings`, fill out the form and save it.
3. If you go any of the Apigee related pages it still says that Drupal cannot connect to Apigee, although the credentials are set. For example, go to `/admin/config/apigee-edge/developer-settings/sync` the page still says that "Cannot connect to Apigee server. Please ensure that [Apigee connection settings](http://localhost:32776/admin/config/apigee-edge/settings) are correct.".

## Actual Behavior
Drupal can't connect to Apigee.

## Expected Behavior
The connection should be working.

## Version Info
4.0.2

morvaim2:
Closes #1166.

### DIFF
--- a/src/Form/AuthenticationForm.php
+++ b/src/Form/AuthenticationForm.php
@@ -184,7 +184,7 @@ class AuthenticationForm extends KeyEditForm {
     $status = parent::save($form, $form_state);
 
     // Save the authentication key entity id to module's configuration.
-    $this->configFactory->getEditable(static::CONFIG_NAME)->set('active_key', $this->entity->id());
+    $this->configFactory->getEditable(static::CONFIG_NAME)->set('active_key', $this->entity->id())->save();
     // Override the redirect destination.
     $form_state->setRedirect('apigee_edge.settings');
 


kedarkhaire:
Hi @morvaim2 
Can you tell me, why save() is required in that line, the active key is already getting set using set() ?

Also after making that change, have you tested the same by saving the credentials?

Please update here, for the above two queries.
Thanks!

morvaim2:
Hi @kedarkhaire,

Thanks for the quick reaction!
Unfortunately `set()` is not enough because it only modifies the in-memory representation of the configuration object. So the change will exist only for the current page request.
The documentation on drupal.org also says that `save()` is necessary: "Configuration is changed or added using the `set()` method and saved using the `save()` method. Note that configuration must be explicitly saved; simply setting data into a configuration does not save it." (https://www.drupal.org/docs/drupal-apis/configuration-api/simple-configuration-api#config-writing).
And yes, I tested it before and after the change. Before the change `active_key` is not saved into configuration when saving the form as it is expected and after the change it is saved and the connection works again.

Best regards,
morvaim2

kedarkhaire:
Hi @morvaim2 
Thanks for you detailed explanation on your changes.
Let me check the implementation through all our cases and then I will update here.
Thanks!

mxr576:
@kedarkhaire Hi, do you have an update on this PR?

kedarkhaire:
Hi @mxr576 
Currently if we use this patch to save the authentication key entity id in config, we get semantic error in Drupal 11 versions.
This specially happens during fresh installation and not after the installation.
We need to get proper method to resolve the semantic error issue and save the auth key entity id in config.
Thanks!

kedarkhaire:
Hi @morvaim2 
Can you test with [this patch](https://patch-diff.githubusercontent.com/raw/apigee/apigee-edge-drupal/pull/1186.patch), and update us if it works for you ?

Thanks!

morvaim2:
Hi @kedarkhaire,
Sorry for the late reply. Yes, the patch works, thanks for fixing this!

PR #1176, opened 2025-08-28
Updates scopes for Credentials if present

### LINKED ISSUE #1174: App Edit Form does not propagate existing scopes after submission
## Description

When you edit an app via the App Edit Form, scopes that are existing on the record are cleared out, instead of being saved along with the rest of the form object.

## Apigee Info

Apigee X

## Steps to Reproduce

1. Click on the edit form for an app group / app object that has existing scopes defined on it
2. Edit some data in the form
3. Submit the form

## Actual Behavior

The existing scopes have been removed from the object

## Expected Behavior

The existing scopes should remain on the object after submitting the edit form

## Screenshots

Image

## Notes

Somewhere around this code, I believe the code needs to also propagate the scopes: https://github.com/apigee/apigee-edge-drupal/blob/1f0f840cf2595fb2f57bd9729cd07735c0eb3a79/src/Entity/Form/AppEditForm.php#L229

The REST API endpoint for `updateAppGroupAppKey` by default will remove anything that is not sent in the API call: https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.appgroups.apps.keys/updateAppGroupAppKey

The sample code in the apigee-client-php repo has an example of how to make an edit request like this: https://github.com/apigee/apigee-client-php/blob/bb055ba0287784f345e3b389fc959b954af4f560/examples/create_new_app_with_credential.php#L48-L53

kedarkhaire:
Fixes #1174

### DIFF
--- a/src/Entity/Form/AppEditForm.php
+++ b/src/Entity/Form/AppEditForm.php
@@ -207,6 +207,7 @@ abstract class AppEditForm extends AppForm {
       foreach ($form_state->getValue('credential', []) as $credential_key => $credential_changes) {
         foreach ($app->getCredentials() as $credential) {
           if ($credential_key === $credential->getConsumerKey()) {
+            $credentialScopes = $credential->getScopes();
             $original_api_product_ids = [];
             // Cast it to array to be able handle the same way the single- and
             // multi-select configuration.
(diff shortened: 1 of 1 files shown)


divya-intelli:
Resolved in #1179 , hence closing.

PR #1186, opened 2025-09-22
Updated configFactory save() method

### LINKED ISSUE #1166: 'active_key' must be saved into configuration when saving the Apigee general settings form

kedarkhaire:
Fixes #1166

### DIFF
--- a/src/Form/AuthenticationForm.php
+++ b/src/Form/AuthenticationForm.php
@@ -184,7 +184,7 @@ class AuthenticationForm extends KeyEditForm {
     $status = parent::save($form, $form_state);
 
     // Save the authentication key entity id to module's configuration.
-    $this->configFactory->getEditable(static::CONFIG_NAME)->set('active_key', $this->entity->id());
+    \Drupal::configFactory()->getEditable(static::CONFIG_NAME)->set('active_key', $this->entity->id())->save();
     // Override the redirect destination.
     $form_state->setRedirect('apigee_edge.settings');
 


kedarkhaire:
The dependency injection is not working as expected in Drupal 11, so we updated with other solution `\Drupal::configFactory()`.

Thanks!

src/Form/AuthenticationForm.php
```diff
@@ -184,7 +184,7 @@ public function save(array $form, FormStateInterface $form_state) {
     $status = parent::save($form, $form_state);
 
     // Save the authentication key entity id to module's configuration.
-    $this->configFactory->getEditable(static::CONFIG_NAME)->set('active_key', $this->entity->id());
+    \Drupal::configFactory()->getEditable(static::CONFIG_NAME)->set('active_key', $this->entity->id())->save();
```
kedarkhaire, line 187: The dependency injection is not working as expected in Drupal 11 due to the parent::save() call, so we updated with other solution `\Drupal::configFactory()`.

Thanks!

kedarkhaire: When the container is rebuilt, any services that were originally injected into form's constructor (like $this->configFactory) become "stale" or invalid. Trying to use them after the container rebuild results in the "synthetic service" error during installation, so we are using `\Drupal::configFactory()`.

Thanks!

shishir-intelli (approved):
LGTM

PR #1193, opened 2025-10-03
Cache Apigee entity ids in memory during one page load

### LINKED ISSUE #1147: Changing allowed_values of list_text fields on Apigee entities potentially times out

kedarkhaire:
Fixes #1147 for 3.x

### DIFF
--- a/src/Entity/Storage/EdgeEntityStorageBase.php
+++ b/src/Entity/Storage/EdgeEntityStorageBase.php
@@ -89,6 +89,11 @@ abstract class EdgeEntityStorageBase extends DrupalEntityStorageBase implements 
   protected $systemTime;
 
   /**
+   * Indicates that all entities should be/have been loaded in this request.
+   */
+  private bool $allEntitiesHaveBeenLoaded = FALSE;
+
+  /**
    * Constructs an EdgeEntityStorageBase instance.
    *
    * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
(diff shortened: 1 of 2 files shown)


divya-intelli (approved):
LGTM

PR #1194, opened 2025-10-03
Clean up finished jobs after batch processing

### LINKED ISSUE #1144: The apigee_edge_job gets HUUUGE because nothing deletes from it

kedarkhaire:
Fixes #1144 for 3.x

### DIFF
--- a/modules/apigee_edge_teams/src/Controller/TeamMemberSyncController.php
+++ b/modules/apigee_edge_teams/src/Controller/TeamMemberSyncController.php
@@ -148,6 +148,7 @@ class TeamMemberSyncController extends ControllerBase {
       'operations' => [
         [[static::class, 'batchGenerateJobs'], [$tag]],
         [[static::class, 'batchExecuteJobs'], [$tag]],
+        [[static::class, 'batchCleanupJobs'], [$tag]],
       ],
       'finished' => [static::class, 'batchFinished'],
     ];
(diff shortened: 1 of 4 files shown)


divya-intelli (approved):
LGTM

PR #1201, opened 2025-11-05
Removed the extra dot from annotation in TeamInvitationForm.php

shailesh-google:
Removed the extra dot from annotation in TeamInvitationForm.php


divya-intelli (approved):
LGTM

PR #1210, opened 2025-11-18
Fixing Performance issue of team page by Adding cache in 3.x

shailesh-google:
Fixing Performance issue of team page by Adding cache 

Partial Fix for #1150

### DIFF
--- a/.github/workflows/php-action.yml
+++ b/.github/workflows/php-action.yml
@@ -26,6 +26,9 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      security-events: write
     
     name: "PHP ${{ matrix.php-version }} | Drupal ${{ matrix.drupal-core }} | ${{ matrix.instance-type }}"
     strategy:
@@ -37,7 +40,6 @@ jobs:
           - "8.3"
         drupal-core:
           # Should update the following as the minimum supported version from Drupal.org
-          - "10.4.x"
           - "10.5.x"
         instance-type:
           - "Edge"
@@ -93,6 +95,7 @@ jobs:
     - name: "Allow plugins and dev dependencies"
       run: |
         cd drupal
+        composer require "justinrainbow/json-schema:^6.5.1" --no-update
         composer config --no-plugins allow-plugins.composer/installers true
         composer config --no-plugins allow-plugins.drupal/core-composer-scaffold true
         composer config --no-plugins allow-plugins.drupal/core-project-message true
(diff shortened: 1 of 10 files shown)


divya-intelli:
Hi @boobaa ,

This PR partially optimizes the team listing. Team data is cached on the first load. For all subsequent requests, the data is retrieved from the cache, leading to a significant performance gain—we observed a nearly 50% reduction in load time starting from the second load.

Next Step:Optimizing Data Rendering in Drupal, which we've identified as the next major area for speed improvement.

boobaa:
Hi @divya-intelli ,

Sounds like a good first step. As a test, I have a local site with drupal/apigee_edge:3.0.13 which is configured to fetch ALL the Apigee-related data for the `/teams` page from a local endpoint (to eliminate both the network latency AND the time Apigee Edge/OPDK would need to collect and return the data), with some 10k (!) teams.

- Without any patches, 10k `/teams` page load time: 11.1/8.4 secs (cold/warm caches)
- With this PR applied, 10k `/teams` page load time: 12.1/7.2 secs (which is interesting, but as it _decreases_ the time needed for warm caches, it should be OK)
- With this PR and a custom patch (that uses the `NullBackend` for `TeamStorage`'s cache) also applied, 10k `/teams` page load time: 9.7/5.5 secs

So while this seems to be a good first step, further optimization is still needed. Whether it's regarding data rendering or something else in Drupal, I'm not sure. Some profiling could help identifying the most painful parts in the code, but I didn't have the bandwidth for that.

shailesh-google:
Thanks @boobaa for testing this PR.

My investigation into the latency issue identified that the primary cause was the time consumed by data rendering. Consequently, pagination was adopted as the optimal solution to effectively address and optimize this rendering constraint.

We have successfully implemented Drupal pagination ( PR #1212 ) to address this, resulting in a 50% reduction in initial page load time. By combining this with caching logic (PRs #1212 and #1210 ), subsequent page loads (2nd, 3rd, and beyond) are now seeing an even greater performance boost a massive 80% to 85% reduction in load time, as all subsequent pagination data is served from the cache.

boobaa:
Hi @shailesh-google ,

I have given both #1210 and #1212 some testing, please find my results at https://github.com/apigee/apigee-edge-drupal/pull/1212#issuecomment-3612826814

However, while the pagination does answer the page load time problem, by doing that it essentially kills a feature that's badly needed for our customers: finding a particular team (from the 20k+ they're having). Please find my regarding notes at https://github.com/apigee/apigee-edge-drupal/pull/1212#issuecomment-3616278432

shailesh-google:
Hi @boobaa,

Thank you for testing PR #1210 and #1212 and for the detailed feedback.

We completely agree that with 20k+ teams, pagination alone is insufficient and a Search Feature is critical for navigability. That is definitely on our roadmap.

However, during our latest round of testing on the caching implementation PR #1210 , we discovered a critical edge case that we need to resolve first to ensure data integrity.

The Issue Identified: We found a "cache poisoning" scenario where adding a single new team was overwriting the entire "All Teams" cache list with just that one new record. This caused the rest of the team list to disappear until the cache was fully cleared again.

As soon as this stability fix is merged and verified, we will immediately start work on the Search Feature to address the usability concerns you raised.

shailesh-google:
Hi @boobaa,

Following up on our previous discussion regarding the usability of the teams list: I have implemented the Search Feature to team listing page.

Please refer to PR #1216  for the implementation and other details.

shishir-intelli (approved):
LGTM

PR #1211, opened 2025-11-20
Fixing Performance issue of team page by Adding cache in 4.x

shailesh-google:
Fixing Performance issue of team list page by Adding cache

Partial Fix for #1150


shishir-intelli (approved):
LGTM

PR #1216, opened 2025-12-18
Added Search Feature For Team Listing Page 3.x

### LINKED ISSUE #1215: Implement search functionality on Team List to resolve pagination usability issues
## Description
We recently implemented pagination on the Team List Page PR #1212 . This was necessary because loading the full list of over 10K+ teams was causing severe performance bottlenecks and very slow initial page loads.

## The Problem 
While pagination solved the performance issue, it introduced a usability regression regarding data discovery. Previously, users relied on the browser's native "Find" feature (Ctrl+F / Cmd+F) to locate specific teams. With pagination now active, Ctrl+F only searches the current page subset (e.g., the visible 50 items). Users are currently unable to find specific teams located on subsequent pages without manually clicking through thousands of records.

## Proposed Solution 

Implement a dedicated server-side search feature on the Team List page.

shailesh-google:
Added Search Feature For Team Listing Page

Fix #1215 

**Screenshot:**

https://screenshot.googleplex.com/4pbbbA2Y4nxkNp9

### DIFF
--- a/css/apigee_edge.search_form.css
+++ b/css/apigee_edge.search_form.css
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Google Inc.
+ * Copyright 2025 Google Inc.
  *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the GNU General Public License version 2 as published by the
--- a/css/apigee_edge.search_form.css
+++ b/css/apigee_edge.search_form.css
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2018 Google Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License version 2 as published by the
+ * Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+ * or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public
+ * License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+/* Target both Search and Reset buttons */
+
+.apigee-edge-search-form .button-action {
+  text-transform: none;
+  margin-left: 10px;
+  vertical-align: middle;
+}
+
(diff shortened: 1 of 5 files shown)


css/apigee_edge.search_form.css
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2018 Google Inc.
```
divya-intelli: Change year to current year

src/Form/EdgeEntitySearchForm.php
```diff
@@ -0,0 +1,99 @@
+<?php
```
divya-intelli, line 1: Copyright section missing

divya-intelli (approved):
LGTM

css/apigee_edge.search_form.css
shailesh-google: Added Current Year

src/Form/EdgeEntitySearchForm.php
shailesh-google: Copyright Section Added.

boobaa:
Hi @shailesh-google ,

Thank you for this search+pagination patch, which I just gave a try. As my usual test, I have a local site with drupal/apigee_edge:3.0.13 which is configured to fetch ALL the Apigee-related data for the `/teams` page from a local endpoint (to eliminate both the network latency AND the time Apigee Edge/OPDK would need to collect and return the data), with some 10k (!) teams.

- With both this and PR#1210 applied, first 1k of the 10k `/teams` page load time: 5.1/3.5 secs; with filtering for `0999` (so only 11 of those 10k teams get listed): 4.4/2.9 secs; with filtering for `09999` (so only one of those 10k teams get listed): 4.4/2.9 secs
- With both this and PR#1210 applied AND a custom patch (that uses the `NullBackend` for `TeamStorage`'s cache) also applied, first 1k of the 10k `/teams` page load time: 2.6/1.7 secs; with filtering for `0999` (so only 11 of those 10k teams get listed): 1.9/1.0 secs; with filtering for `09999` (so only one of those 10k teams get listed): 1.9/1.0 secs

Also worth noting the end-user experience:
1. Users navigates to `/teams`, which takes ~5.1 secs to load.
2. They enter their search phrase to the form and submit it, which takes something between ~5.1 and ~4.4 secs (due to the way Drupal forms work: POST that redirects to a GET).
3. The filtered page gets loaded, which takes ~4.4 secs.

It's still interesting that the mere usage of caching makes these pages about 2 times SLOWER (compared to no caching).

Besides these nuances, this is LGTM.

PR #1218, opened 2026-01-07
Remove Drupal 10.4.x from workflow and bump to 10.5.x

shailesh-google:
Removed version 10.4.x from the build matrix as it is no longer required. Updated the conditional checks for Code Coverage and Codecov uploads to target Drupal 10.5.x (PHP 8.3) to ensure coverage reporting continues.


shishir-intelli (approved):
LGTM

PR #1220, opened 2026-01-08
Fix dead code in apigee_edge_user_presave Function

### LINKED ISSUE #1154: Dead/broken code in apigee_edge_user_presave()
## Description

PHPStorm warns that the following code is incorrect, because ` $user_developer->convertUser()` returns `\Drupal\apigee_edge\Structure\UserToDeveloperConversionResult()` and not a developer interface.

```
            $developer = $user_developer->convertUser($account);
            $developer->setDeveloperId($developer_id);
            $developer->enforceIsNew(FALSE);
```

It also occurred to me if this code is dead or how this potentially fatal error has not been recognized so far.

## Apigee Info
N/A

## Steps to Reproduce
N/A

## Actual Behavior
N/A

## Expected Behavior
N/A

## Screenshots
N/A

## Notes
N/A

## Version Info
4.0.2

shailesh-google:
Fixes #1154 

The 'user presave' logic for syncing existing Apigee developers contained dead code that would throw a fatal error if executed. Methods were being called on the conversion Result object instead of the Developer entity. Fixed by calling the developer entity.

### DIFF
--- a/apigee_edge.module
+++ b/apigee_edge.module
@@ -1400,13 +1400,11 @@ function apigee_edge_user_presave(UserInterface $account) {
         $logger->info($previous->getMessage());
         $developer = Developer::load($account->getEmail());
         if ($developer) {
-          $developer_id = $developer->getDeveloperId();
           // If a user could register on the portal with an email address
           // that already belongs to a developer on Apigee Edge then override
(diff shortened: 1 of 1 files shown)


shishir-intelli (approved):
LGTM

PR #1227, opened 2026-01-14
Fixed the "Remove" button logic in list_text field type | 3.x

shailesh-google:
Fix: #1147 | 3.x

Fixed the "Remove" button logic in list_text field type: previously, the remove button remained active even if values were in uses in apigee entity. It is now correctly disabled (mimicking Core behavior) when the field value is present on an Apigee entity.


shishir-intelli (approved):
LGTM

Issue #1241, opened 2026-02-06
[Bug]: PHP Fatal error: Declaration of SystemEmailToUsersOfRole::doExecute incompatible with Rules module signature


shailesh-google:
We are encountering critical CI failures in the apigee_edge contrib module. The primary issue is a PHP Fatal error caused by a method signature mismatch between apigee_edge_actions and the upstream rules module.

It appears the upstream RulesAction\SystemEmailToUsersOfRole::doExecute method now enforces strict type hints (string) and a void return type, which the overriding method in apigee_edge_actions is missing.

Additionally, there are failures in apigee_edge_teams (Undefined array key) and AppSettingsFormTest (Call to member function on null) which may or may not be related to the environment state caused by the initial fatals.

shailesh-google:
Fix for Drupal 10 : #1245 
Fix for Drupal 11 : #1256 

closing the ticket.

PR #1245, opened 2026-02-12
Fix PHP Fatal error: incompatibility with Rules SystemEmailToUsersOfRole | 3.x

shailesh-google:
Fixes Drupal 10 error #1241


shishir-intelli (approved):
LGTM

PR #1243, opened 2026-02-11
[#1242] Fix masquerade compatibility.

VladimirAus:
Fix: #1242

Issue #1242, opened 2026-02-11
Error: Call to a member function getPath() when using masquerade module


VladimirAus:
## Description
Error: Call to a member function getPath() on null in apigee_edge_system_breadcrumb_alter() (line 255 of /app/web/modules/contrib/apigee_edge/apigee_edge.module)

```
#0 /app/web/core/lib/Drupal/Core/Extension/ModuleHandler.php(460): apigee_edge_system_breadcrumb_alter(Object(Drupal\Core\Breadcrumb\Breadcrumb), Object(Drupal\Core\Routing\CurrentRouteMatch), Array)
#1 /app/web/core/lib/Drupal/Core/Breadcrumb/BreadcrumbManager.php(99): Drupal\Core\Extension\ModuleHandler->alter('system_breadcru...', Object(Drupal\Core\Breadcrumb\Breadcrumb), Object(Drupal\Core\Routing\CurrentRouteMatch), Array)
#2 /app/web/core/modules/system/src/Plugin/Block/SystemBreadcrumbBlock.php(73): Drupal\Core\Breadcrumb\BreadcrumbManager->build(Object(Drupal\Core\Routing\CurrentRouteMatch))
#3 /app/web/core/modules/block/src/BlockViewBuilder.php(212): Drupal\system\Plugin\Block\SystemBreadcrumbBlock->build()
#4 [internal function]: Drupal\block\BlockViewBuilder::preRender(Array)
#5 /app/web/core/lib/Drupal/Core/Security/DoTrustedCallbackTrait.php(107): call_user_func_array('Drupal\\block\\Bl...', Array)
#6 /app/web/core/lib/Drupal/Core/Render/Renderer.php(910): Drupal\Core\Render\Renderer->doTrustedCallback('Drupal\\block\\Bl...', Array, 'Render #pre_ren...', 'exception', 'Drupal\\Core\\Ren...')
#7 /app/web/core/lib/Drupal/Core/Render/Renderer.php(441): Drupal\Core\Render\Renderer->doCallback('#pre_render', 'Drupal\\block\\Bl...', Array)
#8 /app/web/core/lib/Drupal/Core/Render/Renderer.php(253): Drupal\Core\Render\Renderer->doRender(Array, Object(Drupal\Core\Render\RenderContext))
[...]

PR #1248, opened 2026-02-12
Fix fatal error: Call to a member function getPath() on null | 3.x

shailesh-google:
Fixes for Drupal 10 #1242

Credit: @VladimirAus 
Reference PR: #1243


divya-intelli (approved):
LGTM

PR #1258, opened 2026-03-04
Eliminate repated sync of team memberships with no reason

### LINKED ISSUE #1160: Check if `drush apigee:sync` has the same issue for team member sync than for developers
Follow up on #1152

(Since team members stored in an attribute on Apigee X, from the App group's last change date it could be guessed if there is need to sync members for an App Group or not.)

davidbondor12:
Resolves #1160 

### Description
Currently, running `drush apigee-edge:teams-sync` schedules a `TeamMemberUpdate` job for every single team, regardless of whether their memberships have actually changed in Apigee. This results in unnecessary API calls and long sync times.

This PR introduces a memory mechanism (similar to the one implemented in #1153 for developers) to skip redundant syncs.

### What changed:
* **Added a Last Update Tracker:** Created the `apigee_edge_teams.team_sync.last_update_tracker` key-value service to store the timestamp of the last successful sync attempt per team.
* **Smart Scheduling:** Updated `TeamMemberSync.php` to compare the `last_synced` timestamp against the Team's `getLastModifiedAt()` timestamp from Apigee. The update job is now only scheduled if the Apigee entity is newer than our last sync.
* **Timestamp Recording:** `TeamMemberCreateUpdate.php` now records the current time into the tracker upon job execution.
* **CLI Warning Fix:** Handled a minor UI side-effect in `CliService.php`. When the sync skips all teams (queue is empty), `$context['message']` is not set. Added a null coalescing fallback to prevent an "Undefined array key" warning. Not sure if this was the right way to handle the CLI warning message.

### Testing Instructions
1. Run `drush apigee-edge:teams-sync`. The first run will process all teams and populate the tracker.
2. Run `drush apigee-edge:teams-sync` immediately again. The batch process should finish almost instantly without scheduling team updates.
3. Modify a team/AppGroup in Apigee.
4. Run `drush apigee-edge:teams-sync` again. It should only queue and process the team that was just modified.

### DIFF
--- a/modules/apigee_edge_teams/apigee_edge_teams.services.yml
+++ b/modules/apigee_edge_teams/apigee_edge_teams.services.yml
@@ -161,3 +161,8 @@ services:
   apigee_edge_teams.app_group_scope_manager:
     class: Drupal\apigee_edge_teams\Service\AppGroupScopeManager
     arguments: ['@apigee_edge.sdk_connector', '@apigee_edge.controller.organization']
+
+  apigee_edge_teams.team_sync.last_update_tracker:
+    class: Drupal\Core\KeyValueStore\KeyValueStoreInterface
+    factory: [ '@keyvalue', 'get' ]
+    arguments: [ 'apigee_edge_teams.team_sync.last_update_tracker' ]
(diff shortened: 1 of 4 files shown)


shailesh-google:
Hi @davidbondor12,

Thanks For Your PR. We will test this PR and Update you soon.

PR #1261, opened 2026-03-18
Fixed pagination limits for non-admin users on team list

### LINKED ISSUE #1247: Pagination is visible on Teams page for non-admin users with limited team access
**Description**

In the Apigee Edge module (v4.0.5), the Teams listing page incorrectly displays pagination links based on the total number of teams existing in the Apigee organization, even when the logged-in non-admin user only has access to a single team.

**Environment**

1. Apigee Edge Version: 4.0.5
2. Drupal Version: (e.g., 10.x / 11.x)
3. PHP Version: (e.g., 8.3)

**Steps to Reproduce**

1. Log in as an Administrator.
2. Create a large number of Teams (e.g., 1000+ teams) so that pagination is triggered.
3. Create a Non-Admin User and assign them as a member to only one specific Team.
4. Log in as that Non-Admin User.
5. Navigate to the Teams listing page (/teams).

**Expected Behavior**

The user should see only the team(s) they are a member of, and no pagination should appear if the number of their joined teams is less than the pager limit.

**Actual Behavior**

The user sees only their one team in the list, but the pagination/pager appears at the bottom showing multiple pages (e.g., Page 1, 2, 3). Clicking "Next" or "Page 2" results in an empty list because the user has no access to the teams on those pages.

**Technical Investigation**

It appears the query or the pager integration is using a global count of entities from the Apigee Edge backend instead of applying the DeveloperTeam filter before calculating the pager limits. The Pager object is likely initialized with the total count from the organization, ignoring the access control filtering.

**Suggested Fix**

The logic in the Team list Builder needs to ensure that the Pager initialization uses the result count of the filtered team_membership query rather than the generic entity query count.

shailesh-google:
Fix: #1247

### DIFF
--- a/modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
+++ b/modules/apigee_edge_teams/src/Entity/ListBuilder/TeamListBuilder.php
@@ -190,11 +190,11 @@ class TeamListBuilder extends EdgeEntityListBuilder {
     // the pager is initialized.
     if ($this->currentUser->isAuthenticated() && !$this->currentUser->hasPermission('administer team') && !$this->currentUser->hasPermission('view any team')) {
       $teams = $this->teamMembershipManager->getTeams($this->currentUser->getEmail());
-      
-      // FIX: Disable the pager completely if the user have a single team or no teams.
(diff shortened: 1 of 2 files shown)


shishir-intelli (approved):
LGTM

Issue #1262, opened 2026-03-24
Add support for PHP 8.4 and PHP 8.5


shailesh-google:
To ensure the project remains secure, performant, and up-to-date with the broader PHP ecosystem, we need to add official support for PHP 8.4 and prepare for PHP 8.5. This involves updating our environment requirements, resolving any new deprecations, and ensuring our continuous integration pipeline tests against these newer versions.

bdanin:
This issue is needed to unblock our PHP 8.4 upgrade. At this point, it is the only known compatibility blocker preventing us from updating our platform runtime.

shailesh-google:
Hi @bdanin,

We Already Working in this Issue. We will soon add the support of php8.4 and php8.5.

PR #1264, opened 2026-04-14
Added support for PHP 8.4 | 3.x

shailesh-google:
Fix: #1262 

1. **Deprecation notice because, as of PHP 8.4, relying on the default value of the escape parameter in str_getcsv() is officially deprecated:** https://www.php.net/manual/en/function.str-getcsv.php


divya-intelli (approved):
LGTM

PR #1267, opened 2026-04-24
Fix TeamAppAccessHandler::checkCreateAccess() to respect $context array

davidbondor12:
When checking Team App creation access programmatically via the entity access control handler (e.g., `$access_handler->createAccess('team_app', $user, ['team' => $team])`), the check was incorrectly returning neutral/false. This occurred because `checkCreateAccess()` ignored the provided `$context` array and strictly relied on the current URL route containing a `{team}` parameter.

This breaks access checks for headless architectures, custom REST/JSON:API endpoints, and any custom module that does not use standard Drupal UI routes.

This commit updates `TeamAppAccessHandler::checkCreateAccess()` to first attempt extracting the team from `$context['team']`, and only falls back to the route match parameter if the context is empty. It also includes logic to upcast string IDs to Team entities when passed via context.


shailesh-google (approved):
LGTM

PR #1279, opened 2026-06-19
Fix incomplete team list caching in TeamStorage

### LINKED ISSUE #1278: Incomplete Team lists cached as the full entity list
## Description
The `/teams` listing page occasionally displays an incomplete list, showing only a small subset of teams. Once this happens, the incomplete list is cached as the full list and shown to all users until the Drupal cache is cleared or expired.  

## Steps to Reproduce
1. Clear Drupal caches (`drush cr`).
2. Trigger a partial load of teams. For example, programmatically load more than 1 team: `\Drupal::entityTypeManager()->getStorage('team')->loadMultiple(['team_a', 'team_b']);` (or e.g. trigger a scenario where only a user's specific teams are loaded using $developer->getCompanies() for loading some teams).
3. Visit the `/teams` page.
4. **Expected:** The full list of all teams in the Apigee organization is displayed.
5. **Actual:** Only the teams loaded in Step 2 are displayed.

## Root Cause
The bug stems from how the "master list" of team IDs (`all_ids:team`) is handled in `TeamStorage.php`:

1. **Flawed all_ids caching logic in** `TeamStorage`: The `TeamStorage::setPersistentCache()` method was updated to build a "master list" of all team IDs (`all_ids:team`) and cache them. However, it caches any newly loaded entity IDs as the master list if the loaded entities are more than 1. If an operation only loads a subset of teams (e.g. only loads specific teams, or a developers teams.), the `all_ids:team` cache is set with only that small subset. When the main Teams page is later loaded, getFromPersistentCache(NULL) assumes this subset is the complete list of teams in the organization.

## Proposed Resolution
1. Remove the overridden `setPersistentCache()` method entirely from `TeamStorage`.
2. Move the master ID list caching logic into `doLoadMultiple()`.
3. Introduce strict conditions to only cache the master ID list if:
   * A full load was explicitly requested (`$ids === NULL`).
   * The entity type is persistently cacheable and has a valid expiration.

## Version Info
3.x, 4.x possibly as well since the code appears to be the same.

davidbondor12:
Fixes [#1278](https://github.com/apigee/apigee-edge-drupal/issues/1278)
This commit resolves an issue where an incomplete subset of Apigee teams was being cached and displayed as the complete list for the organization.

Changes:
* TeamStorage.php: Removed flawed logic in `setPersistentCache()` that built the `all_ids:team` cache based on partial entity loads. Overrode `doLoadMultiple()` to ensure the master ID list is only cached when a complete list of entities is explicitly requested from the Apigee API ($ids === NULL).

### DIFF
--- a/modules/apigee_edge_teams/src/Entity/Storage/TeamStorage.php
+++ b/modules/apigee_edge_teams/src/Entity/Storage/TeamStorage.php
@@ -261,43 +261,25 @@ class TeamStorage extends AttributesAwareFieldableEdgeEntityStorageBase implemen
   /**
    * {@inheritdoc}
    */
-  protected function setPersistentCache(array $entities) {
-    parent::setPersistentCache($entities);
+  protected function doLoadMultiple(?array $ids = NULL) {
+    $entities = parent::doLoadMultiple($ids);
 
-    $entity_count = 0;
-    if (!empty($entities)) {
-      // Get all entity IDs.
-      $all_entity_ids = array_keys($entities);
-      $entity_count = count($all_entity_ids);
-    }
-
-    // After all chunks are saved, save the master ID list.
-    // Only proceed if we have entities to process.
-    if ($entity_count > 0) {
-      $all_ids_cid = 'all_ids:' . $this->entityTypeId;
-      // Use the main entity type tag so this item is cleared when
-      // the rest of the entity cache is cleared.
+    // Only cache the master ID list if a full load was explicitly requested,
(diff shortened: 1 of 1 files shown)


shailesh-google:
Hi @davidbondor12,

Thanks for the PR! I've tested your changes and can confirm that they successfully resolve the issue described in #1278

Could you also create a matching PR for the 3.x branch to bring this fix over to the Drupal 10 environment? Thanks again for your contribution!

shailesh-google (approved):
LGTM

